PR #6, opened 2025-04-17
Initialize attnumsWithEntries to prevent failures during CO rewrite

adam8157:
The rewrite process detects previous encodings of every column and stores them
in `attnumsWithEntries[]` to decide whether to add a new encoding or update the
existing one. Only the first 8 elements of that array were initialized, which could
result in "ERROR: could not find tuple for attnum X for relid Y during scan on
pg_attribute_encoding" in most cases, especially with wide tables.

### DIFF
--- a/src/backend/catalog/pg_attribute_encoding.c
+++ b/src/backend/catalog/pg_attribute_encoding.c
@@ -932,6 +932,8 @@ ExistValidLastrownums(Oid relid, int natts)
 
 /*
  * Determine which attnums have an entry present in pg_attribute_encoding
+ *
+ * attnum_entry_present[] should be initialized with false.
  */
 void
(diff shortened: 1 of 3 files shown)


linwen (commented):
The bug was caused by non initialization of a bool array.  In function check_attribute_encoding_entry_exist, there is a memset call, but it doesn't memset the whole array.
MemSet(attnum_entry_present, false, sizeof(attnum_entry_present)); 
So this line could be removed.
Or in another way,  modify the interface of function check_attribute_encoding_entry_exist, add a new parameter which is length of the bool array, and memset the whole array.
check_attribute_encoding_entry_exist(Oid relid, int num, bool *attnum_entry_present)
MemSet(attnum_entry_present, false, num * sizeof(bool));

linwen (approved):
I have verified it locally.

PR #8, opened 2025-04-17
Fix src/test/ssl "double free or corruption" failure

adam8157:
The root error is postmaster processing SSL failed with "no PostgreSQL
user name specified in startup packet", cherry pick the upstream fix.

        Use BIO_{get,set}_app_data instead of BIO_{get,set}_data.

        We should have done it this way all along, but we accidentally got
        away with using the wrong BIO field up until OpenSSL 3.2.  There,
        the library's BIO routines that we rely on use the "data" field
        for their own purposes, and our conflicting use causes assorted
        weird behaviors up to and including core dumps when SSL connections
        are attempted.  Switch to using the approved field for the purpose,
        i.e. app_data.
        ...

### DIFF
--- a/configure
+++ b/configure
@@ -14113,7 +14113,7 @@ done
   # defines OPENSSL_VERSION_NUMBER to claim version 2.0.0, even though it
   # doesn't have these OpenSSL 1.1.0 functions. So check for individual
   # functions.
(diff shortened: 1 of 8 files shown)


adam8157:
Updated the commit with cherry-pick information.

linwen (dismissed):
This fix is ported from upstream 12 stable. Without this fix, SSL connection may cause process crash with openssl3.2.0.

PR #12, opened 2025-04-25
Workaround the `capture_output` parameter in cases

adam8157:
Python added the capture_output parameter at Python 3.7, but EL8 only
has Python 3.6, change the `capture_output` parameter to its alternative
`stdout=subprocess.PIPE, stderr=subprocess.PIPE`.


linwen (approved):
warehousepg7 uses python3 provided by Operating System. EL8's python version is 3.6 which has no parameter "capture_output", so use its alternative.

PR #14, opened 2025-04-28
Fix mirror promote failure

Smyatkin-Maxim:
When promoting a mirror segment due to failover we have seen a stacktrace like this:
```
"FATAL","58P01","requested WAL segment pg_xlog/00000023000071D50000001F has already been removed",,,,,,,0,,"xlogutils.c",580,"Stack
trace:
1    0x557bdb9f09b6 postgres errstart + 0x236
2    0x557bdb5fc6cf postgres <symbol not found> + 0xdb5fc6cf
3    0x557bdb5fd021 postgres read_local_xlog_page + 0x191
4    0x557bdb5fb922 postgres <symbol not found> + 0xdb5fb922
5    0x557bdb5fba11 postgres XLogReadRecord + 0xa1
6    0x557bdb5e7767 postgres RecoverPreparedTransactions + 0xd7
7    0x557bdb5f608b postgres StartupXLOG + 0x2a3b
8    0x557bdb870a89 postgres StartupProcessMain + 0x139
9    0x557bdb62f489 postgres AuxiliaryProcessMain + 0x549
10   0x557bdb86d275 postgres <symbol not found> + 0xdb86d275
11   0x557bdb8704e3 postgres PostmasterMain + 0x1213
12   0x557bdb56a1f7 postgres main + 0x497
13   0x7fded4a61c87 libc.so.6 __libc_start_main + 0xe7
```
PG13 and PG14 have a fix for this bug and the bug itself was never reproduced in earlier versions, so the fix never got backported. The discussion can be found here:
https://www.postgresql.org/message-id/flat/743b9b45a2d4013bd90b6a5cba8d6faeb717ee34.camel%40cybertec.at

In a few words, StartupXLOG() renames the last wal segment to .partial but tries to read it by the old name later in
RecoverPreparedTransactions().
It was a bit hard to reproduce it in GP cuz in PG 9.4 XLogRead() cached the most recent file`s descriptor. So it rarely actually reopens the renamed file.

But under high load we often get unfinished 2pc transactions in multiple xlog segments. In this case we miss the cached FD, try to reopen wal-file by the old name and PANIC. Leaving cluster in double-fault mode.

The fix is mostly borrowed from PG14 https://github.com/postgres/postgres/commit/f663b009189ff175a8d134a6cc6e26b5940a1449 with some GP-related excpetions. Also added a regression test which segfaults without this fix.

- [ ] Document changes

### DIFF
--- a/src/backend/access/transam/xlog.c
+++ b/src/backend/access/transam/xlog.c
@@ -7967,6 +7967,64 @@ StartupXLOG(void)
 								   true);
 	}
 
+	/*
+	 * Preallocate additional log files, if wanted.
+	 */
+	PreallocXlogFiles(EndOfLog);
+
+	/*
+	 * Okay, we're officially UP.
+	 */
+	InRecovery = false;
+
+	/* start the archive_timeout timer running */
+	XLogCtl->lastSegSwitchTime = (pg_time_t) time(NULL);
+
+	/* also initialize latestCompletedXid, to nextXid - 1 */
+	LWLockAcquire(ProcArrayLock, LW_EXCLUSIVE);
+	ShmemVariableCache->latestCompletedXid = ShmemVariableCache->nextXid;
+	TransactionIdRetreat(ShmemVariableCache->latestCompletedXid);
+	elog(LOG, "latest completed transaction id is %u and next transaction id is %u",
+		ShmemVariableCache->latestCompletedXid,
+		ShmemVariableCache->nextXid);
+	LWLockRelease(ProcArrayLock);
+
+	/*
+	 * Start up the commit log and subtrans, if not already done for hot
+	 * standby.
+	 */
+	if (standbyState == STANDBY_DISABLED)
+	{
+		StartupCLOG();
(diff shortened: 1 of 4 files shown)


Smyatkin-Maxim:
Hi, warehouse-pg team,
That`s basically a copy of my PR [here](https://github.com/open-gpdb/gpdb/pull/197) which solves FTS double fault issue for clusters with high load. I'm almost 100% sure that the problem is also present in the main branch as it is based on PG12, but for whatever reason I wasn't able to run isolation2 tests in main.
If you run the test I've provided w/o the fix in xlog.c, you will get failure to promote mirror and FTS double failure.
The change differs a bit from a similar fix in PG14, so please double-check that I didn't break anything reordering startup logic.

linwen:
Thank for your contribution! We will review and test it deeply.

adam8157:
I reproduced it on 7X without the fix, it PANICed and failed the promotion. The RCA also looks quite reliable.

Issue #13, opened 2025-04-25
COPY PROGRAM ON SEGMENT fails in fork()


adam8157:
```
"ERROR","53000","can not start command: cat /backup/data1/gpbackup/gpadmin/gpseg6/backups/20250301/20250301090103/gpbackup_6_20250301090103_80132.zst | zstd --decompress -c",,,,,,"COPY XXX FROM PROGRAM 'cat /backup/data1/gpbackup/gpadmin/gpseg<SEGID>/backups/20250301/20250301090103/gpbackup_<SEGID>_20250301090103_80132.zst | zstd --decompress -c' WITH CSV DELIMITER ',' ON SEGMENT;",0,,"copy.c",7993,
"LOG","00000","An exception was encountered during the execution of statement: COPY XXX FROM PROGRAM 'cat /backup/data1/gpbackup/gpadmin/gpseg<SEGID>/backups/20250301/20250301090103/gpbackup_<SEGID>_20250301090103_80132.zst | zstd --decompress -c' WITH CSV DELIMITER ',' ON SEGMENT;",,,,,,"COPY XXX FROM PROGRAM 'cat /backup/data1/gpbackup/gpadmin/gpseg<SEGID>/backups/20250301/20250301090103/gpbackup_<SEGID>_20250301090103_80132.zst | zstd --decompress -c' WITH CSV DELIMITER ',' ON SEGMENT;",0,,"elog.c",2174,
```

That program string works fine in a standard shell. It's the fork() failed by sanity check, but unfortunately we don't print the errno.

![Image](https://github.com/user-attachments/assets/c2ef5eca-662a-41a7-9269-aa47f0123dbc)

PR #15, opened 2025-05-12
Print the errno of COPY program pipes

adam8157:
We observed the COPY PROGRAM failed at the fork() in a corner case but
the errno was not printed, this commit will help a bit.

And free the structure after closed.

Fix https://github.com/warehouse-pg/warehouse-pg/issues/13

### DIFF
--- a/src/backend/commands/copy.c
+++ b/src/backend/commands/copy.c
@@ -19,6 +19,7 @@
 #include "libpq-int.h"
 
 #include <ctype.h>
+#include <string.h>
 #include <unistd.h>
 #include <sys/stat.h>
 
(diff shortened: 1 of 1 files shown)


src/backend/commands/copy.c
```diff
@@ -8080,6 +8081,9 @@ close_program_pipes(CopyState cstate, bool ifThrow)
 	
 	ret = pclose_with_stderr(cstate->program_pipes->pid, cstate->program_pipes->pipes, &sinfo);
 
+	pfree(cstate->program_pipes);
```
linwen, line 8084: Do we need to free the memory which shexec points to before free program_pipes?

adam8157: The memory is controlled by the context, which is likely fine.

My changes here are mainly based on programming habits rather than real issues, because the function checks whether `program_pipes` is null. It’s better to clean it up and set it to null after `pclose`, just like how this function checks and handles `copy_file`.

PR #16, opened 2025-05-12
Avoid harmless fsync errors for removed parent directories

adam8157:
In most cases, the following errors are harmless: they occur when the reported
path is the parent directory of a file/directory we’re removing, and itself is
also expected to be removed.

This commit refactors the code to suppress these errors and ensures that we
fsync each such parent directory only once.

```
pg_rewind: error: could not open file "pg_replslot/internal_wal_replication_slot": No such file or directory
pg_rewind: error: could not open file "errlog": No such file or directory
pg_rewind: error: could not open file "errlog": No such file or directory
pg_rewind: error: could not open file "errlog": No such file or directory
```


adam8157:
These errors were mostly harmless, and this change is covered by our regression tests, the pg_rewind log in gpAdminlogs is clean now with it.

src/bin/pg_rewind/pg_rewind.c
```diff
@@ -892,11 +894,40 @@ syncTargetDirectory(void)
...
+					 *
+					 * filemap->array is sorted by path, so we can check if the
+					 * parent directory is the same as the last one we fsynced.
 					 */
-					fsync_parent_path(entry->path);
+					static char *last_fsynced_parent = NULL;
+					if (!last_fsynced_parent || strcmp(last_fsynced_parent, parentpath) != 0)
+					{
+						if (last_fsynced_parent)
+							pg_free(last_fsynced_parent);
```
huiliang-liu, line 926: I think we should free the last `last_fsynced_parent` before returning too. Otherwise, if the parentpath is `.`, it is hard to make sure it is same for different calling.

huiliang-liu (approved):

PR #18, opened 2025-05-14
Fix gpcheckcat mix_distribution_policy test

adam8157:
We intentionally created such a database during testing and didn’t drop
it. However, accidentally, gpcheckcat only printed the error without
returning an error code. This commit fixes both behaviors.

### DIFF
--- a/gpMgmt/bin/gpcheckcat
+++ b/gpMgmt/bin/gpcheckcat
@@ -2093,8 +2093,8 @@ where
             logger.warning('Unable to generate verify file for {}'.format(filename))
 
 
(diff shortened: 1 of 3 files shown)


linwen (approved):
I am ok with adding return code in the check function and the test case change. 
Better to refine the error message. "both type tables" "use legacy or modern hashops"(actually both hashops are used in the database) is kind of ambiguous to me. 
how about: 
Found tables using legacy hashops and tables using modern hashops for their distribution policies within the database.
or 
Found two types of tables within the database, some use legacy hashops, others use modern hashops as their distribution policies.

PR #27, opened 2025-07-31
Do recursive calls in the plan walker for the MergeAppend

adam8157:
When the planner decides to merge two sorted sub-plans, one of which has a
Share Input Scan node, the executor fails because of wrongly aligned internal
structures.

It was forgotten to do a proper recursive call in the shareinput tree walker.

Signed-off-by: Adam Lee <adam.lee@enterprisedb.com>


linwen (approved):
Looks good to me. The legacy planner doesn't deal with MergeAppend node for shareinputscan. 
MergeAppend
    ├── SubqueryScan
    └── SubqueryScan
This query fails with "cannot execute alien Share Input Scan"

ORCA fallbacks to legacy planner in this case because Multiple Distinct Qualified Aggregates are disabled in ORCA.

PR #28, opened 2025-08-04
Ensure error detection reliability in the status_check case

adam8157:
We cannot guarantee that errors on non-root slices will always surface
in a predictable way. Sometimes the error occurs right at the beginning,
other times it's caught later by GP_PARALLEL_RETRIEVE_CURSOR_CHECK.

Instead of relying on such timing, we now simply check whether the error occurs.


adam8157:
Additional Reading: execution is in a pipeline mode. The original test expected that when DECLARE starts running, it wouldn’t hit that negative value yet, and only after ten seconds would it throw an error, to verify that the timer was working. But that’s hard to guarantee—not just because of timing, but also due to various constraints like each node’s buffer, the bucket size of the hash aggregate, and the random distribution/order of reads.

So I just check that the error can be observed at all. The original issue was simply that “no error was thrown.”

PR #33, opened 2025-08-13
Bye bye PYTHONPATH

beeender:
`PYTHONPATH` has been used from the beginning of gpdb to setup the python libs for management tools. By source the `greenplum_path.sh`, all the sub-processes will see the `PYTHONPATH`, and it pollutes all the python environments in the sub-processes, including the plpython, pgml, development and etc..

From GP7, the 3rd party libs have been removed which makes `gppylib` a bit cleaner. But it is still a pain in the ass which constantly gives users surprises, e.g.: management tools cannot be called in the plpython UDF since the `PYTHONPATH` was remove from there one purpose.

The python code here is a bit messy, this commit doesn't make a huge improvements. It just removes the ENV by define some principles:

- Python interpreter will try to look for libs in the existence directory of the given python source file.
- `python -c` uses the current working directory.
- All the python executables files should exist in `$GPHOME/lib/python` and make symbolics links in the `bin`/`sbin` directories so they can find `gppylib`.
- When calling a remote python statement, use `PyCommand` which will set the CDW to `GPHOME/lib/python` first to make sure `sys.path` has the right value.
- `greenplum_path.sh` won't touch the ENV anymore, if user still want to set it, then it is user's choice.

P.S.:
- In theory, no python files in the `gppylib` should be executed directly. But no idea why do they have `#!python` in the header.
- Since the ENV is gone, no need to unset it in plpython anymore.
- Some `PYTHONPATH` are not removed like those in the test files. They don't seem to be important, remove them later.

- [ ] Document changes

### DIFF
--- a/gpAux/client/scripts/greenplum_clients_path.sh
+++ b/gpAux/client/scripts/greenplum_clients_path.sh
@@ -23,14 +23,12 @@ else
 fi
 
 PATH=${GPHOME_CLIENTS}/bin:${PATH}
-PYTHONPATH=${GPHOME_CLIENTS}/bin/ext:${PYTHONPATH}
 
 # Export GPHOME_LOADERS for GPDB5 compatible
 GPHOME_LOADERS=${GPHOME_CLIENTS}
 export GPHOME_CLIENTS
 export GPHOME_LOADERS
 export PATH
-export PYTHONPATH
 
 # Mac OSX uses a different library path variable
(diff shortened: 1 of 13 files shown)


hyongtao-code (approved):
There is no bug in this PR.

beeender:
> There is no bug in this PR.

No bug in any of my PRs.

PR #35, opened 2025-08-14
Remove gsutil from gpMgmt/requirements-dev.txt

adam8157:
It's only used by concourse CI scripts, and the latest version of it
could not be installed on old operating systems.


linwen (approved):
LGTM.

PR #32, opened 2025-08-07
Using int64 for waitGxids

interma:
This PR fixed #19, and includes two commits:

1. found some compilation warnings of my env, e.g.
```
cdbcat.c: In function ‘GpPolicyFetch’: 
cdbcat.c:485:85: error: ‘distopclasses’ may be used uninitialized [-Werror=maybe-uninitialized] 
485 | policy->opclasses[i] = distopclasses->values[i]; 
| ~~~~~~~~~~~~~~~~~~~^ 
cdbcat.c:424:29: note: ‘distopclasses’ was declared here 
```
My gcc version is `13.3.0 (Ubuntu 13.3.0-6ubuntu2~24.04)`

2. using int64 for waitGxids (as I did it previously)

Glad to contribute to GP related project again, thanks friends.

## Here are some reminders before you submit the pull request
- [ ] Add tests for the change
- [ ] Document changes
- [ ] Communicate in the mailing list if needed
- [x] Pass `make installcheck`
- [ ] Review a PR in return to support the community

PR #37, opened 2025-08-19
The return value of gddCtxGetMaxVid() should be DistributedTransactionId

interma:
The function prototype: `static int gddCtxGetMaxVid(GddCtx *ctx);` but it uses `DistributedTransactionId` value as its return value: https://github.com/warehouse-pg/warehouse-pg/blob/deb5b5d218a0c2aa9f2febf4ca3dd36e94afc902/src/backend/utils/gdd/gdddetector.c#L276

So a "narrowing conversion" happens here. 

When your cluster enables GDD and runs a long time, the GXID may exceed INT_MAX, then it will cause an **infinite loop** here: `maxvid` will be cut down into an invalid int value and add it into `vids` list later.
https://github.com/warehouse-pg/warehouse-pg/blob/deb5b5d218a0c2aa9f2febf4ca3dd36e94afc902/src/backend/utils/gdd/gdddetector.c#L151-L168

As the result: the `vids` list will continue to grow (until an OOM happens). You will find gdd's memory context is very huge in this scenario.

This PR fixes it (with a simple unittest).

- [ ] Document changes

### DIFF
--- a/src/backend/utils/gdd/gdddetector.c
+++ b/src/backend/utils/gdd/gdddetector.c
@@ -22,7 +22,7 @@
 static GddGraph *gddCtxGetGraph(GddCtx *ctx, int segid);
 static GddStat *gddCtxGetGlobalStat(GddCtx *ctx, DistributedTransactionId vid);
 static void gddCtxRemoveVid(GddCtx *ctx, DistributedTransactionId vid);
-static int gddCtxGetMaxVid(GddCtx *ctx);
+static DistributedTransactionId gddCtxGetMaxVid(GddCtx *ctx);
 
 static GddStat *gddStatNew(DistributedTransactionId vid);
 static void gddStatInit(GddStat *stat, DistributedTransactionId vid);
@@ -251,7 +251,7 @@ gddCtxRemoveVid(GddCtx *ctx, DistributedTransactionId vid)
 /*
  * Get the max vert id, return 0 if no vert left at all.
  */
-static int
+static DistributedTransactionId
 gddCtxGetMaxVid(GddCtx *ctx)
 {
 	GddMapIter	graphiter;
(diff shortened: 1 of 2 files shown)


interma:
More further: I think we should check if have similar issues. (e.g. my previous PR #32). 
These issues are easy to be checked by tools but it's very hard to do debug in user env.

Two common solutions:
1. enable `-Wconversion` in gcc:
```
$ gcc -Wall -Wextra -Wconversion -Wmissing-prototypes -Wpointer-arith -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-unused-but-set-variable -Werror=implicit-fallthrough=3 -Wno-format-truncation -Wno-stringop-truncation -g -ggdb -Og -g3 -fno-omit-frame-pointer  -Werror=uninitialized -Werror=implicit-function-declaration -I../../../../src/interfaces/libpq -I../../../../src/include   -D_GNU_SOURCE   -c -o gdddetector.o gdddetector.c

gdddetector.c: In function ‘gddCtxGetMaxVid’:
gdddetector.c:276:16: warning: conversion from ‘DistributedTransactionId’ {aka ‘long unsigned int’} to ‘int’ may change value [-Wconversion]
  276 |         return maxvid;
```
It can detect the issue, but it gives more noisy warnings. (maybe that's the reason why it's disable defaultly)

2. using clang-tidy (indeed I got this solution in chatggpt)
```
$ clang-tidy gdddetector.c -checks=-*,hicpp-narrowing-conversions,cppcoreguidelines-narrowing-conversions -- -I.

gdddetector.c:7:12: warning: narrowing conversion from 'DistributedTransactionId' (aka 'unsigned long') to 'int' [-hicpp-narrowing-conversions]
    return maxvid;
           ^
```
It gives less noise, so I think it's a better choise.

SmartKeyerror (approved):
Looks good to me, and I'll check if there has other same problem.

PR #38, opened 2025-08-19
Fix flaky tests of REINDEX TABLE/INDEX

adam8157:
These two test cases had two issues, in addition to mismatched comments:

When performing a REINDEX operation on a partitioned table, the partitions are
locked and their indexes are rebuilt sequentially, not all at once.

When two REINDEX commands are executed concurrently in the background, it is
not deterministic which command will first rebuild the index on the "de_fault"
partition in the example.

Prior to the commit referenced below, REINDEX TABLE could be placed within an
explicit transaction. This made it plausible that both commands would run
concurrently, and one could observe the index being rebuilt twice. However,
after that change, the test became flaky for the two issues mentioned above,
as the double rebuild was no longer consistently observable.

	commit 67a7c0f06be0da1cd466e24fb0c33b51308dc1e0
	Author: (Jerome)Junfeng Yang <yjerome@vmware.com>
	Date:   Fri Aug 20 15:37:01 2021 +0800

	    Add support for partitioned tables and indexes in REINDEX

This commit addresses the flakiness by running the two commands in a
roughly concurrent manner and only asserting that the index was
successfully rebuilt, without checking the number of times it occurred.

### DIFF
--- a/src/test/isolation2/expected/reindex/reindextable_while_reindex_idx_ao_part_btree.out
+++ b/src/test/isolation2/expected/reindex/reindextable_while_reindex_idx_ao_part_btree.out
@@ -9,7 +9,6 @@ insert into reindex_crtab_part_ao_btree (id, owner, description, property, poli,
 INSERT 0 1000
 create index on reindex_crtab_part_ao_btree(id);
 CREATE INDEX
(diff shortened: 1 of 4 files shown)


linwen (approved):
LGTM.

huiliang-liu (approved):
Good catch!

PR #39, opened 2025-08-20
Fix the flaky explain_format case

adam8157:
```
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /home/runner/gpdb_src/src/test/regress/expected/explain_format.out /home/runner/gpdb_src/src/test/regress/results/explain_format.out
--- /home/runner/gpdb_src/src/test/regress/expected/explain_format.out	2025-08-19 12:09:49.057298453 +0000
+++ /home/runner/gpdb_src/src/test/regress/results/explain_format.out	2025-08-19 12:09:49.079298391 +0000
@@ -874,6 +874,7 @@
 Gather Motion N:N  (sliceN; segments: N) (actual time=N.N..N.N rows=N loops=N)
   ->  Hash Join (actual time=N.N..N.N rows=N loops=N)
         Hash Cond: (s2.b = s1.b)
+        Extra Text: (segN)   Hash chain length N.N avg, N max, using N of N buckets.
         ->  Seq Scan on stat_io_timing s2 (never executed)
         ->  Hash (actual time=N.N..N.N rows=N loops=N)
               Buckets: N  Batches: N  Memory Usage: NkB
@@ -887,7 +888,7 @@
   (sliceN)    Executor memory: NK bytes (segN).
 Memory used:  NkB
 Execution Time: N.N ms
-(16 rows)
+(17 rows)
 -- Test Bitmap Heap Scan block accounting
 SET enable_seqscan = 0;
 SET enable_bitmapscan = 1;
```

For instance in the Hash Join node, `Extra Text` is only printed when a
spill occurs, which heavily depends on the system state. Simply lowering
`statement_mem` does not reliably trigger a spill when the data set is
small.

So instead of checking `Extra Text` output in this case, we can skip
it, since we already have sufficient coverage for it in other tests.

### DIFF
--- a/src/test/regress/expected/explain_format.out
+++ b/src/test/regress/expected/explain_format.out
@@ -41,6 +41,9 @@ begin
         -- Ignore text-mode "Planning:" line because whether it's output
         -- varies depending on the system state
         CONTINUE WHEN (ln = 'Planning:');
+        -- Ignore text-mode "Extra Text:" line because whether it's output
+        -- varies depending on the system state
+        CONTINUE WHEN (ln ~ 'Extra Text:');
         return next ln;
     end loop;
 end;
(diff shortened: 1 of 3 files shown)


adam8157:
Commit e7c2b1f was also intended to address a similar issue, but it did so in a misguided way: it skipped the necessary EXPLAIN output checks and instead only verified the Extra Text lines that were guaranteed to appear. With this fix, the original problem no longer exists, so we are reverting that commit.

linwen (approved):
looks good to me.

PR #42, opened 2025-08-28
Fix CVE-2025-1094 on WHPG 6

adam8157:
https://www.postgresql.org/support/security/CVE-2025-1094/

It affects WHPG 6 (based on PostgreSQL 9.4) too, but the above page doesn't list it because PG 9.4 is EOL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ build.sh
 .project
 .settings/
 .vscode/
+.cache/
 GPATH
 GRTAGS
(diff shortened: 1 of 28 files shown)


adam8157:
Screenshot 2025-08-29 at 16 07 30

linwen (approved):
enhances security by improving how to handle and escape invalid multibyte character encodings in frontend tools and libraries.
Looks good to me.

PR #46, opened 2025-09-07
Enable Parallel Workers Context for Index Builds

linwen:
This patch is to enable parallel worker context for index creation by correctly handling the process roles and GUCs within the parallel worker's lifecycle. Third-party extension like pgvector index creation can leverage it. Built-in index like BTree index creation won't be affected by this change.

The key changes are as follows:
1. Set Gp_role and pass gp_session_id to parallel workers, otherwise when initialize bgworker processes, an error "connections to primary segments are not allowed" occurs;
2. Pass numsegmentsFromQD to parallel workers, GpPolicyFetch needs this information;
3. Set Gp_is_writer to false, since parallel workers behave like reader QEs;
4. Skip motion layer setup for parallel workers.

### DIFF
--- a/src/backend/access/transam/parallel.c
+++ b/src/backend/access/transam/parallel.c
@@ -42,6 +42,8 @@
 #include "utils/relmapper.h"
 #include "utils/snapmgr.h"
 #include "utils/typcache.h"
+#include "cdb/cdbvars.h"
+#include "cdb/cdbutil.h"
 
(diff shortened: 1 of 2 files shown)


adam8157 (approved):
All built-in index types have parallel workers disabled, so starting to explore parallel performance from third-party extensions is quite safe.

Issue #49, opened 2025-09-09
compression failed: Destination buffer is too small


adam8157:
## Bug Report
```
"LOG","00000","server process (PID 3621904) was terminated by signal 6: Aborted","Failed process was running: DELETE FROM **** ",,,,,,0,,"postmaster.c",4184,
...
"PANIC","XX000","compression failed: Destination buffer is too small uncompressed len 32760 (xloginsert.c:865)",,,,,,"DELETE FROM **** ",0,,"xloginsert.c",865,"Stack trace:
1 0xcefbac postgres errstart + 0x2cc
2 0x7b20b3 postgres <symbol not found> + 0x7b20b3
3 0x7b244b postgres <symbol not found> + 0x7b244b
4 0x7b2dc2 postgres XLogInsert + 0xa2
5 0x749728 postgres heap_delete + 0x8d8
6 0x74999a postgres simple_heap_delete + 0x3a
7 0x758994 postgres <symbol not found> + 0x758994
8 0x759561 postgres toast_delete + 0x91
9 0x7494e0 postgres heap_delete + 0x690
10 0x987572 postgres <symbol not found> + 0x987572
11 0x98959a postgres <symbol not found> + 0x98959a
12 0x95e047 postgres <symbol not found> + 0x95e047
13 0x954afb postgres <symbol not found> + 0x954afb
14 0x95555f postgres standard_ExecutorRun + 0x37f
15 0x955765 postgres ExecutorRun + 0xc5
16 0xb6f962 postgres <symbol not found> + 0xb6f962
17 0xb6fe8b postgres <symbol not found> + 0xb6fe8b
18 0xb710b0 postgres PortalRun + 0x170
19 0xb6a5b6 postgres <symbol not found> + 0xb6a5b6
20 0xb6e5fc postgres PostgresMain + 0x262c
21 0xacd4fb postgres <symbol not found> + 0xacd4fb
22 0xace4e7 postgres PostmasterMain + 0xd57
23 0x6ed5c4 postgres main + 0x494
24 0x7fad590897e5 libc.so.6 __libc_start_main + 0xe5
25 0x6f943e postgres _start + 0x2e
"
```

### WarehousePG version or build
7.2.1-WHPG

PR #50, opened 2025-09-09
Insert WAL record uncompressed if compression fails

adam8157:
We observed errors such as: "compression failed: Destination buffer is too
small uncompressed len 32760 (xloginsert.c:865)"

This typically occurs when the "compressed" output is larger than the original
input. Regardless of the exact cause, it is always safe to insert the
uncompressed WAL record instead of failing.

### DIFF
--- a/src/backend/access/transam/xloginsert.c
+++ b/src/backend/access/transam/xloginsert.c
@@ -30,6 +30,7 @@
 #include "replication/origin.h"
 #include "storage/bufmgr.h"
 #include "storage/proc.h"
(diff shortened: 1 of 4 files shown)


adam8157:
Fixes #49 

I'm trying to write a reproducer as the test case, but no luck yet. I don’t think fault injection is really necessary, since the issue was just a small oversight, and the current code and behavior are consistent with upstream PostgreSQL.

adam8157:
It’s much simpler to write a fault injection test, so I added another commit.

PR #52, opened 2025-09-12
Fix extra blank lines in EXPLAIN

adam8157:
On CI we observed that EXPLAIN produces blank lines, sometimes it produces multiple blank lines causing instability. The root cause was an earlier commit that printed an extra newline. When multiple "Extra Text" lines are present, this results in multiple blank lines.

That commit also did not use the `cdbexplainfun` callback as other nodes, making the issue harder to trace. This patch fixes both problems.

```
  ->  HashAggregate  (cost=N.N..N.N rows=N width=N) (actual time=N.N..N.N rows=N loops=N)
        Group Key: share0_ref2.b

+
        ->  Redistribute Motion N:N  (sliceN; segments: N)  (cost=N.N..N.N rows=N width=N) (actual time=N.N..N.N rows=N loops=N)
              Hash Key: share0_ref2.b
              ->  Result  (cost=N.N..N.N rows=N width=N) (actual time=N.N..N.N rows=N loops=N)
```

```
regression=# EXPLAIN ANALYZE SELECT a, avg(b) AS b FROM test_src_tbl GROUP BY grouping sets ((a), (b));
                                                                                                      QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1299.51 rows=19980 width=12) (actual time=12.980..24.696 rows=20000 loops=1)
   ->  Sequence  (cost=0.00..1298.62 rows=6660 width=12) (actual time=13.764..24.940 rows=6771 loops=1)
         ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.61 rows=16667 width=1) (actual time=0.000..3.632 rows=0 loops=1)
               ->  Seq Scan on test_src_tbl  (cost=0.00..431.35 rows=16667 width=8) (actual time=0.033..1.232 rows=16930 loops=1)
         ->  Append  (cost=0.00..866.93 rows=6660 width=12) (actual time=10.380..21.208 rows=6771 loops=1)
               ->  HashAggregate  (cost=0.00..433.40 rows=3330 width=8) (actual time=10.378..13.653 rows=3385 loops=1)
                     Group Key: share0_ref2.b
                     Extra Text: (seg1)   hash table(s): 1; 3385 groups total in 15 batches, 53496 spill partitions; disk usage: 1152KB; chain length 2.1 avg, 4 max; using 3385 of 32768 buckets; total 0 expansions.

                     Extra Text: (seg2)   hash table(s): 1; 3247 groups total in 8 batches, 50728 spill partitions; disk usage: 1152KB; chain length 2.2 avg, 4 max; using 3247 of 18432 buckets; total 0 expansions.

                     ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.37 rows=16667 width=4) (actual time=0.854..6.850 rows=16925 loops=1)
                     ...
```

	commit c22b4a3558666654efb674b4f2d4686a981298fb
	Author: Yao Wang <wayao@vmware.com>
	Date:   Thu Aug 24 15:36:41 2023 +0800

	    Add explain analyze detailed info of hash agg (#15917)

### DIFF
--- a/src/backend/executor/nodeAgg.c
+++ b/src/backend/executor/nodeAgg.c
@@ -3657,6 +3657,8 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
 	{
 		/* Allocate string buffer. */
 		aggstate->ss.ps.cdbexplainbuf = makeStringInfo();
+
+		/* Request a callback at end of query. */
 		aggstate->ss.ps.cdbexplainfun = ExecAggExplainEnd;
 	}
 
@@ -5262,8 +5264,8 @@ ReuseHashTable(AggState *node)
  * ExecAggExplainEnd
  *      Called before ExecutorEnd to finish EXPLAIN ANALYZE reporting.
  *
- * Check cdbexplain_depositStatsToNode(), Greenplum only saves extra
- * message text for the most interesting winning qExecs.
+ * Check cdbexplain_depositStatsToNode(), it only saves extra extra text for
+ * the most interesting winning qExecs.
  */
 static void
 ExecAggExplainEnd(PlanState *planstate, struct StringInfoData *buf)
--- a/src/backend/executor/nodeAgg.c
+++ b/src/backend/executor/nodeAgg.c
@@ -489,7 +489,7 @@ static int	find_compatible_pertrans(AggState *aggstate, Aggref *newagg,
 
 static void ExecEagerFreeAgg(AggState *node);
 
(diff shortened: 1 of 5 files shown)


linwen (approved):
Use hook function ExecAggExplainEnd to display explain info, just like ExecSortExplainEnd, ExecShareInputScanExplainEnd etc.
Looks good to me.

PR #54, opened 2025-09-17
Fix mirror promotion failure due to WAL segment rename

adam8157:
When promoting a mirror after failover, we sometimes hit a fatal error:

```
"FATAL","58P01","requested WAL segment pg_xlog/00000023000071D50000001F has already been removed",,,,,,,0,,"xlogutils.c",580,"Stack
trace:
1    0x557bdb9f09b6 postgres errstart + 0x236
2    0x557bdb5fc6cf postgres <symbol not found> + 0xdb5fc6cf
3    0x557bdb5fd021 postgres read_local_xlog_page + 0x191
4    0x557bdb5fb922 postgres <symbol not found> + 0xdb5fb922
5    0x557bdb5fba11 postgres XLogReadRecord + 0xa1
6    0x557bdb5e7767 postgres RecoverPreparedTransactions + 0xd7
7    0x557bdb5f608b postgres StartupXLOG + 0x2a3b
8    0x557bdb870a89 postgres StartupProcessMain + 0x139
9    0x557bdb62f489 postgres AuxiliaryProcessMain + 0x549
10   0x557bdb86d275 postgres <symbol not found> + 0xdb86d275
11   0x557bdb8704e3 postgres PostmasterMain + 0x1213
12   0x557bdb56a1f7 postgres main + 0x497
13   0x7fded4a61c87 libc.so.6 __libc_start_main + 0xe7
```

Upstream PostgreSQL fixed this in 13 and 14, but the issue never appeared in earlier releases, so it was not backpatched. Discussion here: https://www.postgresql.org/message-id/flat/743b9b45a2d4013bd90b6a5cba8d6faeb717ee34.camel%40cybertec.at

The root cause is that `StartupXLOG()` renames the last WAL segment to .partial but later `RecoverPreparedTransactions()` attempts to read it using the old name.

In PostgreSQL earlier versions, this rarely reproduces because XLogRead() caches the most recent file descriptor, so the renamed file is seldom reopened.

In WarehousePG/Greenplum, however, high load often leaves unfinished 2PC transactions spread across multiple WAL segments. This causes the cached FD to be bypassed, forcing a reopen by the old name, which results in a PANIC and leaves the cluster in double-fault mode.

This patch borrows the upstream fix from 14 (f663b009), with some specific adjustments. It also adds a regression test that reliably segfaults without the fix.

### DIFF
--- a/src/backend/access/transam/xlog.c
+++ b/src/backend/access/transam/xlog.c
@@ -8130,6 +8130,73 @@ StartupXLOG(void)
 			CreateCheckPoint(CHECKPOINT_END_OF_RECOVERY | CHECKPOINT_IMMEDIATE);
 	}
 
+	/*
+	 * Preallocate additional log files, if wanted.
+	 */
+	PreallocXlogFiles(EndOfLog);
+
+	/*
+	 * Okay, we're officially UP.
+	 */
+	InRecovery = false;
+
+	SIMPLE_FAULT_INJECTOR("out_of_recovery_in_startupxlog");
+
+	/* start the archive_timeout timer and LSN running */
+	XLogCtl->lastSegSwitchTime = (pg_time_t) time(NULL);
+	XLogCtl->lastSegSwitchLSN = EndOfLog;
+
+	/* also initialize latestCompletedXid, to nextXid - 1 */
+	LWLockAcquire(ProcArrayLock, LW_EXCLUSIVE);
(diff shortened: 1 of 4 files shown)


adam8157:
This is the main branch version https://github.com/warehouse-pg/warehouse-pg/pull/14, thanks to @Smyatkin-Maxim

linwen (approved):
This is a valid fix from upstream with well-written test case. Looks good to me.
@Smyatkin-Maxim Thanks for your contribution to WarehousePG!

Issue #53, opened 2025-09-16
Wrong query results in EXPR SUBLINK


SmartKeyerror:
### reproduce steps

```sql
create table foo (a int, b int);
create table bar (a int, b int);
insert into foo select i, i+1 from generate_series(1,20) i;
insert into bar select i%6+1, i+2 from generate_series(1,20) i;
```

### query results

In PostgreSQL 17, below query will return 17 rows:

```sql
-- postgresql 17
select count(*) from foo x where x.a > (select count(*) from bar y where y.a = x.b);
 count 
-------
    17
(1 row)

select count(*) from foo x where x.a > (select regr_count(a,b) from bar y where y.a = x.b);
 count 
-------
    17
(1 row)
```

However, in warehouse-pg 7, it will return just 2 rows:

```sql
-- warehouse-pg 7
select count(*) from foo x where x.a > (select count(*) from bar y where y.a = x.b);
 count 
-------
     2
(1 row)

select count(*) from foo x where x.a > (select regr_count(a,b) from bar y where y.a = x.b);
 count 
-------
     2
(1 row)
```

### query plan

```sql
-- warehouse-pg 7
explain select count(*) from foo x where x.a > (select count(*) from bar y where y.a = x.b);
                                                     QUERY PLAN                                                      
---------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=1013.04..1013.05 rows=1 width=8)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=1012.98..1013.03 rows=3 width=8)
         ->  Partial Aggregate  (cost=1012.98..1012.99 rows=1 width=8)
               ->  Hash Join  (cost=497.00..989.07 rows=9567 width=0)
                     Hash Cond: (x.b = "Expr_SUBQUERY".csq_c0)
                     Join Filter: (x.a > "Expr_SUBQUERY".csq_c1)
                     ->  Seq Scan on foo x  (cost=0.00..321.00 rows=28700 width=8)
                     ->  Hash  (cost=484.50..484.50 rows=1000 width=12)
                           ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=464.50..484.50 rows=1000 width=12)
                                 ->  Subquery Scan on "Expr_SUBQUERY"  (cost=464.50..471.17 rows=333 width=12)
                                       ->  HashAggregate  (cost=464.50..467.83 rows=333 width=12)
                                             Group Key: y.a
[...]

PR #56, opened 2025-09-30
Fix incorrect join type when pulling up EXPR SubLink

SmartKeyerror:
### Context

This PR is trying to fix the issue of #53 

Currently, warehouse-pg will try to pull up EXPR SubLink by rewiring it to 
subqueries with inner join, for example:

```sql
create table foo (a int, b int);
create table bar (x int, y int);

explain (costs off)
select * from foo where a > (select count(*) from bar where foo.a = bar.x);
                      QUERY PLAN                       
-------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)
   ->  Hash Join
         Hash Cond: (foo.a = "Expr_SUBQUERY".csq_c0)
         Join Filter: (foo.a > "Expr_SUBQUERY".csq_c1)
         ->  Seq Scan on foo
         ->  Hash
               ->  Subquery Scan on "Expr_SUBQUERY"
                     ->  HashAggregate
                           Group Key: bar.x
                           Filter: (bar.x > count(*))
                           ->  Seq Scan on bar
```

This expression SubLink will be transformed into:

```sql
select foo.a, foo.b from foo inner join 
    (select x, count(*) as sub_count from bar group by bar.x) sub
on foo.a = sub.x where a > sub_count;
```

This rewrite is incorrect because it is not semantically equivalent.
In PostgreSQL, `count()` is a strict function, but unlike most strict
functions, it does not return `NULL` when the input is `NULL`; instead,
it returns `0`. As a result, if the inner sublink finds no matching rows,
`count(*)` evaluates to `0`. In contrast, when the query is rewritten
using an inner join, if `foo.a = sub.x` does not match, that row is
eliminated entirely from the output rather than producing a count of `0`,
which introduced the wrong results:

```sql
create table foo (a int, b int);
create table bar (x int, y int);
insert into foo select i,i+1 from generate_series(1,20) i;
insert into bar select i%6+1,i+2 from generate_series(1,20) i;

-- PostgreSQL 17
select count(*) from foo
where a > (select count(*) from bar where foo.a = bar.x);
 count 
-------
    17
(1 row)

-- Warehouse-PG 7
select count(*) from foo
where a > (select count(*) from bar where foo.a = bar.x);
 count 
-------
     3
(1 row)
```

### Correct Translation

The correct approach is to rewrite it using a LEFT JOIN with COALESCE().
LEFT JOIN ensures that all rows from the left table are preserved even if
no matching rows exist in the right table. `COALESCE()` is required because
when a LEFT JOIN has no matches, the right table’s values are filled with
NULL. We must replace these NULLs with the zero value of count(*), which
is why `COALESCE()` is necessary. Without the `COALESCE()` node, the result
would still be incorrect. For example, rewriting the original query with
LEFT JOIN alone will not yield the correct result.

```sql
select count(*) from foo left join 
(select bar.x, count(*) as c from bar group by bar.x) sub
on foo.a = sub.x where foo.a > sub.c;
 count 
-------
     3
(1 row)
```

When we remove the condition `where foo.a > sub.c`, we can see that the
original join data contains NULL values in column `sub.c`. Since the `>`
operator is strict, its output is NULL whenever the input is NULL. As a
result, the filter condition evaluates to false, and the row will not be
emitted in the output:

```sql
select * from foo left join 
(select bar.x, count(*) as c from bar group by bar.x) sub
on foo.a = sub.x;
 a  | b  | x | c 
----+----+---+---
  5 |  6 | 5 | 3
  6 |  7 | 6 | 3
  9 | 10 |   |  
 10 | 11 |   |  
 11 | 12 |   |  
 13 | 14 |   |  
 14 | 15 |   |  
 17 | 18 |   |  
  2 |  3 | 2 | 4
  3 |  4 | 3 | 4
  4 |  5 | 4 | 3
  7 |  8 |   |  
  8 |  9 |   |  
 16 | 17 |   |  
 18 | 19 |   |  
 19 | 20 |   |  
  1 |  2 | 1 | 3
 12 | 13 |   |  
 15 | 16 |   |  
 20 | 21 |   |  
(20 rows)
```

When we add a COALESCE() node, the result becomes correct:

```sql
select count(*) from foo left join 
(select bar.x, count(*) as c from bar group by bar.x) sub
on foo.a = sub.x where foo.a > COALESCE(sub.c, 0);
 count 
-------
    17
(1 row)
```

Therefore, both LEFT JOIN and COALESCE() are indispensable.

### Code changes

Different types of aggregate functions behave differently when pulling up 
EXPR SUBLINK into JOIN:

- Strict aggregate functions (i.e., functions that return NULL when the input is 
NULL) — such as MAX(), SUM(), AVG(), PERCENTILE_CONT(), RANK(), etc. 
Can be safely pulled up and rewritten using an INNER JOIN, since they will naturally 
return NULL if the join condition fails to match.

- Non-strict aggregate functions, such as COUNT(*), are different: even when 
the input is NULL, they return a non-NULL value (typically 0). To preserve semantic 
correctness when pulling these up, the subquery must be rewritten using a LEFT 
JOIN combined with COALESCE() to ensure that NULL cases are properly handled.

In `cdb_pre_process_expr_sublink_agg()`, we use `expression_tree_mutator()`
instead of `expression_tree_walker()` to traverse the expr node. The purpose
is to replace the original Aggref node with a `Const` node, so that
`eval_const_expressions()` can be used to evaluate the constant expression.

For example, consider the query:

```sql
select count(*) from foo
where a > (select (count(*) + 5) * 3 from bar where foo.a = bar.x);
```

The expr in this case is an expression containing an Aggref:
`(count(*) + 5) * 3`. We first recursively enter the Aggref node and use
`calculate_agg_value_for_empty_input()` to compute the default zero value
of `count(*)` at runtime, generating a `Const` node to replace the `Aggref`
node. Then, the expression `(0 + 5) * 3` is evaluated, and the final
`COALESCE()` default value is obtained.

### Reference

- [Orthogonal optimization of subqueries and aggregation](https://sigmodrecord.org/publications/sigmodRecord/0106/pdfs/Orthogonal%20Optimization%20of%20Subqueries%20and%20Aggregation.pdf)
- [Unnesting Arbitrary Queries](https://cs.emis.de/LNI/Proceedings/Proceedings241/383.pdf)


adam8157 (dismissed):
The commit addresses the issue where pulling up a subquery with a non-strict aggregate (like count(*)) as an INNER JOIN would incorrectly filter out rows from the outer table.

Changing the logic to use a LEFT JOIN and wrapping the result in COALESCE preserves all outer rows and correctly handles cases where there are no matching rows in the subquery.

Looks good to me.

PR #57, opened 2025-10-07
Allow parallel workers to get combo cid and distributed snapshot from DSM.

linwen:
When use parallel worker to create hnsw index for heap partitioned tables, pgvector reports an error: ERROR:  Reader qExec tried to allocate a Combo Command Id (combocid.c:241) (combocid.c:241)  (seg0 127.0.0.1:7002 pid=115876) (combocid.c:241)

The reason is, function ParallelWorkerMain calls RestoreComboCIDStateto restore Combo CID from dynamic shared memory segment. Combo CID is put into dynamic shared memory segment by function InitializeParallelDSM as Writer QE. RestoreComboCIDStateto calls GetComboCommandId to build a local hashtable. The function GetComboCommandId is assumed to work for QE writer. So the error happened. 

However in fact, parallel workers(with Gp_role GP_ROLE_EXECUTE,  Gp_is_writer is false) should also be able to get a combo cid, this is how PostgreSQL parallel workers behave. 

The reason why there is error for partitioned heap table and no error for heap table is, 

- for partitioned heap table parallel index creation, the combo cid is generated due to some dealing for parent and child tables, 
- but for non-partitioned heap table, no combo cid is generated. So GetComboCommandId is not called because num_elements is zero. 

We can know a process is a parallel worker or not by MyProc->lockGroupLeader, parallel workers has a lock group leader, which is the QE process who forked them. After a parallel worker is forked, it will call BecomeLockGroupMember to become a member of group locking, with its parent process as leader.

In this PR, we allow parallel workers to get combo cid from dynamic shared memory segment and build a local hashtable for it. But they won't dump combo cid into DSM.

In this PR, I also serialize and deserialize distributed snapshot for parallel workers, so that they can get correct snapshot from QE(their leader).

This PR won't affect the built-in index of WarehousePG. It will take effect on 3rd party index which use parallel workers to build index.

### DIFF
--- a/src/backend/access/transam/parallel.c
+++ b/src/backend/access/transam/parallel.c
@@ -377,7 +377,14 @@ InitializeParallelDSM(ParallelContext *pcxt)
 		SerializeGUCState(guc_len, gucspace);
 		shm_toc_insert(pcxt->toc, PARALLEL_KEY_GUC, gucspace);
 
-		/* CDB: Serialize DtxContextInfo */
+		/*
+		 * CDB: Serialize DtxContextInfo
+		 * QEs use DtxContextInfo to keep distributed transaction consistency with QD.
+		 * QD dispatches DtxContextInfo to QEs, including Writer and Reader QE.
+		 * However, parallel workers can't get dispatched information from QD.
+		 * So the leader QE needs to serialize its DtxContextInfo into DSM.
+		 * When parallel workers are created, they get DtxContextInfo from DSM.
+		 */
 		dtxspace = shm_toc_allocate(pcxt->toc, sizeof(uint32) + dtxlen);
 		*((uint32 *) dtxspace) = (uint32) dtxlen;
 		DtxContextInfo_Serialize(dtxspace + sizeof(uint32), &QEDtxContextInfo);
@@ -1447,7 +1454,11 @@ ParallelWorkerMain(Datum main_arg)
 	uint32 dtx_len = *((uint32 *) dtxspace);
 	DtxContextInfo_Deserialize(dtxspace + sizeof(uint32), (int) dtx_len, &QEDtxContextInfo);
 
-	/* CDB: Set DistributedTransactionContext */
+	/*
+	 * CDB: Set DistributedTransactionContext for Parallel workers.
+	 * Parallel workers behaves as Reader QEs, their DistributedTransactionContext
+	 * needs to be set since it is referenced when check tuple visibility.
+	 */
 	if (QEDtxContextInfo.haveDistributedSnapshot)
 	{
 		if (IS_QUERY_DISPATCHER())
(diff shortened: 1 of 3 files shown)


src/backend/utils/time/combocid.c
```diff
@@ -233,15 +233,23 @@ GetComboCommandId(CommandId cmin, CommandId cmax)
 	ComboCidEntry entry;
 	bool		found;
 
-	if (Gp_role == GP_ROLE_EXECUTE && !Gp_is_writer)
+	/*
+	 * Entrydb and QE reader will report ERROR here.
+	 * Parallel workers can get combo cid from dynamic shared memory segment
+	 * by function RestoreComboCIDState.
+	 */
+	if (Gp_role == GP_ROLE_EXECUTE && !Gp_is_writer && !MyProc->lockGroupLeader)
```
adam8157: So, a parallel worker uses `RestoreComboCIDState()` but not `loadSharedComboCommandIds()` to rebuild/restore Combo IDs.

A suggestion:
Comment more to explain how entrydbs, QE readers, or parallel workers get the Combo ID information.

Question:
In `InitializeParallelDSM()`, where are the Combo IDs to serialize from? I doubt parallel workers have got the same  information as QE readers.

src/backend/utils/time/snapmgr.c
```diff
@@ -2355,6 +2401,18 @@ SerializeSnapshot(Snapshot snapshot, char *start_address)
 		memcpy((TransactionId *) (start_address + subxipoff),
 			   snapshot->subxip, snapshot->subxcnt * sizeof(TransactionId));
 	}
+
+	/* Copy inProgressXidArray in distributed snapshot */
+	if (snapshot->haveDistribSnapshot && serialized_snapshot.ds.count > 0)
+	{
+		char *xidoff = (char *) MAXALIGN((uintptr_t)(start_address +
```
adam8157: We’re only using memcpy to read and write the data, without dereferencing it as a C struct, so MAXALIGN shouldn’t be necessary. The two arrays above also don’t use MAXALIGN.

Also, the variable name `xidoff` isn’t very accurate — how about renaming it to `dxipoff`?

src/backend/access/transam/parallel.c
```diff
@@ -1424,6 +1440,20 @@ ParallelWorkerMain(Datum main_arg)
...
+	uint32 dtx_len  = *((uint32 *) dtxspace);
+	DtxContextInfo_Deserialize(dtxspace + sizeof(uint32), (int) dtx_len, &QEDtxContextInfo);
+
+	/* CDB: Set DistributedTransactionContext */
+	if (QEDtxContextInfo.haveDistributedSnapshot)
+	{
+		if (IS_QUERY_DISPATCHER())
+			setDistributedTransactionContext(DTX_CONTEXT_QE_ENTRY_DB_SINGLETON);
+		else
+			setDistributedTransactionContext(DTX_CONTEXT_QE_READER);
```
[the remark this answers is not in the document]
adam8157: If I understand correctly, parallel workers don’t have any special role in WarehousePG’s distributed architecture (in the context of writer QEs and reader QEs). Regardless of whether they perform write operations, they are technically part of the distributed system but don’t participate in any distributed responsibilities. Therefore, we set their context to the general types DTX_CONTEXT_QE_ENTRY_DB_SINGLETON and DTX_CONTEXT_QE_READER.

I’d suggest adding more comments here to clarify this behavior.

PR #66, opened 2025-11-03
Use LWLock instead of SpinLock at InstrumentationHeader

SmartKeyerror:
### Issue Summary

A crash was observed when acquiring the spinlock `InstrumentGlobal->lock`, 
which was caused by severe lock contention. Further investigation revealed 
deeper structural issues in the locking design.

### Root Cause

In `instrShmemRecycleCallback()`, a large loop performs multiple system calls 
(such as memory reset and free) while holding `InstrumentGlobal->lock`.
This code path can take a significant amount of time under heavy system load, 
preventing other backends from acquiring the same lock.

Moreover, spinlocks must never be held across kernel or blocking calls, as they 
are designed for very short, atomic critical sections. Holding a spinlock while 
performing system calls violates this principle and can easily lead to contention, 
starvation, or deadlocks.

### Resolution

The current use of a spinlock is inappropriate for this scenario.

The fix replaces the spinlock with a lightweight lock (LWLock), which is designed 
for longer critical sections and allows kernel calls within the protected region 
without risking system instability.

### DIFF
--- a/src/include/executor/instrument.h
+++ b/src/include/executor/instrument.h
@@ -125,6 +125,8 @@ typedef struct InstrumentationHeader
 {
 	void	   *head;
 	int			free;
+
+	/* GP_ABI_BUMP_FIXME: deprecated field, keep it for ABI, do not use it */
 	slock_t		lock;
 } InstrumentationHeader;
 
--- a/src/backend/executor/instrument.c
+++ b/src/backend/executor/instrument.c
@@ -26,6 +26,7 @@
 #include "storage/shmem.h"
 #include "cdb/cdbdtxcontextinfo.h"
 #include "cdb/cdbtm.h"
+#include "storage/lwlock.h"
 
 BufferUsage pgBufferUsage;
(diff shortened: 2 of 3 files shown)


src/include/executor/instrument.h
```diff
@@ -125,6 +125,8 @@ typedef struct InstrumentationHeader
 {
 	void	   *head;
 	int			free;
+
+	/* deprecated field, keep it for ABI, do not use it */
 	slock_t		lock;
```
adam8157: Suggest renaming the variable to `slock_t abi_reserved;`, or flag it with `GP_ABI_BUMP_FIXME` in the comment, for feature cleanup.

SmartKeyerror: Thanks for the advice, already added the `GP_ABI_BUMP_FIXME` flag.

SmartKeyerror:
CI cases have all passed, merge it: [Instrument Lock #150](https://github.com/EnterpriseDB/warehouse-pg-packaging/actions/runs/19456957737)

PR #69, opened 2025-11-04
6X: Fix assertion failures caused by incorrect use of fmtId()

adam8157:
This PR fixes two assertion failures introduced during the CVE-2025-1094 fix, both
due to subtle code differences between WarehousePG and upstream Postgres.

1, fmtId() was called before setFmtEncoding(), triggering an assertion failure on debug builds.
2, in reindexdb, fmtId() was used instead of fmtIdEnc(), causing the encoding information to be lost.

Both issues are now corrected to match the intended Postgres behavior.


adam8157:
The CI with assert passed the regression tests.

PR #70, opened 2025-11-10
change severity to `WARNING` in ftsprobe.c under certain conditions

SmartKeyerror:
In cases such as failures or reaching the maximum number of retries, the log level 
should be raised from `LOG` to `WARNING`. These log messages should not be 
controlled by the `gp_log_fts` GUC.


linwen (commented):
Use WARNING instead of LOG when FTS probe meets some abnormal, looks good to me.

in cdbfts.c, line:160, segment is down but still use LOG to record it, do we need to fix it? 
		if (FtsIsSegmentDown(segInfo))
		{
			ereport(LOG, (errmsg_internal("FTS: found fault with segment dbid %d. "
										  "Reconfiguration is in progress", (int) segInfo->config->dbid)));
			return true;
		}

PR #71, opened 2025-11-11
Fix NULL pointer in COPY (SELECT) TO option validation

adam8157:
COPY checks if the relation is an external table and skips option
validation since external tables may define their own options.

However, `COPY (SELECT) TO` has no relation (cstate->rel) at all, which
leads to a NULL pointer dereference. This fixes it.

### DIFF
--- a/src/backend/commands/copy.c
+++ b/src/backend/commands/copy.c
@@ -1642,7 +1642,7 @@ ProcessCopyOptions(ParseState *pstate,
 						 errmsg("conflicting or redundant options")));
 			cstate->skip_foreign_partitions = true;
 		}
-		else if (!rel_is_external_table(cstate->rel->rd_id))
+		else if (!cstate->rel || !rel_is_external_table(cstate->rel->rd_id))
 			ereport(ERROR,
 					(errcode(ERRCODE_SYNTAX_ERROR),
 					 errmsg("option \"%s\" not recognized",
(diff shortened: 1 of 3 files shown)


andreasscherbaum:
This PR, to a large extend, includes changes which are not related to this bugfix.
Both the whitespace/tab changes and the removal of the mailinglist entry should be a different PR.

.github/pull_request_template.md
```diff
@@ -1,6 +1,5 @@
 ## Here are some reminders before you submit the pull request
 - [ ] Add tests for the change
 - [ ] Document changes
-- [ ] Communicate in the mailing list if needed
```
andreasscherbaum: This change does not belong to this PR.

src/backend/commands/copy.c
```diff
@@ -7845,16 +7845,16 @@ static void
...
 	Oid		conversion_proc;
-	
+
 	/*
 	 * COPY FROM and RET: convert from file to server
 	 * COPY TO   and WET: convert from server to file
 	 */
 	if (iswritable)
 		conversion_proc = FindDefaultConversionProc(GetDatabaseEncoding(), encoding);
-	else
```
andreasscherbaum: This change does not belong into this PR.

.github/pull_request_template.md
adam8157: This one is in a separate commit :)

adam8157:
According to the review feedback, only the fix was kept, and the parts that tried to “sneak in” were removed.

linwen (approved):
looks good to me.

PR #77, opened 2025-11-21
[6X] Use LWLock instead of SpinLock at InstrumentationHeader

SmartKeyerror:
This is the backport of  #66

---

### Issue Summary

A crash was observed when acquiring the spinlock `InstrumentGlobal->lock`, 
which was caused by severe lock contention. Further investigation revealed 
deeper structural issues in the locking design.

### Root Cause

In `instrShmemRecycleCallback()`, a large loop performs multiple system calls 
(such as memory reset and free) while holding `InstrumentGlobal->lock`.
This code path can take a significant amount of time under heavy system load, 
preventing other backends from acquiring the same lock.

Moreover, spinlocks must never be held across kernel or blocking calls, as they 
are designed for very short, atomic critical sections. Holding a spinlock while 
performing system calls violates this principle and can easily lead to contention, 
starvation, or deadlocks.

### Resolution

The current use of a spinlock is inappropriate for this scenario.

The fix replaces the spinlock with a lightweight lock (LWLock), which is designed 
for longer critical sections and allows kernel calls within the protected region 
without risking system instability.


.abi-check/6.27.0/postgres.symbols.ignore
```diff
@@ -1 +1,4 @@
 ConfigureNamesInt_gp
+MainLWLockArray
```
adam8157: Where are these coming from?

SmartKeyerror: Do you mean that the existing ABI check doesn’t monitor changes to modified symbols?

PR #85, opened 2025-12-02
With GDD active, push the xid into waitGxids so QD can wait on it

SmartKeyerror:
### Core

Fix commit-state divergence between QD and QEs during two-phase commit.

In the existing 2PC implementation, QEs advance the distributed
transaction to “committed” immediately after completing CommitPrepare,
whereas the QD does not mark the transaction committed until it has
received successful CommitPrepare responses from all segments.  This
ordering gap introduces a window in which QEs consider the transaction
committed while the QD still regards it as in progress, leading to
inconsistent visibility and potential lock-wait anomalies.

This patch introduces a QD-side serialization mechanism to close that
window.  By enforcing a lock around global commit progression, concurrent
distributed transactions are prevented from observing intermediate
states, ensuring that QD and QEs remain consistent with respect to the
transaction’s commit status.

This ensures correct global lock behavior when GDD is enabled and
eliminates the transient divergence between coordinator and segments.

### Context

The basic pattern is as follows:

```
-- open gp_enable_global_deadlock_detector;
gpconfig -c gp_enable_global_deadlock_detector -v on
gpstop -rai

create table test (a int, b int);

begin;
update test set b = X where a = 1;
select * from test where a = 1;
commit;
```

When GDD is enabled, Greenplum shifts lock management from table-level 
locks to fine-grained tuple-level locks. Under this model, a transaction that 
performs an UPDATE must subsequently see its own changes when issuing 
a SELECT within the same transaction, preserving the expected MVCC visibility 
rules.

However, concurrent workload testing (e.g., pg_bench) revealed cases where 
this guarantee was not upheld. SELECT statements issued after an UPDATE within 
the same transaction could fail to observe the transaction’s own modifications, 
resulting in incorrect behavior under concurrent execution.

### DIFF
--- a/src/test/isolation2/expected/gdd/concurrent_update.out
+++ b/src/test/isolation2/expected/gdd/concurrent_update.out
@@ -48,7 +48,7 @@ SET
 3&: UPDATE t_concurrent_update SET b=b+10 WHERE a=1;  <waiting ...>
 
 -- transaction 2 suspend before commit, but it will wake up transaction 3 on segment
-2: select gp_inject_fault('before_xact_end_procarray', 'suspend', dbid) FROM gp_segment_configuration WHERE role='p' AND content=-1;
+2: select gp_inject_fault('before_xact_end_procarray', 'suspend', dbid, current_setting('gp_session_id')::int) FROM gp_segment_configuration WHERE role='p' AND content=-1;
  gp_inject_fault 
 -----------------
  Success:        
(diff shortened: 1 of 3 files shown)


adam8157 (approved):
```
Session 1 (My Transaction)    Session 2 (Other Txn)          QD State           QE State
────────────────────────────────────────────────────────────────────────────────────────
begin;
update test set b=10
  where a=1;                                                 T1: in-progress   T1: in-progress
                              begin;
                              update test set b=20
                                where a=1;                   T2: in-progress   T2: waiting for T1's lock
                              commit; (start 2PC)
                                                             T2: in-progress   ← KEY ISSUE!
                                ↓ QE finishes CommitPrepare                    T2: committed ✓
                                                             T2: in-progress   T2: committed
select * from test            ← Takes distributed snapshot   (Snapshot: T2 not  (Reality: T2 already
  where a=1;                    at this moment                yet committed)     committed!)
                                ↓ QD receives all responses
                              (commit completes)             T2: committed ✓
```

And if GDD is not enabled, we don't have concurrent UPDATEs on the same table, so we don't have the issue either. Am I understanding correctly?

adam8157 (approved):

PR #87, opened 2025-12-04
Fix pg_waldump flaky case

SmartKeyerror:
The test manually creates a WAL file with a corrupt header and a size of
32 KB. Its purpose is to verify that `pg_waldump --last-valid-walname`
correctly skips this invalid file.

Before processing the file name provided to `--last-valid-walname`,
pg_waldump first scans the target directory. During this directory
scan, `search_directory()` may pick any file whose name matches a valid
WAL file pattern. It then reads the file to determine `WalSegSz` and
validates that the size is acceptable. If the corrupt test file reports
an invalid segment size, pg_waldump errors out before
`LastValidWalSegment` is even reached.

This commit avoids that premature failure by writing
`DEFAULT_XLOG_SEG_SIZE` into the xlp_seg_size field of the corrupt WAL
file’s header, ensuring the initial directory scan does not abort
before the tool can evaluate the file properly with
LastValidWalSegment.


SmartKeyerror:
CI all pass: https://github.com/EnterpriseDB/warehouse-pg-packaging/actions/runs/19947338356

adam8157 (approved):

PR #114, opened 2026-01-28
Fix buffer over-read at TranslateDXLDatumGenericToScalar()

SmartKeyerror:
For scalar values, the length of the buffer returned by `GetByteArray()` 
is less than or equal to `sizeof(Datum)`, as the assertion said:

```
GPOS_ASSERT((ULONG) constant->constlen <= sizeof(Datum));
```

So we should use `constlen`(smaller value) to determine the buffer
length.

### DIFF
--- a/src/backend/gpopt/translate/CTranslatorUtils.cpp
+++ b/src/backend/gpopt/translate/CTranslatorUtils.cpp
@@ -2478,7 +2478,6 @@ CTranslatorUtils::CreateDatumFromCDXLDatumGeneric(BOOL passed_by_val, ULONG leng
 	Datum result = 0;
 	if (passed_by_val)
 	{
-		GPOS_ASSERT(length >= 0);
 		GPOS_ASSERT(length <= sizeof(Datum));
 		// make sure the length passed in equals to the result of datum_generic_dxl->Length()
 		GPOS_ASSERT(length == datum_generic_dxl->Length());
@@ -2487,9 +2486,9 @@ CTranslatorUtils::CreateDatumFromCDXLDatumGeneric(BOOL passed_by_val, ULONG leng
 	else
 	{
 		Datum val = gpdb::DatumFromPointer(datum_generic_dxl->GetByteArray());
-		length = (ULONG) gpdb::DatumSize(val, false, length);
-		BYTE *buffer = (BYTE *) gpdb::GPDBAlloc(length);
-		memcpy(buffer, datum_generic_dxl->GetByteArray(), length);
+		ULONG new_length = (ULONG) gpdb::DatumSize(val, false, length);
+		BYTE *buffer = (BYTE *) gpdb::GPDBAlloc(new_length);
+		memcpy(buffer, datum_generic_dxl->GetByteArray(), new_length);
 		result = gpdb::DatumFromPointer(buffer);
 	}
 	return result;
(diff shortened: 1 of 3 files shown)


src/backend/gpopt/translate/CTranslatorUtils.cpp
```diff
@@ -2470,4 +2470,28 @@ CTranslatorUtils::RelContainsForeignPartitions(const IMDRelation *rel,
...
 }
+
+Datum
+CTranslatorUtils::CreateDatumFromCDXLDatumGeneric(BOOL passed_by_val, ULONG length,
+												  const CDXLDatumGeneric *datum_generic_dxl)
+{
+	Datum result = 0;
+	if (passed_by_val)
+	{
+		GPOS_ASSERT(length >= 0);
```
adam8157: `length` is unsigned long, it's always >=0

src/backend/gpopt/translate/CTranslatorUtils.cpp
```diff
@@ -2470,4 +2470,28 @@ CTranslatorUtils::RelContainsForeignPartitions(const IMDRelation *rel,
...
+		GPOS_ASSERT(length <= sizeof(Datum));
+		// make sure the length passed in equals to the result of datum_generic_dxl->Length()
+		GPOS_ASSERT(length == datum_generic_dxl->Length());
+		memcpy(&result, datum_generic_dxl->GetByteArray(), length);
+	}
+	else
+	{
+		Datum val = gpdb::DatumFromPointer(datum_generic_dxl->GetByteArray());
+		length = (ULONG) gpdb::DatumSize(val, false, length);
+		BYTE *buffer = (BYTE *) gpdb::GPDBAlloc(length);
```
adam8157: The pass-by-reference branch used to allocate a `length + 1` size buffer and set the NULL terminator. I suggest keeping that behavior.

src/backend/gpopt/translate/CTranslatorUtils.cpp
```diff
@@ -2470,4 +2470,28 @@ CTranslatorUtils::RelContainsForeignPartitions(const IMDRelation *rel,
...
+		GPOS_ASSERT(length >= 0);
+		GPOS_ASSERT(length <= sizeof(Datum));
+		// make sure the length passed in equals to the result of datum_generic_dxl->Length()
+		GPOS_ASSERT(length == datum_generic_dxl->Length());
+		memcpy(&result, datum_generic_dxl->GetByteArray(), length);
+	}
+	else
+	{
+		Datum val = gpdb::DatumFromPointer(datum_generic_dxl->GetByteArray());
+		length = (ULONG) gpdb::DatumSize(val, false, length);
```
adam8157: Suggest using another local variable, instead of setting the function argument.

src/backend/gpopt/translate/CTranslatorUtils.cpp
SKeyerror: removed

src/backend/gpopt/translate/CTranslatorUtils.cpp
SKeyerror: changed to `new_length`

src/backend/gpopt/translate/CTranslatorUtils.cpp
SKeyerror: That's what we should fix

adam8157 (dismissed):

adam8157 (approved):

PR #143, opened 2026-04-15
Improve workflow run-names with prefixes and commit titles

adam8157:
- Add [CI]/[ABI] prefixes to run-names for easy identification
- Show commit title on push events instead of just branch name
- Add run-name to abi-tests.yml (previously had none)


24nishant (approved):
LGTM!

PR #144, opened 2026-04-16
Optimize AO/AOCS table sampling by using adaptive tuples-per-block

SKeyerror:
This PR optimizes TABLE SAMPLE performance for AO and AOCS tables by dynamically computing 
the number of tuples per logical block based on the 
table's average tuple width.

Previously, AO/AOCS table sampling used a fixed `AO_MAX_TUPLES_PER_HEAP_BLOCK` (32768) 
as the logical block size for all tables. This was suboptimal for tables with large tuples (e.g., vector 
types where each tuple is several KB), as it forced the sampling algorithm to scan through many 
more tuples than necessary to cover the logical address space.

This PR introduces an adaptive approach:  

  - Added a new function `ao_compute_sample_tuples_per_block()` that calculates the optimal 
tuples-per-block based on the relation's average tuple width from statistics                                                                      
  - For small tuples (e.g., integers): uses the maximum value (32768)
  - For large tuples (e.g., vectors): uses a proportionally smaller value based on 
`(BLCKSZ * 2) / avg_tuple_width`                                                                                                                          
  - The computed value is bounded between 32 and 32768 to ensure reasonable behavior

### DIFF
--- a/src/backend/access/aocs/aocsam_handler.c
+++ b/src/backend/access/aocs/aocsam_handler.c
@@ -2653,6 +2653,7 @@ aoco_scan_sample_next_block(TableScanDesc scan, SampleScanState *scanstate)
 	TsmRoutine 			*tsm = scanstate->tsmroutine;
 	AOCSScanDesc 		aoscan = (AOCSScanDesc) scan;
 	int64 				totalrows = AOCSScanDesc_TotalTupCount(aoscan);
+	int32				tuplesPerBlock = aoscan->sampleTuplesPerBlock;
 
 	/* return false immediately if relation is empty */
 	if (aoscan->targrow >= totalrows)
@@ -2660,7 +2661,7 @@ aoco_scan_sample_next_block(TableScanDesc scan, SampleScanState *scanstate)
 
 	if (tsm->NextSampleBlock)
 	{
-		int64 nblocks = (totalrows + (AO_MAX_TUPLES_PER_HEAP_BLOCK - 1)) / AO_MAX_TUPLES_PER_HEAP_BLOCK;
+		int64 nblocks = (totalrows + (tuplesPerBlock - 1)) / tuplesPerBlock;
 		int64 nextblk;
 
 		nextblk = tsm->NextSampleBlock(scanstate, nblocks);
@@ -2688,7 +2689,7 @@ aoco_scan_sample_next_block(TableScanDesc scan, SampleScanState *scanstate)
 			/* target the first row of the selected block */
 			Assert(aoscan->sampleTargetBlk < nblocks);
 
-			aoscan->targrow = aoscan->sampleTargetBlk * AO_MAX_TUPLES_PER_HEAP_BLOCK;
+			aoscan->targrow = aoscan->sampleTargetBlk * tuplesPerBlock;
 			return true;
 		}
 	}
@@ -2699,7 +2700,7 @@ aoco_scan_sample_next_block(TableScanDesc scan, SampleScanState *scanstate)
 
 		/* target the first row of the next block */
 		aoscan->sampleTargetBlk++;
-		aoscan->targrow = aoscan->sampleTargetBlk * AO_MAX_TUPLES_PER_HEAP_BLOCK;
+		aoscan->targrow = aoscan->sampleTargetBlk * tuplesPerBlock;
 
 		/* ran out of blocks, scan is done */
 		if (aoscan->targrow >= totalrows)
@@ -2715,7 +2716,8 @@ aoco_scan_sample_next_tuple(TableScanDesc scan, SampleScanState *scanstate,
 {
 	TsmRoutine 			*tsm = scanstate->tsmroutine;
 	AOCSScanDesc 		aoscan = (AOCSScanDesc) scan;
-	int64  				currblk = aoscan->targrow / AO_MAX_TUPLES_PER_HEAP_BLOCK;
+	int32				tuplesPerBlock = aoscan->sampleTuplesPerBlock;
+	int64  				currblk = aoscan->targrow / tuplesPerBlock;
 	int64 				totalrows = AOCSScanDesc_TotalTupCount(aoscan);
 
 	Assert(aoscan->sampleTargetBlk >= 0);
@@ -2732,12 +2734,12 @@ aoco_scan_sample_next_tuple(TableScanDesc scan, SampleScanState *scanstate,
 		 * Ask the tablesample method which rows to scan on this block. Refer
 		 * to AOCSScanDesc->sampleTargetBlk for our blocking scheme.
 		 *
-		 * Note: unlike heapam, we are guaranteed to have
-		 * AO_MAX_TUPLES_PER_HEAP_BLOCK tuples in this block (unless this is the
-		 * last such block in the relation)
+		 * Note: unlike heapam, we are guaranteed to have sampleTuplesPerBlock
+		 * tuples in this block (unless this is the last such block in the
+		 * relation)
 		 */
-		maxoffset = Min(AO_MAX_TUPLES_PER_HEAP_BLOCK,
-						totalrows - currblk * AO_MAX_TUPLES_PER_HEAP_BLOCK);
+		maxoffset = Min(tuplesPerBlock,
+						totalrows - currblk * tuplesPerBlock);
 		targetoffset = tsm->NextSampleTuple(scanstate,
 											currblk,
 											maxoffset);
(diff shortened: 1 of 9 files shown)


src/backend/access/appendonly/appendonlytid.c
```diff
@@ -34,3 +35,43 @@ AOTupleIdToString(AOTupleId *aoTupleId)
 
 	return AOTupleIdBuffer;
 }
+
+/*
+ * ao_compute_sample_tuples_per_block
```
adam8157: `appendonlytid.c` seems not the right file, is `appendonlyam_handler.c` better?

src/backend/access/appendonly/appendonlytid.c
```diff
@@ -14,6 +14,7 @@
 #include "postgres.h"
 
 #include "access/appendonlytid.h"
+#include "optimizer/plancat.h"
```
adam8157: A file at the access layer including the planner header doesn't look right.

src/include/access/appendonlytid.h
```diff
@@ -15,6 +15,8 @@
 #define APPENDONLYTID_H
 
 #include "c.h"
+#include "postgres.h"
```
adam8157: This file is also compiled as a client, please avoid including the root backend header, or put the changes at other places.

src/backend/access/appendonly/appendonlytid.c
```diff
@@ -34,3 +35,43 @@ AOTupleIdToString(AOTupleId *aoTupleId)
...
+
+	if (avg_tuple_width <= 0)
+	{
+		/* No statistics available, fall back to default */
+		return AO_MAX_TUPLES_PER_HEAP_BLOCK;
+	}
+
+	/*
+	 * Compute tuples per block based on BLCKSZ.
+	 * Use a multiplier of 2 to account for compression and provide margin.
```
adam8157: The `get_rel_data_width()` returns the not compressed width, so we should not *2?

src/backend/access/aocs/aocsam_handler.c
```diff
@@ -2715,7 +2716,8 @@ aoco_scan_sample_next_tuple(TableScanDesc scan, SampleScanState *scanstate,
 {
 	TsmRoutine 			*tsm = scanstate->tsmroutine;
 	AOCSScanDesc 		aoscan = (AOCSScanDesc) scan;
-	int64  				currblk = aoscan->targrow / AO_MAX_TUPLES_PER_HEAP_BLOCK;
+	int32				tuplesPerBlock = aoscan->sampleTuplesPerBlock;
+	int64  				currblk = aoscan->targrow / tuplesPerBlock;
```
adam8157, line 2720: Is there any case `sampleTuplesPerBlock` could be zero, and this will be divided by zero?

adam8157 (commented):
1. Can we have a big tuple test case?
2. How is the performance improved?
3. How does this work with ao_column(aocs)?

src/backend/access/appendonly/appendonlytid.c
SKeyerror: Reasonable, moved.

src/backend/access/appendonly/appendonlytid.c
SKeyerror: This is an average compression ratio, meaning the default compression rate is 50%. Therefore, the actual tuple count is estimated to be twice the compressed value.

src/include/access/appendonlytid.h
SKeyerror: Thanks for the reminding, have removed

src/backend/access/aocs/aocsam_handler.c
SKeyerror:
```cpp
	tuples_per_block = Max(tuples_per_block, 32);
	tuples_per_block = Min(tuples_per_block, AO_MAX_TUPLES_PER_HEAP_BLOCK);
```

This calculation promises the results will not be zero.

SmartKeyerror:
> Can we have a big tuple test case?

Yes, we can, in the local env it shows good performance, but I don't want to put it into the regression test, since it's not regression case, rather performance case.

---

> How is the performance improved?

Before this patch, the following case:

```sql
CREATE TABLE test_heap AS SELECT i, random()::text AS data FROM generate_series(1, 100000) i DISTRIBUTED BY (i);
CREATE TABLE test_ao_row WITH (appendonly=true) AS SELECT * FROM test_heap DISTRIBUTED BY (i);

SELECT count(*) FROM test_heap    TABLESAMPLE SYSTEM(100);  -- 100000  
SELECT count(*) FROM test_ao_row  TABLESAMPLE SYSTEM(100);  -- 100000 

SELECT count(*) FROM test_heap    TABLESAMPLE SYSTEM(10);      -- ~10000 (correct)  
SELECT count(*) FROM test_ao_row  TABLESAMPLE SYSTEM(10);    -- 0 or ~33333 (wrong!) 
```

The number of results returned for AO tables during standard sampling is severely distorted.

After the patch:

```sql
CREATE TABLE demo_heap(id int, embedding bytea);
CREATE TABLE demo_ao(id int, embedding bytea) USING ao_row;

INSERT INTO demo_heap SELECT i, repeat('x', 3072)::bytea FROM generate_series(1, 100000) i;
INSERT INTO demo_ao SELECT i, repeat('x', 3072)::bytea FROM generate_series(1, 100000) i;

ANALYZE demo_heap;
ANALYZE demo_ao;

SELECT 'heap' as table_type, count(*) as sampled_rows FROM demo_heap TABLESAMPLE SYSTEM(10);
 table_type | sampled_rows 
------------+--------------
 heap       |        10830
(1 row)

SELECT 'ao_row' as table_type, count(*) as sampled_rows FROM demo_ao TABLESAMPLE SYSTEM(10);
 table_type | sampled_rows 
------------+--------------
 ao_row     |         9920
(1 row)

SELECT 'ao_row' as table_type, count(*) as sampled_rows FROM demo_ao TABLESAMPLE SYSTEM(10);
 table_type | sampled_rows 
------------+--------------
 ao_row     |        10592
(1 row)
```

As we can see, the sampling results for the AO table are very close to those of the heap table, even in a wide table scenario.

At the same time, for narrow tables, such as those containing int type data, the patch still performs well.

```sql
CREATE TABLE test_heap AS SELECT i, i as data FROM generate_series(1,1000000) i DISTRIBUTED BY (i);
CREATE TABLE test_ao_row WITH (appendonly=true) AS SELECT * FROM test_heap DISTRIBUTED BY (i);

analyze test_heap;
analyze test_ao_row;

-- 10%
SELECT count(*) FROM test_heap TABLESAMPLE SYSTEM(10);
 count 
-------
 98172
(1 row)

-- AO returns almost 10%
SELECT count(*) FROM test_ao_row TABLESAMPLE SYSTEM(10);
 count 
-------
 90112
(1 row)

SELECT count(*) FROM test_ao_row TABLESAMPLE SYSTEM(10);
 count 
-------
 98304
(1 row)
```

----

> How does this work with ao_column(aocs)?

AOCS shares the same `ao_compute_sample_tuples_per_block()` function with AO row tables. The underlying logic remains largely unchanged; both use variable-length blocks, with the only difference being the shift from row-store to column-store.

src/backend/access/appendonly/appendonlytid.c
adam8157: Do we need to read pg_appendonly to check if the table is compressed? Cost too much?

SKeyerror: Make sense, added the `is_compressed` variable.

src/backend/access/appendonly/appendonlyam_handler.c
```diff
@@ -39,6 +40,7 @@
 #include "commands/progress.h"
 #include "commands/vacuum.h"
 #include "executor/executor.h"
+#include "optimizer/plancat.h"
```
adam8157, line 43: How about just declaring the function (I believe only get_rel_data_width()), instead of importing a planner header? The latter does not look right for a storage layer file.

adam8157 (approved):
For ao_colum, the sampling code doesn't map to anything physical, but the point is to get a reasonable nblocks. Together with the rest of the logic, the code does pull in a reasonable amount of data, so the improvement still holds.

The catch is that sampling on column store means more seeks — that's just how AOCS works. Long term, we could raise the lower bound of tuplesPerBlock for AOCS (maybe scale it with the number of columns) to use slightly bigger logical blocks and do fewer seeks — the trade-off is sampling bias from correlated rows in the same block.

src/backend/access/appendonly/appendonlyam_handler.c
SKeyerror: Thanks for the advice, I searched the function of `get_rel_data_width()` and found it's also used in `heapam_handler.c`, which is storage layer, and includes the all header of `optimizer/plancat.h`, so I think it's not a problem?

adam8157: In that case, it's fine with me hahaha

PR #146, opened 2026-04-23
fix incorrect DQA type classification under HAVING clause

SmartKeyerror:
### Context

When a HAVING clause is present with DQA (Distinct Qualified Aggregates), 
the planner inadvertently drops the HAVING clause, leading to incorrect query 
results. 

This issue manifests as inconsistent results across multiple executions of the 
same query. The root cause of this inconsistency is that the arrival order of the 
first tuple during network transmission is non-deterministic, which ultimately 
leads to unpredictable outputs.

```sql
CREATE TABLE sales (
  sale_date  date,  
  sale_key   varchar(50),
  cust_no    varchar(11),
  cust_type  varchar(1),
  order_no   varchar(20),          
  sale_amt   numeric(18,3)
) DISTRIBUTED BY (sale_key);

INSERT INTO sales (
    sale_date, 
    sale_key, 
    cust_no, 
    sale_amt, 
    order_no
)
SELECT
    generated_date, 
    'KEY-' || i::text,
    'C' || lpad((mod(i, 1000) + 1)::text, 8, '1'),
    (random() * 50000)::numeric(19,2),
    'ORD-' || i::text
FROM (
    SELECT 
        i, 
        '2026-04-01'::date + (random() * 6)::integer AS generated_date
    FROM generate_series(1, 100000) s(i)
) sub;

SELECT count(*) FROM (
  SELECT cust_no, sum(sale_amt), count(distinct sale_date) 
  FROM sales GROUP BY cust_no 
  HAVING sum(sale_amt) >= 100000 AND count(distinct sale_date) >= 5
)as sub;
 count 
-------
   879

SELECT count(*) FROM (
  SELECT cust_no, sum(sale_amt), count(distinct sale_date) 
  FROM sales GROUP BY cust_no 
  HAVING sum(sale_amt) >= 100000 AND count(distinct sale_date) >= 5
)as sub;
 count 
-------
   882
```

### RCA

The root cause lies in the absence of the aggregate columns, the plan difference makes the bug visible:                                                                                                                                                                                                
   
```sql
  -- Before fix (SINGLE_DQA: sum evaluated over deduplicated rows only)                                                                                                                                                                     
  HashAggregate  Group Key: cust_no                                                                                                                                                                                                         
    Filter: sum(sale_amt) >= 100000
    -> HashAggregate  Output: cust_no, sale_amt, sale_date   ← AGGSPLIT_SIMPLE, no partial agg                                                                                                                                              
                                                                                                                                                                                                                                            
  -- After fix (SINGLE_DQA_WITHAGG: partial sum accumulated during deduplication)                                                                                                                                                           
  Finalize HashAggregate  Group Key: cust_no                                                                                                                                                                                                
    Filter: sum(sale_amt) >= 100000                                                                                                                                                                                                         
    -> Partial HashAggregate  Output: cust_no, sale_amt, sale_date, PARTIAL sum(sale_amt)
```

### changes

  1. `recognize_dqa_type()`: After `scanning ctx->target->exprs`, also scan `ctx->havingQual`. 
 The HAVING qual retains the original aggregate expressions and correctly exposes plain aggregates, 
allowing the query to be promoted to `SINGLE_DQA_WITHAGG` or `MULTI_DQAS_WITHAGG`.
  3. `fetch_single_dqa_info()`: For `SINGLE_DQA_WITHAGG`, pull Vars referenced by plain aggregates 
 in `partial_grouping_target` and add any missing ones to `input_proj_target`, ensuring HAVING-required 
columns survive intermediate projection.  
  5. `fetch_multi_dqas_info()`: Apply the same Var back-fill for `MULTI_DQAS_WITHAGG`.

### DIFF
--- a/src/backend/cdb/cdbgroupingpaths.c
+++ b/src/backend/cdb/cdbgroupingpaths.c
@@ -2876,11 +2876,7 @@ fetch_multi_dqas_info(PlannerInfo *root,
 					}
 
 					if (!found)
-					{
 						add_column_to_pathtarget(proj_target, (Expr *) var, 0);
-						proj_target->sortgrouprefs = (Index *) repalloc(proj_target->sortgrouprefs,
-																	list_length(proj_target->exprs) * sizeof(Index));
-					}
 				}
 			}
 		}
--- a/src/backend/cdb/cdbgroupingpaths.c
+++ b/src/backend/cdb/cdbgroupingpaths.c
@@ -2878,11 +2878,8 @@ fetch_multi_dqas_info(PlannerInfo *root,
 					if (!found)
 					{
 						add_column_to_pathtarget(proj_target, (Expr *) var, 0);
-						if (proj_target->sortgrouprefs == NULL)
-							proj_target->sortgrouprefs = (Index *) palloc0(list_length(proj_target->exprs) * sizeof(Index));
-						else
-							proj_target->sortgrouprefs = (Index *) repalloc(proj_target->sortgrouprefs,
-																			list_length(proj_target->exprs) * sizeof(Index));
+						proj_target->sortgrouprefs = (Index *) repalloc(proj_target->sortgrouprefs,
+																	list_length(proj_target->exprs) * sizeof(Index));
 					}
 				}
 			}
--- a/src/backend/cdb/cdbgroupingpaths.c
+++ b/src/backend/cdb/cdbgroupingpaths.c
@@ -2381,6 +2381,18 @@ recognize_dqa_type(cdb_agg_planning_context *ctx)
 									   PVC_INCLUDE_AGGREGATES |
 									   PVC_INCLUDE_WINDOWFUNCS |
 									   PVC_INCLUDE_PLACEHOLDERS);
+		/*
+		 * Also check havingQual for non-DISTINCT aggregates.
+		 * Normal aggregates in HAVING clause also need special handling.
+		 */
+		if (ctx->havingQual)
+		{
+			List *having_aggs = pull_var_clause((Node *) ctx->havingQual,
+											   PVC_INCLUDE_AGGREGATES |
+											   PVC_INCLUDE_WINDOWFUNCS |
+											   PVC_INCLUDE_PLACEHOLDERS);
+			varnos = list_concat(varnos, having_aggs);
+		}
 		foreach (lc, varnos)
 		{
 			Node	   *node = lfirst(lc);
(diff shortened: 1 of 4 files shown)


src/backend/cdb/cdbgroupingpaths.c
```diff
@@ -2830,17 +2842,64 @@ fetch_multi_dqas_info(PlannerInfo *root,
...
+						{
+							found = true;
+							break;
+						}
+					}
+
+					if (!found)
+					{
+						add_column_to_pathtarget(proj_target, (Expr *) var, 0);
+						if (proj_target->sortgrouprefs == NULL)
```
adam8157: There is a same if/else block at the beginning of this function, here it's redundant.

src/test/regress/sql/gp_dqa.sql
```diff
@@ -685,3 +685,41 @@ DROP TABLE part_f;
...
+EXPLAIN (COSTS OFF, VERBOSE)
+SELECT count(*) FROM (
+  SELECT cust_no, sum(sale_amt), count(distinct sale_date) 
+  FROM sales GROUP BY cust_no 
+  HAVING sum(sale_amt) >= 100000 AND count(distinct sale_date) >= 5
+)as sub;
+
+-- Without this patch, this SQL will output 78, which is not correct
+SELECT count(*) FROM (
+  SELECT cust_no, sum(sale_amt), count(distinct sale_date)
```
adam8157, line 720: This is a single DQA with HAVING, better to have another multi-DQA with HAVING query too.

adam8157 (commented):
The root cause is not explained clearly in the message, I propose:
```
The plan type stayed at the DISTINCT-only variant instead of *_WITHAGG, that path's partial stage groups by GROUP BY + DISTINCT keys, collapsing each group to one tuple, and the non-DISTINCT agg's inputs were never projected for aggregation — only passed through from whichever tuple landed in the hash bucket first.

So the final agg ran over one representative per group, not all rows. The representative is whoever wins the hash-bucket race, which depends on arrival order from the upstream Redistribute Motion — non-deterministic across segments and runs.
```

adam8157:
In real-world cases, sale_amt is not identical within the same cust_no group, so the result is inconsistent—unlike your test case, which returns a consistently incorrect result.

Please confirm that this scenario is also fixed (I believe it is), or consider updating the test case.

src/backend/cdb/cdbgroupingpaths.c
SKeyerror: I didn't get it, could please point it out? Thanks a lot

SmartKeyerror:
> In real-world cases, sale_amt is not identical within the same cust_no group, so the result is inconsistent—unlike your test case, which returns a consistently incorrect result.
> 
> Please confirm that this scenario is also fixed (I believe it is), or consider updating the test case.

Yes, this scenario is indeed covered and fixed by this patch. However, it's quite tricky to construct a stable test case that reliably reproduces the inconsistent behavior caused by network tuple arrival order. The current test case focuses on verifying the planner's logic.

src/test/regress/sql/gp_dqa.sql
SKeyerror: Added.

src/backend/cdb/cdbgroupingpaths.c
adam8157: At the beginning of this `fetch_multi_dqas_info()` function
```
  proj_target = copy_pathtarget(path->pathtarget);
  if (proj_target->sortgrouprefs)
  {
      for (...) { }
  }
  else
      proj_target->sortgrouprefs = (Index *) palloc0(... * sizeof(Index));
```
After this, `proj_target->sortgrouprefs` must be NOT NULL, so the NULL branch is never touched.

And in add_column_to_pathtarget():
```
  target->exprs = lappend(target->exprs, expr);      //  N → N+1
  if (target->sortgrouprefs)
  {
      int nexprs = list_length(target->exprs);      // N+1
      target->sortgrouprefs = repalloc(target->sortgrouprefs,
                                       nexprs * sizeof(Index));
      target->sortgrouprefs[nexprs - 1] = sortgroupref;  // = 0
  }
```
It already did the exact same repalloc(), so the NOT NULL branch is a redundant action.


I bought this up because I noticed `fetch_single_dqa_info()` has no such code block after the `add_column_to_pathtarget()`.

SKeyerror: You're right, thanks for the pointing, I have removed the if branch.

adam8157 (approved):

PR #147, opened 2026-04-23
Add whpg6 workflow

pravesh-sharma:
This PR adds a CI workflow for whpg6.

`whpg-ci.yml:` Updated to support whpg6.

At line 43 on `whpg-ci.yml` not using the whpg version detected from `detect-config.bash` as we don't have docker image for whpg6 and whpg7 docker image can be used.


.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
+name: WarehousePG CI
+
+# =============================================================================
+# Configuration: EL versions for each WHPG major version
+# Change these values to update which EL versions are tested
+# =============================================================================
+env:
+  WHPG7_EL_VERSIONS: '["8","9"]'            # WHPG 7 supported EL versions
+  WHPG6_EL_VERSIONS: '["7", "8", "9"]'      # WHPG 6 supported EL versions
```
24nishant: we should remove the "7" for whpg6 as in the current worflow it is not supported?
What will happen is "7" is selected, will the workflow fail ?

.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
...
+          make clean -C src/pl/plpython
+          make -j$(nproc) install -C src/pl/plpython
+
+      - name: Setup gpadmin user
+        run: |
+          # Create symlinks for easier navigation
+          ln -s . gpdb_src                          # For concourse scripts compatibility
+          ln -s "${WHPG_SRC}" /workspace            # Easy to remember path
+          ln -s "${WHPG_SRC}" /home/gpadmin/gpdb_src  # Accessible from gpadmin home
+          # We delete the ggpadmin user for WHPG 6 compatibility as the setup script doesn't handle existing users.
```
24nishant: [nit] *gpadmin

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,77 @@
...
+        el_version: ${{ inputs.el_version == 'all' && fromJson('["8", "9"]') || fromJson(format('["{0}"]', inputs.el_version)) }}
+
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+
+    env:
+      IMAGE_NAME: ghcr.io/warehouse-pg/whpg7-rocky${{ matrix.el_version }}-build
```
24nishant: we should raise a PR to update the image names here, as we are using same image for whpg6 & 7. Also the `.github/workflows/build-docker-image.yml `should be updated for it.

.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
...
+    needs: detect-config
+    if: github.event_name != 'workflow_dispatch' || inputs.test_type == 'all' || inputs.test_type == 'installcheck'
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix:
+        el_version: ${{ fromJson(needs.detect-config.outputs.el_versions) }}
+    container:
+      image: ghcr.io/warehouse-pg/whpg7-${{ format('rocky{0}', matrix.el_version) }}-build
```
24nishant, line 95: same as above, name should be updated in a followup PR

.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
...
+    needs: detect-config
+    if: github.event_name != 'workflow_dispatch' || inputs.test_type == 'all' || inputs.test_type == 'orca-unit-tests'
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        el_version: ${{ fromJson(needs.detect-config.outputs.el_versions) }}
+    container:
+      image: ghcr.io/warehouse-pg/whpg7-${{ format('rocky{0}', matrix.el_version) }}-build
```
24nishant, line 322: same as above, name should be updated in a followup PR.

.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
+name: WarehousePG CI
+
+# =============================================================================
+# Configuration: EL versions for each WHPG major version
+# Change these values to update which EL versions are tested
+# =============================================================================
+env:
+  WHPG7_EL_VERSIONS: '["8","9"]'            # WHPG 7 supported EL versions
```
24nishant: [nit] spacing between `["8","9"]`

24nishant:
Can you please check the ABI tests are failing in the build step?

.github/workflows/whpg-ci.yml
pravesh-sharma: I have not checked if it fails for el 7 or not. However, I will remove el 7.

.github/workflows/whpg-ci.yml
pravesh-sharma: Fixed

.github/workflows/whpg-ci.yml
pravesh-sharma: Once changes on `.github/workflows/build-docker-image.yml` is done I will update PR

pravesh-sharma:
> Can you please check the ABI tests are failing in the build step?

@24nishant  Fixed

.github/workflows/README.md
```diff
@@ -0,0 +1,197 @@
...
+This dual-build approach ensures the optimizer works correctly in both production-like and debug environments. The underlying script (`concourse/scripts/unit_tests_gporca.bash`) handles both builds automatically.
+
+#### Configuration
+
+Configuration is centralized at the top of the workflow file (single source of truth). Scripts require these values from the workflow environment and will fail if not provided.
+
+```yaml
+env:
+  WHPG7_EL_VERSIONS: '["8","9"]'            # WHPG 7 supported EL versions
+  WHPG6_EL_VERSIONS: '["7", "8", "9"]'  # WHPG 6 supported EL versions
```
adam8157: Does CI support EL7? If not, don't mention.

Or, are we going to support EL7?

.github/workflows/build-docker-image.yml
adam8157: Or just build images for WHPG 6, there should be some differences such as Python packages.

.github/scripts/run-installcheck.bash
```diff
@@ -0,0 +1,135 @@
...
+    if [[ "${test_target}" == "installcheck-world" ]]; then
+        cd "${WHPG_SRC}"
+    else
+        cd "${WHPG_SRC}/src/test/regress"
+    fi
+
+    # Run tests based on version
+    if [[ "${WHPG_MAJORVERSION}" == 6* ]]; then
+        # WHPG 6: Test PL/Python3 first
+        make installcheck -C "${WHPG_SRC}/src/pl/plpython" python_majorversion=3 || true
```
adam8157: Why `|| true`?

.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
...
+      - name: Install
+        id: install
+        run: make install -j$(nproc)
+
+      - name: Rebuild PL/Python with Python 3 (WHPG 6 only)
+        if: ${{ env.WHPG_MAJORVERSION == '6' }}
+        run: |
+          PYTHON=/usr/bin/python3 \
+          CC='ccache gcc -m64' \
+          CFLAGS='-O2 -g3' LDFLAGS='-Wl,--enable-new-dtags -Wl,--export-dynamic ' \
```
adam8157: Trailing space in LDFLAGS, and missed the rpath settings.

.github/workflows/whpg-ci.yml
```diff
@@ -0,0 +1,442 @@
...
+            ldconfig
+
+      - name: Set up Python 2 for initial build (WHPG 6 only)
+        if: ${{ env.WHPG_MAJORVERSION == '6' }}
+        run: |
+          alternatives --set python /usr/bin/python2
+          python --version
+
+      # Note: --enable-mapreduce and --with-pythonsrc-ext are recognized
+      # only in whpg6. They produce "unrecognized options" warnings in whpg7 but don't
```
adam8157: Comments not for WHPG 6

.github/workflows/build-docker-image.yml
adam8157: If we want to use the same image, just don't add the dockerfiles and build-docker-image.yaml

adam8157 (commented):
I think detect_whpg_version might be unnecessary here — could be wrong though.

My understanding is that each branch carries its own copy of the workflow files, and GitHub Actions always reads them from the branch where the event happens (push uses the pushed commit's tree; pull_request uses the base branch's HEAD). So a single workflow file can never be "shared" across branches at runtime — main's whpg-ci.yml only ever runs on main events, and the 6X copy only ever runs on 6X events.

If that's right, then on main the WHPG major version is effectively always 7, and on 6X it's always 6. The git describe + parse + validate dance just produces a constant per branch, which we could hardcode in the workflow env (e.g. WHPG_MAJOR: 7 on main, WHPG_MAJOR: 6 on 6X) and skip the detection step + the WHPG6_EL_VERSIONS / if whpg_major == "7" branches entirely.

@24nishant, am I missing a case where the detection actually buys us something?

.github/scripts/run-installcheck.bash
pravesh-sharma: I have taken this file from main branch. @24nishant can you please help with this?

.github/workflows/whpg-ci.yml
pravesh-sharma: we might not need rpath as I have linked the terces path already in step `Build Xerces with Python 2 (WHPG 6 only)`

.github/workflows/build-docker-image.yml
pravesh-sharma: I am removing `build-docker-image.yaml`. We can use the same image since Python 2 is just an extra package and we need Python 3 as well.

.github/scripts/run-installcheck.bash
24nishant: I think it is a left over from the testing, should be removed.  Thanks for pointing it out.

24nishant:
> on actually

yes, agree. We can fix the value, with the current use case it is not required.

adam8157 (approved):

adam8157:
I went ahead and pushed two more small fixes directly so we don't drag this out further. Apologies for being nitpicky on this one!

pravesh-sharma:
> I went ahead and pushed two more small fixes directly so we don't drag this out further. Apologies for being nitpicky on this one!

@adam8157 Thanks! For updating the docker image names I will raise a separate PR as it is required for main branch also.

PR #149, opened 2026-04-29
ORCA(fix): Narrow CTE Producer columns in MDQA to Join xform

24nishant:
### Summary
Fixes a SIGSEGV encountered (debug-build assertion at `CTranslatorExprToDXL.cpp`) when the optimizer rewrites a query
with multiple distinct-qualified aggregates (MDQAs) over a **partitioned table**. 

```
CREATE TABLE mdqa_part(
  a int,    -- group-by key + distribution key
  b int,    -- DISTINCT arg
  c int,    -- DISTINCT arg
  d int     -- unused unless explicitly referenced
)
DISTRIBUTED BY (a)
PARTITION BY RANGE(a) (
  START (0) END (100) EVERY (50),
  DEFAULT PARTITION other
);
INSERT INTO mdqa_part SELECT i, i%5, i%3, i%7 FROM generate_series(1, 100) i;
CREATE INDEX mdqa_part_b_idx ON mdqa_part(b);
ANALYZE mdqa_part;

EXPLAIN SELECT a, count(DISTINCT b) AS cnt_b, count(DISTINCT c) AS cnt_c FROM mdqa_part WHERE b BETWEEN 0 AND 2 GROUP BY a;
```

image

`CXformGbAggWithMDQA2Join` was declaring every column, the child could produce on the CTE Producer, rather than just the columns the GbAgg actually needs. The CTE producer only needs *(group keys ∪ aggregate-argument refs)* clipped to what the child can produce. 

On the partitioned-table path the over-declared columns include unreferenced and system columns whose `EUsage` state is `EUnknown`/`EUnused`, violating an invariant in `MakeDXLTableDescr` that requires required columns to be `EUsed`.

The narrowing preserves correctness on every shape (including system columns when referenced, group-only keys, no-GROUP-BY queries, and inline views) because DeriveUsedColumns already walks the agg list correctly and Include(Pdrgpcr()) always picks up the group keys regardless of whether they appear in the projection.

After the fix plan - 

                                                       QUERY PLAN
```
        ----------------------------------------------------------------------------------------------------------
         Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..868.00 rows=1 width=20)
           ->  Sequence  (cost=0.00..868.00 rows=1 width=20)
                 ->  Shared Scan (share slice:id 1:0)  (cost=0.00..6.00 rows=1 width=1)
                       ->  Dynamic Index Scan on mdqa_part_b_idx on mdqa_part  (cost=0.00..6.00 rows=1 width=12)
                             Index Cond: ((b >= 0) AND (b <= 2))
                             Number of partitions to scan: 3 (out of 3)
                 ->  Hash Join  (cost=0.00..862.00 rows=1 width=20)
                       Hash Cond: (NOT (share0_ref3.a IS DISTINCT FROM share0_ref2.a))
                       ->  GroupAggregate  (cost=0.00..431.00 rows=1 width=12)
                             Group Key: share0_ref3.a
                             ->  Sort  (cost=0.00..431.00 rows=1 width=8)
                                   Sort Key: share0_ref3.a
                                   ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=1 width=8)
                       ->  Hash  (cost=431.00..431.00 rows=1 width=12)
                             ->  GroupAggregate  (cost=0.00..431.00 rows=1 width=12)
                                   Group Key: share0_ref2.a
                                   ->  Sort  (cost=0.00..431.00 rows=1 width=8)
                                         Sort Key: share0_ref2.a
                                         ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=1 width=8)
         Optimizer: GPORCA
        (20 rows)
```

### DIFF
--- a/.github/workflows/whpg-ci.yml
+++ b/.github/workflows/whpg-ci.yml
@@ -167,7 +167,7 @@ jobs:
           CC='ccache gcc -m64' \
           CFLAGS='-O2 -g3' LDFLAGS='-Wl,--enable-new-dtags -Wl,--export-dynamic' \
           ./configure --with-quicklz --disable-gpperfmon --with-gssapi --enable-mapreduce --enable-orafce --enable-ic-proxy \
-                      --enable-orca --with-libxml --with-pythonsrc-ext --with-uuid=e2fs --with-pgport=5432 --enable-tap-tests \
+                      --enable-orca --with-libxml --with-pythonsrc-ext --with-uuid=e2fs --with-pgport=5432 --enable-tap-tests --with-llvm \
                       --enable-debug-extensions --with-perl --with-python --with-openssl --with-pam --with-ldap --with-includes="" \
                       --with-libraries="" --disable-rpath \
                       --prefix=/usr/local/greenplum-db-devel \
@@ -205,7 +205,7 @@ jobs:
         id: cluster
         run: |
           source concourse/scripts/common.bash
-          export WITH_MIRRORS=false
+          export WITH_MIRRORS=true
           make_cluster
 
       - name: Run installcheck tests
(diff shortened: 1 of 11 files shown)


SKeyerror (approved):
LGTM.

PR #150, opened 2026-05-01
Enable mirrors and JIT

24nishant:
The installcheck suite was running against a mirrorless cluster (WITH_MIRRORS=false) and without JIT (--with-llvm missing from configure). This caused 11 test failures on every run — not flakes. 

Also fixed - `fts_segment_reset isolation2 test` :
Added pg_sleep(2) between the PANIC-triggering session and the session that expects to block on the resetting segment. Without this, on fast runners session 2 can dispatch before the PANIC takes effect, causing
the CREATE TABLE to succeed instead of blocking.

image


pravesh-sharma:
@24nishant Can we backport this to WHPG 6?

adam8157 (approved):

24nishant:
> @24nishant Can we backport this to WHPG 6?

yes, please add it. You can add in the current [PR](https://github.com/warehouse-pg/warehouse-pg/pull/147) also.

PR #151, opened 2026-05-07
strip RelabelType when pulling up distribution key through binary-compatible cast in partitioned table scans

SKeyerror:
## Problem                                                                                                                                                                                                                                
                                                                                                                                                                                                                                            
When a query casts a partitioned table's distribution key to a                                                                                                                                                                            
binary-compatible type (e.g. `varchar::text`), the planner incorrectly                                                                                                                                                                    
process the re-label Var in `pullUpExpr_mutator()`.

## Reproduction

```sql
CREATE TABLE test_pullup_ec (id varchar(10), val int)
DISTRIBUTED BY (id)
PARTITION BY RANGE (val) (START (1) END (11) EVERY (5));

EXPLAIN (COSTS OFF) SELECT id::text, count(*) FROM test_pullup_ec GROUP BY id::text;
ERROR:  could not pull up equivalence class using projected target list (pathkeys.c:1539)
```

## RCA

`apply_scanjoin_target_to_paths()` pushes the outer query's projection                                                                                                                                                                    
down to partition child rels. For a binary-compatible cast like `id::text`,
this produces a `RelabelType(Var(id, varchar))` node in each partition                                                                                                                                                                    
child's `reltarget->exprs`.                                                                                                                                                                                                               
                                                                                                                                                                                                                                            
When `create_append_path()` later calls `cdbpathlocus_pull_above_projection()`                                                                                                                                                            
to derive the `AppendPath`'s locus, it invokes `cdbpullup_expr()` to rewrite                                                                                                                                                              
the distribution key expression in terms of the partition child's target. 
Inside `pullUpExpr_mutator()`, the code that handles `List of Expr`                                                                                                                                                                       
targetlists (i.e., planner-side `reltarget->exprs`, not plan-node                                                                                                                                                                         
`TargetEntry` lists) compared the distribution key Var directly against                                                                                                                                                                   
each targetlist entry:                                                                                                                                                                                                                  
                                                                                                                                                                                                                                            
```cpp                                                                                                                                                                                                                            
if (IsA(tlistexpr, Var))  /* fails when tlistexpr is RelabelType(Var) */
```                                                                                                                                                              
                                                                                                                                                                                                                                            
Because tlistexpr is a RelabelType node rather than a plain Var,                                                                                                                                                                          
the check fails, the Var is not found, and Pull-Up returns NULL.                                                                                                                                                                          
`cdbpathlocus_pull_above_projection()` then falls back to a Strewn locus,                                                                                                                                                                   
losing the Hashed(id) distribution information for the `AppendPath`.  

## Fix

Strip RelabelType nodes before matching, mirroring the approach already
used in `cdbpullup_findEclassInTargetList()`:

```cpp
naked_tlistexpr = tlistexpr;                                                                                                                                                                                                            
while (IsA(naked_tlistexpr, RelabelType))
    naked_tlistexpr = (Expr *) ((RelabelType *) naked_tlistexpr)->arg;                                                                                                                                                                    
   
if (IsA(naked_tlistexpr, Var))  /* now correctly matches */ 
```

When substituting from newvarlist, also strip `RelabelType` from the                                                                                                                                                                        
substituted entry before returning it. This is required because                                                                                                                                                                         
`expression_tree_mutator` re-wraps the substituted inner `Var` in the outer                                                                                                                                                                   
`RelabelType` at the call site; without stripping, the `newvarlist` entry                                                                                                                                                                     
would be returned in full and then re-wrapped.

### DIFF
--- a/src/backend/cdb/cdbpullup.c
+++ b/src/backend/cdb/cdbpullup.c
@@ -119,6 +119,7 @@ pullUpExpr_mutator(Node *node, void *context)
 		{
 			ListCell   *cell;
 			Expr	   *tlistexpr = NULL;
+			Expr	   *naked_tlistexpr = NULL;
 			AttrNumber	targetattno = 1;
 
 			foreach(cell, ctx->targetlist)
@@ -128,10 +129,20 @@ pullUpExpr_mutator(Node *node, void *context)
 				/*
 				 * We don't use equal(), because we want to ignore typmod.
 				 * A projection sometimes loses typmod, and that's OK.
+				 *
+				 * Also strip RelabelType nodes, consistent with
+				 * cdbpullup_findEclassInTargetList(), to handle
+				 * binary-compatible casts (e.g. varchar->text) that appear
+				 * in reltarget->exprs after apply_scanjoin_target_to_paths()
+				 * pushes projections down to partition child rels.
 				 */
-				if (IsA(tlistexpr, Var))
+				naked_tlistexpr = tlistexpr;
+				while (IsA(naked_tlistexpr, RelabelType))
+					naked_tlistexpr = (Expr *) ((RelabelType *) naked_tlistexpr)->arg;
+
+				if (IsA(naked_tlistexpr, Var))
 				{
-					Var		   *tlistvar = (Var *) tlistexpr;
+					Var		   *tlistvar = (Var *) naked_tlistexpr;
 
 					if (var->varno == tlistvar->varno &&
 						var->varattno == tlistvar->varattno &&
@@ -143,16 +154,31 @@ pullUpExpr_mutator(Node *node, void *context)
 			if (!cell)
 				goto fail;
 
-			/* Substitute the corresponding entry from newvarlist, if given. */
+			/*
+			 * Substitute the corresponding entry from newvarlist, if given.
+			 *
+			 * Strip RelabelType from the newvarlist entry for the same
+			 * reason as above: we are substituting the inner Var node only.
+			 * expression_tree_mutator re-wraps the result in RelabelType at
+			 * the outer level, so returning the full cast expression here
+			 * would produce a double-cast and fail the type Assert below.
+			 */
 			if (ctx->newvarlist)
-				newnode = (Node *) copyObject(list_nth(ctx->newvarlist,
-													   targetattno - 1));
+			{
+				Expr	   *newvar_entry = (Expr *) list_nth(ctx->newvarlist,
+															  targetattno - 1);
+
+				while (IsA(newvar_entry, RelabelType))
+					newvar_entry = (Expr *) ((RelabelType *) newvar_entry)->arg;
+				newnode = (Node *) copyObject(newvar_entry);
+			}
 
 			/* Substitute a Var node referencing the targetlist entry. */
 			else
 				newnode = (Node *) cdbpullup_make_expr(ctx->newvarno,
 													   targetattno,
-													   tlistexpr, true);
+													   naked_tlistexpr,
+													   true);
 		}
 
 		/* Make sure we haven't inadvertently changed the data type. */
--- a/src/backend/cdb/cdbpullup.c
+++ b/src/backend/cdb/cdbpullup.c
@@ -101,10 +101,22 @@ pullUpExpr_mutator(Node *node, void *context)
 			if (!tle)
 				goto fail;
 
-			/* Substitute the corresponding entry from newvarlist, if given. */
+			/*
+			 * Substitute the corresponding entry from newvarlist, if given.
+			 *
+			 * Strip RelabelType for the same reason as the RelOptInfo branch
+			 * below: we substitute the inner Var only and rely on
+			 * expression_tree_mutator to re-wrap RelabelType at the outer
+			 * level, avoiding a double-cast that would fail the type Assert.
+			 */
 			if (ctx->newvarlist)
-				newnode = (Node *) copyObject(list_nth(ctx->newvarlist,
-													   tle->resno - 1));
+			{
+				Expr	*newvar_entry = (Expr *) list_nth(ctx->newvarlist,
+																tle->resno - 1);
+				while (IsA(newvar_entry, RelabelType))
+					newvar_entry = (Expr *) ((RelabelType *) newvar_entry)->arg;
+				newnode = (Node *) copyObject(newvar_entry);
+			}
 
 			/* Substitute a Var node referencing the targetlist entry. */
 			else
@@ -142,7 +154,7 @@ pullUpExpr_mutator(Node *node, void *context)
 
 				if (IsA(naked_tlistexpr, Var))
 				{
-					Var		   *tlistvar = (Var *) naked_tlistexpr;
+					Var	*tlistvar = (Var *) naked_tlistexpr;
 
 					if (var->varno == tlistvar->varno &&
 						var->varattno == tlistvar->varattno &&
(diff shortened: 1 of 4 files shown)


adam8157 (dismissed):
LGTM. Nice fix for the EC pullup degenerating to Strewn when apply_scanjoin_target_to_paths() pushes a binary-compatible cast (e.g. varchar -> text) down into a partition child's reltarget->exprs, turning bare Var into
RelabelType(Var). Stripping RelabelType for both the match and the newvarlist substitution (to avoid double-cast after expression_tree_mutator re-wraps) is right, and consistent with the earlier 0babc980b1f.

One follow-up worth considering: the TargetEntry branch has the same shape — matching is RelabelType-tolerant via tlist_member_ignore_relabel(), but the newvarlist substitution doesn't strip. Today's planner can't reach it (cdbsubselect.c passes newvarlist = NULL; cdbpathlocus_make_for_subquery() filters new_vars to bare Vars), so it's rather than a real bug. But the two branches should share the same processing.

Suggest a small defensive so a future caller can't silently regress it:
```
-                     /* Substitute the corresponding entry from newvarlist, if given. */
+                     /*
+                      * Substitute the corresponding entry from newvarlist, if given.
+                      * Strip RelabelType for the same reason as the RelOptInfo branch
+                      * below: we substitute the inner Var only and rely on
+                      * expression_tree_mutator to re-wrap RelabelType at the outer
+                      * level, avoiding a double-cast that would fail the type Assert.
+                      */
                      if (ctx->newvarlist)
-                             newnode = (Node *) copyObject(list_nth(ctx->newvarlist,
-                                                                                                        tle->resno - 1));
+                     {
+                             Expr       *newvar_entry = (Expr *) list_nth(ctx->newvarlist,
+                                                                                                                       tle->resno - 1);
+
+                             while (IsA(newvar_entry, RelabelType))
+                                     newvar_entry = (Expr *) ((RelabelType *) newvar_entry)->arg;
+                             newnode = (Node *) copyObject(newvar_entry);
+                     }
```
Not blocking.

src/test/regress/sql/partition.sql
```diff
@@ -4252,3 +4252,21 @@ PARTITION BY RANGE (year)
...
+PARTITION BY RANGE (val) (START (1) END (11) EVERY (5));
+
+INSERT INTO test_pullup_ec VALUES ('a', 1), ('a', 6), ('b', 2), ('b', 7), ('c', 3);
+
+-- Trigger: Casting the distribution key to text in the SELECT clause (a binary-compatible cast).
+-- After `apply_scanjoin_target_to_paths` pushes down `id::text`,
+-- the child partition's `reltarget` becomes `RelabelType(Var(id, varchar))`.
+-- Old logic: `IsA(tlistexpr, Var)` checks the `RelabelType` -> fails -> marks as Strewn.
+-- New logic: Strips the `RelabelType` -> finds the underlying `Var` -> Pull-Up succeeds -> retains Hashed(id).
+EXPLAIN (COSTS OFF) SELECT id::text, count(*) FROM test_pullup_ec GROUP BY id::text;
```
24nishant, line 4268: Should we consider adding other test cases like - 

1. chain of binary-compatible casts

`SELECT (id::text)::varchar(20), count(*) FROM t1 GROUP BY (id::text)::varchar(20);`

2. Multi-column distribution key, only one column cast.
3. Different binary-compatible cast pair (varchar )
4. A join query on partitioned table with cast on the distribution key

src/backend/cdb/cdbpullup.c
```diff
@@ -128,10 +129,20 @@ pullUpExpr_mutator(Node *node, void *context)
...
+				 *
+				 * Also strip RelabelType nodes, consistent with
+				 * cdbpullup_findEclassInTargetList(), to handle
+				 * binary-compatible casts (e.g. varchar->text) that appear
+				 * in reltarget->exprs after apply_scanjoin_target_to_paths()
+				 * pushes projections down to partition child rels.
 				 */
-				if (IsA(tlistexpr, Var))
+				naked_tlistexpr = tlistexpr;
+				while (IsA(naked_tlistexpr, RelabelType))
```
24nishant, line 152: Should we consider moving this check & convert into a function , as this is used multiple time in the code? may be in another PR.

src/test/regress/sql/partition.sql
SKeyerror: Thanks for the suggestion!

`(id::text)::varchar(20)` looks like it's meant to test the while loop in the fix (handling nested RelabelTypes), but actually, this SQL doesn't generate nested RelabelType nodes.

---

```
DISTRIBUTED BY (id, region) 
SELECT id::text, region, count(*) FROM t2 GROUP BY id::text, region; 
```
For the query above, the code path is `cdbpathlocus_pull_above_projection` calling `cdb_pull_up_eclass` for each distribution key column separately. This just executes the exact same fix multiple times without hitting any new code branches.

---

The fix itself simply does `while (IsA(naked, RelabelType))` — it only checks the node type and is completely agnostic to the specific cast OID. Any binary-compatible cast will go through the exact same code path. Therefore, testing `bpchar::text` or `name::text` exercises the exact same logic as `varchar::text`

---

I tried 

```
CREATE TABLE dim (id varchar(10), name text) DISTRIBUTED BY (id);
EXPLAIN (COSTS OFF) SELECT f.id::text, d.name FROM t1 f JOIN dim d ON f.id::text = d.id;
```

But it not triggers the wrong path, so it uses another code path to process.

SKeyerror:
> LGTM. Nice fix for the EC pullup degenerating to Strewn when apply_scanjoin_target_to_paths() pushes a binary-compatible cast (e.g. varchar -> text) down into a partition child's reltarget->exprs, turning bare Var into RelabelType(Var). Stripping RelabelType for both the match and the newvarlist substitution (to avoid double-cast after expression_tree_mutator re-wraps) is right, and consistent with the earlier [0babc98](https://github.com/warehouse-pg/warehouse-pg/commit/0babc980b1fe0f363f406b6a66da401283e45496).
> 
> One follow-up worth considering: the TargetEntry branch has the same shape — matching is RelabelType-tolerant via tlist_member_ignore_relabel(), but the newvarlist substitution doesn't strip. Today's planner can't reach it (cdbsubselect.c passes newvarlist = NULL; cdbpathlocus_make_for_subquery() filters new_vars to bare Vars), so it's rather than a real bug. But the two branches should share the same processing.
> 
> Suggest a small defensive so a future caller can't silently regress it:
> 
> ```
> -                     /* Substitute the corresponding entry from newvarlist, if given. */
> +                     /*
> +                      * Substitute the corresponding entry from newvarlist, if given.
> +                      * Strip RelabelType for the same reason as the RelOptInfo branch
> +                      * below: we substitute the inner Var only and rely on
> +                      * expression_tree_mutator to re-wrap RelabelType at the outer
> +                      * level, avoiding a double-cast that would fail the type Assert.
> +                      */
>                       if (ctx->newvarlist)
> -                             newnode = (Node *) copyObject(list_nth(ctx->newvarlist,
> -                                                                                                        tle->resno - 1));
> +                     {
> +                             Expr       *newvar_entry = (Expr *) list_nth(ctx->newvarlist,
> +                                                                                                                       tle->resno - 1);
> +
> +                             while (IsA(newvar_entry, RelabelType))
> +                                     newvar_entry = (Expr *) ((RelabelType *) newvar_entry)->arg;
> +                             newnode = (Node *) copyObject(newvar_entry);
> +                     }
> ```
> 
> Not blocking.

That makes sense, added. Please check again, thanks.

src/backend/cdb/cdbpullup.c
SKeyerror: I agree that this logic is used in quite a few places, and extracting it into a separate function makes perfect sense. 

And yes, I'd prefer to handle this in a dedicated refactoring PR too.

adam8157 (approved):

24nishant (approved):
LGTM!

PR #153, opened 2026-05-18
ORCA (enhancement): cache array interval constraints to speed up IN-clause planning

24nishant:
## Summary

ORCA's planning time on partitioned tables with medium-to-large `IN` / `NOT IN` clauses is dominated by repeated derivation of `CConstraintInterval` objects from the same `CScalarArrayCmp` operator during a single optimisation pass. This PR adds a per-query cache for those intervals and a tuning threshold that caps the sort+dedup cost on very large IN lists. 

### Two new GUCs

| GUC | Default | Purpose |
|---|---|---|
| `optimizer_array_constraint_cache` | `on` | Enable the per-query interval cache. |
| `optimizer_array_interval_threshold` | `1000` | Above this IN-list size, skip the interval path and let the cheaper disjunction path (`optimizer_array_expansion_threshold`) take over. |

### Mechanism

- **Per-query cache** lives on `COptCtxt`, keyed on the  `CScalarArrayCmp` operator pointer plus the infer-nulls flag.
- **AddRef on the cache key** pins the operator's address so the allocator cannot recycle a freed `CScalarArrayCmp` while the cache still references it (fixes a class of dangling-pointer hazard surfaced during dev on the partition-pruning path).
- **Destructor reorder on `COptCtxt`** tears the cache down before the refcount pool whose entries it pins, so shutdown is well-ordered.
- **Threshold gate** in `PcnstrIntervalFromScalarArrayCmp` short-circuits the interval path above the new threshold; the
  existing disjunction path handles those cases.
- **DXL plumbing** carries the threshold through  `CHint::m_ulArrayIntervalThreshold` and the new
  `ArrayIntervalThreshold` attribute in `OptimizerConfig.xml`, so minidumps round-trip correctly.
- **Trace-flag mapping**: `optimizer_array_constraint_cache` ↔`EopttraceArrayConstraintCache = 103048`.


src/backend/gporca/libgpopt/src/base/CConstraintInterval.cpp
```diff
@@ -293,6 +294,29 @@ CConstraintInterval::PcnstrIntervalFromScalarArrayCmp(CMemoryPool *mp,
...
+
+	// Skip interval constraint creation for arrays exceeding the interval
+	// threshold.
+	COptimizerConfig *optimizer_config =
+		COptCtxt::PoctxtFromTLS()->GetOptimizerConfig();
+	ULONG array_interval_threshold =
+		optimizer_config->GetHint()->UlArrayIntervalThreshold();
+	if (ulArrayExprArity > array_interval_threshold)
+	{
+		return nullptr;
```
adam8157, line 317: With the default GUCs this disables constraint derivation entirely for large IN lists. When the array has more than `optimizer_array_interval_threshold` (default 1000) elements, this path returns nullptr, but the disjunction fallback in `CConstraint::PcnstrFromScalarArrayCmp` is gated by `optimizer_array_expansion_threshold`, whose default is only 20. So for `col IN (<1001+ constants>)` we derive no constraint at all — losing partition pruning and cardinality bounds that we had before this PR. The GUC description ("the cheaper disjunction path takes over") is only true if the user also raises the expansion threshold above 1000.

Suggestions: default `optimizer_array_interval_threshold` to INT_MAX so this is an opt-in knob, or only bail out when the disjunction path can actually take over (interval threshold >= expansion threshold). Either way the GUC description needs fixing.

src/backend/gporca/libgpopt/include/gpopt/base/COptCtxt.h
```diff
@@ -32,10 +34,67 @@ using UlongToBitSetMap =
...
+//	@struct:
+//		SArrayCnstrCacheKey
+//
+//	@doc:
+//		Composite key for caching CConstraintInterval derived from
+//		ScalarArrayCmp expressions.
+//		Keyed by {operator ptr, colref ptr,infer_nulls_as}.
+//
+//---------------------------------------------------------------------------
+struct SArrayCnstrCacheKey
```
adam8157, line 54: The cache key is {operator pointer, colref, infer_nulls_as}, but the derived constraint also depends on the array child's datums, which belong to the CExpression, not the operator. Operator sharing across rebuilt expressions is common in ORCA (`PopCopyWithRemappedColumns` returns `this` via `PopCopyDefault`, and many places rebuild expressions with `pop->AddRef()` + new children). Today no in-tree rewrite pairs the same arraycmp operator with a different array, so this isn't currently reachable, but any future rewrite (e.g. array constant folding) would silently hit a stale cached constraint and produce wrong plans.

Cheap fix: include the array child pointer in the key, or at least `GPOS_ASSERT` on cache hit that the recorded array matches `CUtils::PexprScalarArrayChild(pexpr)`.

src/backend/gporca/libgpopt/include/gpopt/base/COptCtxt.h
```diff
@@ -32,10 +34,67 @@ using UlongToBitSetMap =
...
+{
+	COperator *m_pop;
+	const CColRef *m_pcr;
+	BOOL m_infer_nulls_as;
+
+	// Pins the operator pointer for the lifetime of the key.
+	SArrayCnstrCacheKey(COperator *pop, const CColRef *pcr, BOOL infer_nulls_as)
```
adam8157, line 61: `SArrayCnstrCacheKey` AddRefs in the ctor and Releases in the dtor, but copy constructor/assignment aren't deleted. An accidental by-value copy would compile fine and double-Release the pinned operator. Please add `SArrayCnstrCacheKey(const SArrayCnstrCacheKey &) = delete;` (and assignment).

adam8157:
Sorry, this was on my list, then I got busy and forgot...

PR looks good, I post several minor concerns.

PR #155, opened 2026-05-21
jit: fix build with LLVM 21

adam8157:
LLVM 21 made two source-incompatible API changes that broke compilation
of the JIT layer. Add version guards so the same source builds against
both LLVM <21 and LLVM 21.

1. llvm::GlobalValue::getGUID(StringRef) was renamed to
   getGUIDAssumingExternalLinkage(). Used in llvm_build_inline_plan() to
   look up function summaries by GUID.

2. llvm::orc::RTDyldObjectLinkingLayer's GetMemoryManagerFunction type
   changed from
       unique_function<unique_ptr<MemoryManager>()>
   to
       unique_function<unique_ptr<MemoryManager>(const MemoryBuffer&)>
   The lambda passed to its constructor now takes a const MemoryBuffer&.
   This affects LLVMOrcCreateRTDyldObjectLinkingLayerWithSafeSectionMemoryManager(),
   which is compiled unconditionally when USE_LLVM_BACKPORT_SECTION_MEMORY_MANAGER
   is defined.
 
 Cherry-picked two upstream commits to fix them.

### DIFF
--- a/src/backend/jit/llvm/llvmjit_inline.cpp
+++ b/src/backend/jit/llvm/llvmjit_inline.cpp
@@ -243,7 +243,11 @@ llvm_build_inline_plan(LLVMContextRef lc, llvm::Module *mod)
 
 		llvm_split_symbol_name(symbolName.data(), &cmodname, &cfuncname);
 
+#if LLVM_VERSION_MAJOR >= 21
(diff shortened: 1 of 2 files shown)


SKeyerror (approved):
LGTM.

PR #156, opened 2026-05-22
ci: stabilize WHPG6 installcheck workflow

pravesh-sharma:
whpg-ci.yml:
- Limit matrix to EL8 only. WHPG 6 needs Python 2 (PyGreSQL 4.0, gpdemo, xerces-c 3.1 build script) which EL9 does not ship; EL9 coverage lives in the warehouse-pg-packaging pipeline.
- Bump installcheck timeout 120 -> 180 minutes.
- Configure git safe.directory for both root and gpadmin before any git commands run (fixes "dubious ownership" failures in configure and the test phase that runs as gpadmin).
- Checkout with fetch-depth: 1024 and fetch-tags: true so version derivation works without the cost of a full clone.
- Pre-install gpload's Python 2 test deps (pytest_requirement.txt, pytest-order==0.11.0, lxml) as root. The Makefile pip step later runs under gpadmin and cannot write to site-packages; installing here makes those pip calls a no-op.
- Enable WITH_MIRRORS=true for the demo cluster (closer to production topology; required by some isolation2 tests).
- Drop the redundant 'alternatives --set python /usr/bin/python3' step — plpython uses python_majorversion=3 directly.
- Pass TEST_TARGET, MAKE_FLAGS, and PGOPTIONS to run-installcheck.bash as separate environment variables instead of a single MAKE_TEST_COMMAND string; avoids quoting bugs around PGOPTIONS="-c optimizer=on".

run-installcheck.bash:
- Replace MAKE_TEST_COMMAND with required TEST_TARGET / MAKE_FLAGS / PGOPTIONS env vars and apply them to the make invocation.
- Add apply_ci_test_skips() which comments out flaky tests in the runner's copy of src/test/isolation2/isolation2_schedule with a '# CI-SKIP:' prefix. The repo copy is untouched, so local 'make installcheck-world' still runs everything. Currently skipped: fts_segment_reset, pg_rewind_fail_missing_xlog.

README.md:
- Document the EL8-only matrix decision, the CI-only test skip mechanism, and the list of skipped tests with reasons.

- [ ] Document changes

### DIFF
--- a/.github/scripts/run-installcheck.bash
+++ b/.github/scripts/run-installcheck.bash
@@ -7,7 +7,9 @@ set -eox pipefail
 # Required configuration (from workflow env)
 : "${WHPG_SRC:?WHPG_SRC not set}"
 : "${RESULTS_DIR:?RESULTS_DIR not set}"
-: "${MAKE_TEST_COMMAND:?MAKE_TEST_COMMAND not set}"
+: "${TEST_TARGET:?TEST_TARGET not set}"
+: "${MAKE_FLAGS:?MAKE_FLAGS not set}"
+: "${PGOPTIONS:?PGOPTIONS not set}"
 : "${WHPG_MAJORVERSION:?WHPG_MAJORVERSION not set}"
 
 # Source environment explicitly (no login shell / .bash_profile dependency)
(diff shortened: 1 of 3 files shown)


.github/scripts/run-installcheck.bash
adam8157: `export RESULTS_DIR TEST_TARGET MAKE_FLAGS PGOPTIONS WHPG_MAJORVERSION WHPG_SRC`

adam8157: No more `MAKE_TEST_COMMAND`

adam8157 (approved):
Two minor issues, give my approval in advance.

PR #172, opened 2026-06-05
Fix CVE-2026-6637 (refint SQL injection / buffer overrun) [PTT-1190]

adam8157:
## Summary
Backport the fix for **CVE-2026-6637** — stack buffer overflow + SQL injection in the `refint` contrib module (`check_foreign_key` / `check_primary_key`) — to GP6 (PostgreSQL 9.4 base). Missed by PTT-1125's evaluation of the May 14, 2026 PostgreSQL security release.

Two commits:
- **refint: Fix SQL injection and buffer overruns.** — upstream `2b026df` (Security: CVE-2026-6637); replaces the fixed `sql[8192]` stack buffers with `StringInfo` and escapes all values via `quote_literal_cstr()`.
- **Remove dependency to system calls for memory allocation in refint** — prerequisite upstream `b0b619`, never picked up on this branch.

## Background
GP6's `refint.c` carried the same vulnerable fixed `sql[8192]` buffers and unescaped value interpolation as v7. The CVE-relevant changes are semantically identical to upstream; pre-existing GP6-only divergences in this file (older `%d`/`SPI_result` error strings, `pg_atoi`, `UINT64_FORMAT`) are left intact, so unlike the v7 PRs the result is not byte-identical to upstream REL_14_STABLE but is functionally equivalent.

Not applicable to GP6/9.4 from the same release: CVE-2026-6479 (SSL/GSS recursion) — 9.4 has no GSS startup negotiation path, so the vuln does not exist here.

## Test plan
- `contrib/spi` compiles clean (`-Wall -Werror -DREFINT_VERBOSE`, no warnings).

Refs: PTT-1190

### DIFF
--- a/contrib/spi/refint.c
+++ b/contrib/spi/refint.c
@@ -12,6 +12,7 @@
 #include "commands/trigger.h"
 #include "executor/spi.h"
 #include "utils/builtins.h"
+#include "utils/memutils.h"
 #include "utils/rel.h"
 
 PG_MODULE_MAGIC;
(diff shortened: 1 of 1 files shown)


linwen (approved):
Looks good to me.

PR #174, opened 2026-06-05
Authenticate QD-to-QE internal connections with mutual TLS

adam8157:
## Summary

Add mutual-TLS authentication for QD-to-QE internal connections (query dispatch
and FTS), behind a new `gp_internal_tls` GUC (`disable`/`verify-ca`, default
`disable`). The default `disable` keeps the legacy behavior for compatibility;
`verify-ca` replaces the historical authentication bypass with real credentials.
Design details are in the commit message.

## Background

Internal connections were identified only by a magic bit in the protocol version
and were then granted access with no authentication -- unconditionally on a
segment -- so any host able to reach a segment's postgres port could obtain
superuser access with zero credentials simply by setting that bit. This change
makes it possible to authenticate those connections with mutual TLS instead:
under `gp_internal_tls = verify-ca` the bit grants nothing without a client
certificate that verifies against the internal cluster CA, demoting it to a
connection-type hint. The default `disable` preserves the legacy behavior. The
internal (intra-cluster) credentials are kept separate from the external
server-SSL trust domain via `gp_internal_tls_cert_file`/`_key_file`/`_ca_file`
(no fallback), and the accepting node verifies internal peers against
`gp_internal_tls_ca_file` rather than `ssl_ca_file`. Drop-in: no catalog, ABI or
WAL changes.

## Test plan

- `src/test/ssl/t/003_gp_internal_tls.pl`: `disable` and `verify-ca`, with the
  internal and external CAs configured as distinct trust domains -- a
  certificate that verifies only against the internal cluster CA is accepted, one
  that verifies only against `ssl_ca_file` is rejected, plus the no-certificate
  and plaintext rejection paths.
- Optional end-to-end on a demo cluster via
  `WITH_INTERNAL_TLS=1 gpAux/gpdemo/demo_cluster.sh`.


reshke:
> ## Summary
> a new `gp_internal_tls` GUC (`off`/`prefer`/`require`, default `off`). Design details are in the commit message.

Maybe not `off` but `disable`, to match existing PostgreSQL sslmode?
What about `allow`?

adam8157:
> Maybe not off but disable, to match existing PostgreSQL sslmode?
> What about allow?

@reshke Good call -- I've reworked the GUC to `disable` / `verify-ca` and dropped the `off`/`prefer`/`require` naming.

- `disable` follows the sslmode vocabulary as you suggested, and paired with `verify-ca` it reads as one consistent set rather than mixing `off` with sslmode terms.
- I named the enforce state `verify-ca` rather than `require` because that is exactly the libpq `sslmode` it resolves to on the dialing side (chain-verified against the cluster CA). libpq's own `require` encrypts *without* verifying the certificate, so reusing that name here would have been misleading.

On `allow`/`prefer`: I deliberately left them out. Both mean "authenticate if possible, otherwise silently fall back to plaintext" -- an automatic, silent security downgrade, which is exactly what this change is trying to remove. The policy is binary: `disable` (legacy) or `verify-ca` (enforced).

reshke (approved):

PR #177, opened 2026-06-09
fix trailing-whitespace mismatch in expected/gp_dqa

24nishant:
The "count" column header for two count(*) FROM (... sales ...) queries added by #149 was emitted without a trailing space in the expected file, but psql actually outputs " count " with one trailing space. This makes the diff fail under optimizer=off (the gp_dqa.out path; the gp_dqa_optimizer.out sibling was corrected before the merge of #149).

Two-character fix at lines 3656 and 3698.

The bug surfaces in main's own post-merge CI run for #149 (installcheck / EL8|9 / optimizer=off)


linwen (approved):
looks good to me.

PR #178, opened 2026-06-16
fix: uninitialized current table oid in execCurrentOf

gruuya:
There's a potential QD path on non-partitioned tables that may read uninitialized values otherwise.

In addition, the `getversion` script is updated to take into account lightweight tags too. Without that `version()` would give a misleading value, i.e. give `7.2.1` (the last annotated tag) even on later versions.

- [ ] Document changes

### DIFF
--- a/getversion
+++ b/getversion
@@ -10,7 +10,7 @@ BUILDNUMBER=dev
 # Call git describe, and convert it to a semi-semantic version, like
 # 5.0.0-alpha.0+dev.52.g123abc
 generate_dev_version() {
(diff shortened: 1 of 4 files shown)


adam8157 (dismissed):
Yes, we patched the getversion issue during packaging as below, but it should also be fixed properly in the codebase.

```
if ! grep -q "git describe --tags" ./src/getversion; then
        sed -i -e "s/git describe/git describe --tags/" ./src/getversion
fi
```

adam8157 (approved):

PR #182, opened 2026-06-23
feature: Add gpAdminLogs logging support to gpcheckperf

Bonartze:
### Summary
`gpcheckperf` only wrote its output to stdout, leaving no historical records once the terminal scrollback was gone.

### Key Changes
* **Logging Integration:** Initialized `gplog.setup_tool_logging` early in `main()` so the resolved command line, banners and the full `RESULT` block are automatically captured
* **Stdout Mirroring:** Added a thin module-level wrapper around the builtin `print()` that calls `gplog.log_to_file_only()`. This ensures that stdout output remains byte-identical to before (no double printing in the console), while everything is mirrored to `~/gpAdminLogs/gpcheckperf_<date>.log`
* **Documentation:** Updated `gpcheckperf_help` to mention the new logfile location and append behavior
* **Testing:** Added a new unit test `test_output_is_mirrored_to_logfile` in `test_unit_gpcheckperf.py` to assert logfile creation and correct content capturing

### DIFF
--- a/gpMgmt/bin/gpcheckperf
+++ b/gpMgmt/bin/gpcheckperf
@@ -30,18 +30,24 @@ Usage: gpcheckperf <options>
     --duration : how long to run network test (default 5 seconds)
     --netperf  : use netperf instead of gpnetbenchServer/gpnetbenchClient
     --buffer-size : the size of the send buffer in kilobytes ( default 8 kilobytes)
+
+All output is also written to ~/gpAdminLogs/gpcheckperf_<date>.log
 """
 
 import datetime
+import logging
 import os
 import sys
 import re
+import builtins
 
 sys.path.append(sys.path[0] + '/lib')
 try:
     import getopt, math, io, stat, subprocess, signal
+    from gppylib import gplog
     from gppylib.gpparseopts import OptParser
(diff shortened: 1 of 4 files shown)


adam8157:
The unit test failed:
`PYTHONPATH=$PWD/gpMgmt/bin:$PWD/gpMgmt/bin/lib GPHOME=$PWD/gpMgmt USER=... python3 gpMgmt/bin/gppylib/test/unit/test_unit_gpcheckperf.py`

We also have behave tests for gpcheckperf, which is not covered by the public CI, please check.

BTW, better not to reference internal tracking id in public repo.

Bonartze:
> The unit test failed: `PYTHONPATH=$PWD/gpMgmt/bin:$PWD/gpMgmt/bin/lib GPHOME=$PWD/gpMgmt USER=... python3 gpMgmt/bin/gppylib/test/unit/test_unit_gpcheckperf.py`
> 
> We also have behave tests for gpcheckperf, which is not covered by the public CI, please check.
> 
> BTW, better not to reference internal tracking id in public repo.

Fixed and checked two type of tests, now it works fine. Thanks

adam8157 (approved):

PR #183, opened 2026-06-24
fix(orca): avoid crash on ordered-set aggregate without direct args

linwen:
There is a crash when use mode() for ORCA.
```
$ psql
psql (12.12)
Type "help" for help.
gpadmin=# create table test_mode (a int, b text);
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'a' as the Greenplum Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
CREATE TABLE
gpadmin=# set optimizer=off;
SET
gpadmin=# explain select mode() within group (order by b) from test_mode limit 10;
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 Limit  (cost=10000000893.00..10000000893.01 rows=1 width=32)
   ->  Aggregate  (cost=10000000893.00..10000000893.01 rows=1 width=32)
         ->  Gather Motion 4:1  (slice1; segments: 4)  (cost=10000000000.00..10000000769.00 rows=49600 width=32)
               ->  Seq Scan on test_mode  (cost=10000000000.00..10000000149.00 rows=12400 width=32)
(5 rows)
gpadmin=# set optimizer=on;
SET
gpadmin=# explain select mode() within group (order by b) from test_mode limit 10;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
!> \q

```

Reason: CUtils::FHasOrderedAggToSplit() unconditionally dereferenced the direct-args child of an ordered-set aggregate. That child exists but is empty for mode(), the only built-in ordered-set aggregate with no direct argument, so the access tripped a GPOS_ASSERT in debug builds and caused a SIGSEGV in release builds during ORCA query preprocessing.

Guard the ordered aggregate child accesses and bail out early when the args, direct-args, or order children are empty. This leaves mode() as a regular ordered-set Aggref for ORCA to plan, instead of trying to split it into the gp_percentile-specific rewrite.

Add a minidump test and sql regress cases covering mode().


src/test/regress/sql/percentile.sql
```diff
@@ -45,6 +45,10 @@ select percentile_cont(0.5) within group (order by a),
 	median(a), percentile_disc(0.5) within group(order by a) from perct;
 select b, percentile_cont(0.5) within group (order by a),
 	median(a), percentile_disc(0.5) within group(order by a) from perct group by b order by b;
+-- mode() is the only ordered-set aggregate with no direct argument; ORCA must
+-- plan it as an ordered-set aggregate instead of crashing in preprocessing.
+select mode() within group (order by b) from perct;
+select b, mode() within group (order by a) from perct group by b order by b;
```
24nishant, line 51: Should we consider adding test case of mode() with normal agg , splitable percentile ?
I think this would be a new code path which is not tested till now as mode() cause a crash. These test should cover those paths.

24nishant (commented):
Nice fix, the changes look good to me.

src/test/regress/sql/percentile.sql
linwen: Thanks, added.

24nishant (approved):
Looks good to me , thanks!

PR #184, opened 2026-06-24
fix(planner): lift correlated quals above Motion for Entry tables to prevent crash

Bonartze:
## Summary
Fixes a backend crash that occurs when executing correlated subqueries that join a coordinator-only catalog table (such as `gp_segment_configuration`) with a distributed table.

## Step to reproduce
```sql
create table gp_segment(a int, b text);
create table gp_tmp(a int, b int, c int);
insert into gp_tmp select i, i, i from generate_series(1, 10) i;
insert into gp_segment select 1, hostname from gp_segment_configuration limit 1;

-- This query used to crash the backend:
select (select count(*) from gp_segment_configuration s join gp_tmp t2 on s.dbid = t2.b where s.hostname = gp_segment.b) from gp_segment;
```
## Solution
This PR introduces a mechanism to keep Entry-correlated subqueries local to the coordinator where the parameter is available:

1. **Subquery Detection:** Added `subquery_contains_distributed_rel()` to detect if a subquery scans any partitioned or replicated relations that require a `Motion`.
2. **Forcing Single QE:** In `initsplan.c`, if a correlated restriction targets an Entry table and the subquery contains a distributed relation, we set `root->config->force_singleQE = true`.
3. **Lifting above Motion:** In `grouping_planner()`, we use `tlist_has_entry_correlated_subplan()` to check if the final target list references such a subplan. If found, we explicitly inject a `Gather Motion` to pull the scan/join results to the coordinator (`entryLocus`) **before** the target list is evaluated. This ensures the SubPlan runs above the final Motion where the correlation parameter is natively accessible.

### DIFF
--- a/src/test/regress/sql/subselect_gp.sql
+++ b/src/test/regress/sql/subselect_gp.sql
@@ -1684,3 +1684,90 @@ select count(*) from foo where a > (select regr_count(x, y) from bar where foo.a
 
 drop table foo;
 drop table bar;
+
+--
+-- EntryDB path: a correlated subquery that joins a coordinator-only (Entry)
+-- catalog table (gp_segment_configuration) with a distributed table.  Such a
+-- SubPlan must execute on the coordinator: it scans an Entry catalog (whose
+-- rows only exist there) and needs the outer correlation parameter, which GPDB
+-- cannot pass across a Motion.  A correlated SubPlan runs in the slice that
+-- produces its parameter, so when the outer query scans the parameter-producing
+-- table on the segments, the SubPlan runs there too -- scanning the catalog on
+-- the segments (wrong/empty data) or crashing when the parameter crosses a
+-- Motion.  The planner now forces the outer query's base rels to Entry
+-- (coordinator) locus (force_entry, see distribute_restrictinfo_to_rels() and
+-- bring_to_entry()), so the parameter is produced on the coordinator and the
+-- SubPlan is evaluated above the gather Motion, where the catalog is local.
+-- This is exercised below with the SubPlan in the target list, in WHERE and in
+-- HAVING.
+--
+create table gp_entry_outer(x int) distributed by (x);
+insert into gp_entry_outer values (1), (2), (3);
+create table gp_tmp(a int, b int, c int) distributed by (a);
+insert into gp_tmp select i, i, i from generate_series(1, 10) i;
+create table gp_segment(a int, b text) distributed by (a);
+insert into gp_segment values (1, 'no_such_host');
+-- Correlated on dbid: for each outer x, count catalog rows whose dbid = x and
+-- whose dbid also appears in the distributed gp_tmp (1..10).  dbids 1, 2 and 3
+-- always exist, so each outer row yields exactly 1 -- deterministic regardless
+-- of the cluster's segment count.
+select x, (select count(*)
+           from gp_segment_configuration s
+           join gp_tmp t2 on s.dbid = t2.b
+           where s.dbid = gp_entry_outer.x) as cnt
+from gp_entry_outer order by x;
+-- The original report shape (correlated on hostname).  Use a hostname that does
+-- not exist so the count is a deterministic 0 on any cluster.
+select (select count(*)
+        from gp_segment_configuration s
+        join gp_tmp t2 on s.dbid = t2.b
+        where s.hostname = gp_segment.b) from gp_segment;
+-- Same correlated SubPlan in a WHERE clause: the SubPlan-bearing qual is lifted
+-- above the Entry gather.  All three outer rows qualify (count 1 > 0).
+select x from gp_entry_outer o
+where (select count(*)
+       from gp_segment_configuration s
+       join gp_tmp t2 on s.dbid = t2.b
+       where s.dbid = o.x) > 0
+order by x;
+-- Same WHERE-clause SubPlan under a bare count(*), where no column of the outer
+-- rel is otherwise selected.  The correlation Var must still be carried through
+-- the gather Motion (regression: "variable not found in subplan target list").
+-- All three outer rows qualify, so the count is 3.
+--
+-- Also lock the plan shape so a future change that silently moves the SubPlan
+-- back below the Motion is caught even if it no longer errors.  The fix is in
+-- the Postgres planner, so scope the EXPLAIN to it with SET LOCAL inside a
+-- transaction (self-restoring); that way both the planner and the ORCA expected
+-- files show the same plan.  costs are omitted -- they are not stable across
+-- cluster sizes.
+begin;
+set local optimizer = off;
+explain (costs off)
+select count(*) from gp_entry_outer o
+where (select count(*)
+       from gp_segment_configuration s
+       join gp_tmp t2 on s.dbid = t2.b
+       where s.dbid = o.x) > 0;
+rollback;
+select count(*) from gp_entry_outer o
+where (select count(*)
+       from gp_segment_configuration s
+       join gp_tmp t2 on s.dbid = t2.b
+       where s.dbid = o.x) > 0;
+-- ... and in a HAVING clause, correlated on the grouping key.
+select o.x from gp_entry_outer o
+group by o.x
+having (select count(*)
+        from gp_segment_configuration s
+        join gp_tmp t2 on s.dbid = t2.b
+        where s.dbid = o.x) > 0
+order by o.x;
+-- A correlated scan of the Entry catalog with no join (no Motion in the
+-- subplan) must keep working unchanged.
+select (select count(*)
+        from gp_segment_configuration s
+        where s.hostname = gp_segment.b) from gp_segment;
+drop table gp_entry_outer;
+drop table gp_segment;
+drop table gp_tmp;
--- a/.abi-check/7.0.0/postgres.types.ignore
+++ b/.abi-check/7.0.0/postgres.types.ignore
@@ -8,3 +8,4 @@ struct AppendOnlyFetchDescData
 struct AppendOnlyInsertDescData
 struct CreateForeignTableStmt
 enum PMSignalReason
+struct PlannerConfig
--- a/.abi-check/7.1.0/postgres.types.ignore
+++ b/.abi-check/7.1.0/postgres.types.ignore
@@ -8,3 +8,4 @@ struct AOCSWriteColumnDescData
 struct AppendOnlyFetchDescData
 struct AppendOnlyInsertDescData
 struct ResGroupCaps
+struct PlannerConfig
--- a/src/backend/optimizer/path/allpaths.c
+++ b/src/backend/optimizer/path/allpaths.c
@@ -161,6 +161,7 @@ static void remove_unused_subquery_outputs(Query *subquery, RelOptInfo *rel);
 
 static void bring_to_outer_query(PlannerInfo *root, RelOptInfo *rel, List *outer_quals);
 static void bring_to_singleQE(PlannerInfo *root, RelOptInfo *rel);
+static void bring_to_entry(PlannerInfo *root, RelOptInfo *rel);
 static bool is_query_contain_limit_groupby(Query *parse);
 static void handle_gen_seggen_volatile_path(PlannerInfo *root, RelOptInfo *rel);
 
@@ -646,6 +647,138 @@ bring_to_singleQE(PlannerInfo *root, RelOptInfo *rel)
 }
 
 /*
+ * The following function "steals" ideas and most of the code from the
+ * function bring_to_singleQE.
+ *
+ * Decorate the Paths of 'rel' with Motions to bring the relation's result to
+ * Entry (coordinator) locus.  This is used when the query references a
+ * correlated SubPlan that must execute on the coordinator -- it scans a
+ * coordinator-only (Entry) catalog and needs an outer correlation parameter,
+ * which cannot be passed across a Motion.  A correlated SubPlan runs in the
+ * slice that produces its parameter, so the rel that produces the parameter
+ * must be gathered to the coordinator here; the SubPlan, applied above this
+ * Motion, then runs on the coordinator where the catalog is local and the
+ * parameter is available.  See force_entry in distribute_restrictinfo_to_rels().
+ *
+ * Two kinds of quals must be evaluated above the Motion rather than at the
+ * scan (which is below the Motion, on the segments):
+ *
+ *  - A correlated SubPlan that sits in a WHERE qual lives in
+ *    rel->baserestrictinfo.  Pull the SubPlan-bearing quals out so they are
+ *    applied above the Motion; plain quals (no SubPlan) stay at the scan, where
+ *    they can still filter before the gather.  (SubPlans in the target list or
+ *    in HAVING are already evaluated above the scan, hence above this Motion,
+ *    so they need no special handling here.)
+ *
+ *  - Outer-correlated quals that rel_need_to_separate_outer_query_restrictinfos()
+ *    diverted to rel->upperrestrictinfo (e.g. an equivalence-class-derived
+ *    "t2.b = outer.x" filter on a distributed rel).  Normally bring_to_outer_query()
+ *    applies these above the Motion, but the force_entry branch in
+ *    set_rel_pathlist() runs in its place -- so we must apply them here too,
+ *    otherwise they are silently dropped and the subquery's join degenerates
+ *    into a Cartesian product.
+ *
+ * Both, like bring_to_outer_query() does with upperrestrictinfo, are re-applied
+ * in a Result node above the Motion (at Entry locus, where the outer parameter
+ * is available on the coordinator).
+ */
+static void
+bring_to_entry(PlannerInfo *root, RelOptInfo *rel)
+{
+	List	   *origpathlist;
+	List	   *above_motion_quals = NIL;
+	List	   *scan_quals = NIL;
+	ListCell   *lc;
+
+	foreach(lc, rel->baserestrictinfo)
+	{
+		RestrictInfo *ri = (RestrictInfo *) lfirst(lc);
+
+		if (contain_subplans((Node *) ri->clause))
+			above_motion_quals = lappend(above_motion_quals, ri);
+		else
+			scan_quals = lappend(scan_quals, ri);
+	}
+	rel->baserestrictinfo = scan_quals;
+
+	/* Correlated quals separated into upperrestrictinfo go above the Motion. */
+	above_motion_quals = list_concat(above_motion_quals,
+									 list_copy(rel->upperrestrictinfo));
+	rel->upperrestrictinfo = NIL;
+
+	origpathlist = rel->pathlist;
+	rel->cheapest_startup_path = NULL;
+	rel->cheapest_total_path = NULL;
+	rel->cheapest_unique_path = NULL;
+	rel->cheapest_parameterized_paths = NIL;
+	rel->pathlist = NIL;
+
+	foreach(lc, origpathlist)
+	{
+		Path	     *origpath = (Path *) lfirst(lc);
+		Path	     *path;
+		CdbPathLocus  target_locus;
+
+		if (CdbPathLocus_IsGeneral(origpath->locus) ||
+			CdbPathLocus_IsEntry(origpath->locus))
+			path = origpath;
+		else
+		{
+			/*
+			 * Cannot pass a param through motion, so if this is a parameterized
+			 * path, we can't use it.
+			 */
+			if (origpath->param_info)
+				continue;
+
+			/*
+			 * param_info cannot cover the case that an index path's orderbyclauses
+			 * See github issue: https://github.com/greenplum-db/gpdb/issues/9733
+			 */
+			if (IsA(origpath, IndexPath))
+			{
+				IndexPath *ipath = (IndexPath *) origpath;
+				if (contains_outer_params((Node *) ipath->indexorderbys,
+										  (void *) root))
+					continue;
+			}
+
+			CdbPathLocus_MakeEntry(&target_locus);
+
+			path = cdbpath_create_motion_path(root,
+											  origpath,
+											  NIL, // DESTROY pathkeys
+											  false,
+											  target_locus);
+
+			/*
+			 * Shield the Motion with an (unparameterized) Materialize, exactly
+			 * as bring_to_singleQE() does.  A correlated SubPlan is rescanned
+			 * once per outer row, and the lifted correlated qual we attach just
+			 * below makes the path above this Motion parameterized.  Without the
+			 * Materialize here -- whose subtree is just the Motion and therefore
+			 * has no parameter dependency -- a rescan would propagate down to
+			 * the Motion itself ("illegal rescan of motion node").  With it, the
+			 * Motion runs once and rescans replay the materialized tuples.
+			 */
+			path = (Path *) create_material_path(root, rel, path);
+		}
+
+		/* Evaluate the lifted quals (SubPlan-bearing + correlated) above the Motion. */
+		if (above_motion_quals)
+			path = (Path *) create_projection_path_with_quals(root,
+															  rel,
+															  path,
+															  path->parent->reltarget,
+															  above_motion_quals,
+															  true);
+
+		add_path(rel, path);
+	}
+	set_cheapest(rel);
+}
+
+/*
  * handle_gen_seggen_volatile_path
  *
  * Only use for base replicated rel.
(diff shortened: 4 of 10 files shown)


adam8157:
Thanks for this — the root-cause analysis is accurate. I traced it through and agree: `rel_need_to_separate_outer_query_restrictinfos()` (initsplan.c) returns true for partitioned/replicated rels, so outer-correlated restrictions on those go to `upperrestrictinfo` → `bring_to_outer_query()` and are evaluated above the Motion on the coordinator. For an **Entry** table (NULL/Entry policy) it returns false, so the correlated qual lands in `baserestrictinfo`; once that Entry table is joined with a distributed table, the join needs a Broadcast Motion and the qual gets pushed below it onto the segments — and we can't pass a parameter across a Motion, hence the crash. Using `force_singleQE` here is also consistent with the existing precedent in `allpaths.c` (LATERAL + limit/groupby).

A few things I'd like to sort out before this goes in:

**1. Why not reuse the existing `upperrestrictinfo` / `bring_to_outer_query()` path?** Since the gap is precisely that `rel_need_to_separate_outer_query_restrictinfos()` returns false for the Entry case, the most surgical fix would seem to be returning true for "Entry table + subquery contains a distributed rel" and letting the existing mechanism lift the restriction above the Motion — rather than introducing a new `force_singleQE` flag plus in-place pathlist rewriting in `grouping_planner()`. Did you try that route? If it doesn't suffice (I suspect because the Entry table is joined with a distributed rel and the whole join can't be lifted by `upperrestrictinfo` alone), please add a note explaining why.

**2. Completeness — only `processed_tlist` is handled.** `tlist_has_entry_correlated_subplan()` only walks the final target list. What about the same correlated subplan appearing in the outer query's WHERE/HAVING/ORDER BY, e.g.:

```sql
select * from gp_segment o
where (select count(*) from gp_segment_configuration s join gp_tmp t on s.dbid = t.b
       where s.hostname = o.b) > 0;
```

Is that covered/tested, or should it be called out as a follow-up?

**3. ORCA expected file.** The change is entirely in the Postgres planner, but `subselect_gp_optimizer.out` was updated too. Can you confirm those results come from ORCA with `optimizer=on` and not a planner fallback?

A few concrete code points:

- **`grouping_planner()` rewrites `pathlist` in place** (`lfirst(lc) = gathered;`). The established idiom for this kind of locus adjustment (`bring_to_singleQE()`, `handle_gen_seggen_volatile_path()`) clears the pathlist, re-adds via `add_path()`, then `set_cheapest()`, which preserves the dedup/dominance invariants. Please follow that pattern.
- **No parameterized-path guard.** `bring_to_singleQE()` explicitly skips paths with `param_info` before wrapping them in a Motion (you can't motion a parameterized path). The new loop doesn't check `subpath->param_info`. At the top-level rel this is probably moot, but it's inconsistent with the safety check elsewhere — please add a guard or a comment on why it's unnecessary.
- **Header include order** in planner.c: `catalog/gp_distribution_policy.h` and `optimizer/walkers.h` are inserted after `optimizer/orca.h`, breaking the alphabetical grouping. Minor.
- The tests are nicely deterministic across cluster sizes and include the no-Motion control case — good. Consider adding an `EXPLAIN` to lock the plan shape so a future regression that silently changes the locus is caught even if the result happens to stay correct.

Bonartze:
> Thanks for this — the root-cause analysis is accurate. I traced it through and agree: `rel_need_to_separate_outer_query_restrictinfos()` (initsplan.c) returns true for partitioned/replicated rels, so outer-correlated restrictions on those go to `upperrestrictinfo` → `bring_to_outer_query()` and are evaluated above the Motion on the coordinator. For an **Entry** table (NULL/Entry policy) it returns false, so the correlated qual lands in `baserestrictinfo`; once that Entry table is joined with a distributed table, the join needs a Broadcast Motion and the qual gets pushed below it onto the segments — and we can't pass a parameter across a Motion, hence the crash. Using `force_singleQE` here is also consistent with the existing precedent in `allpaths.c` (LATERAL + limit/groupby).
> 
> A few things I'd like to sort out before this goes in:
> 
> **1. Why not reuse the existing `upperrestrictinfo` / `bring_to_outer_query()` path?** Since the gap is precisely that `rel_need_to_separate_outer_query_restrictinfos()` returns false for the Entry case, the most surgical fix would seem to be returning true for "Entry table + subquery contains a distributed rel" and letting the existing mechanism lift the restriction above the Motion — rather than introducing a new `force_singleQE` flag plus in-place pathlist rewriting in `grouping_planner()`. Did you try that route? If it doesn't suffice (I suspect because the Entry table is joined with a distributed rel and the whole join can't be lifted by `upperrestrictinfo` alone), please add a note explaining why.
> 
> **2. Completeness — only `processed_tlist` is handled.** `tlist_has_entry_correlated_subplan()` only walks the final target list. What about the same correlated subplan appearing in the outer query's WHERE/HAVING/ORDER BY, e.g.:
> 
> ```sql
> select * from gp_segment o
> where (select count(*) from gp_segment_configuration s join gp_tmp t on s.dbid = t.b
>        where s.hostname = o.b) > 0;
> ```
> 
> Is that covered/tested, or should it be called out as a follow-up?
> 
> **3. ORCA expected file.** The change is entirely in the Postgres planner, but `subselect_gp_optimizer.out` was updated too. Can you confirm those results come from ORCA with `optimizer=on` and not a planner fallback?
> 
> A few concrete code points:
> 
> * **`grouping_planner()` rewrites `pathlist` in place** (`lfirst(lc) = gathered;`). The established idiom for this kind of locus adjustment (`bring_to_singleQE()`, `handle_gen_seggen_volatile_path()`) clears the pathlist, re-adds via `add_path()`, then `set_cheapest()`, which preserves the dedup/dominance invariants. Please follow that pattern.
> * **No parameterized-path guard.** `bring_to_singleQE()` explicitly skips paths with `param_info` before wrapping them in a Motion (you can't motion a parameterized path). The new loop doesn't check `subpath->param_info`. At the top-level rel this is probably moot, but it's inconsistent with the safety check elsewhere — please add a guard or a comment on why it's unnecessary.
> * **Header include order** in planner.c: `catalog/gp_distribution_policy.h` and `optimizer/walkers.h` are inserted after `optimizer/orca.h`, breaking the alphabetical grouping. Minor.
> * The tests are nicely deterministic across cluster sizes and include the no-Motion control case — good. Consider adding an `EXPLAIN` to lock the plan shape so a future regression that silently changes the locus is caught even if the result happens to stay correct.

Thank you for the detailed and insightful review. 

Here is the breakdown of how each item is addressed now:

## 1. Why Not Reuse `upperrestrictinfo` / `bring_to_outer_query()`?
Your suspicion was correct. 
The new implementation solves this holistically by introducing a clean `force_entry` pipeline:
* **Locus Alignment:** During `SS_process_sublinks` (in `initsplan.c`), when the Entry-correlated pattern is detected, we set `force_entry = true` on both the subquery and the outer query (`root->parent_root`).
* **Path Generation:** In `allpaths.c`, `set_rel_pathlist()` invokes a new `bring_to_entry()` helper (modeled after `bring_to_singleQE()`). This gathers the outer relation paths to the **Entry locus (coordinator)** from the very beginning, ensuring the execution slice is perfectly aligned.
* **Subquery Propagation:** The `force_entry` constraint now propagates down into the subquery level (removing the previous reset in `subselect.c`). This forces internal distributed relations (like `gp_tmp`) to safely gather to the coordinator as well, making the subquery's internal join an `Entry-vs-Entry` join on the coordinator and preventing hazardous Broadcast Motions of catalog tables onto segments.

## 2. WHERE and HAVING Clauses
They are now fully covered

As you noted, simply forcing the relation to Entry locus would leave `WHERE` and `HAVING` subplans trapped inside `baserestrictinfo` below the Entry gather. To fix this, `bring_to_entry()` explicitly splits `rel->baserestrictinfo`:
* Qualifications containing subplans (`contain_subplans`) are pulled out and re-applied inside a `Result` node **above** the Entry gather Motion (where the parameter is safely available on the coordinator).
* Plain filters remain at the scan level to filter rows early on the segments.
* For equivalence-class-derived filters (like `t2.b = outer.x`), `bring_to_entry()` also pulls them from `rel->upperrestrictinfo` and applies them above the Motion, avoiding Cartesian products.

The regression tests (`subselect_gp.sql`) have been restored and expanded to include your exact `WHERE` clause shape as well as a `HAVING` clause scenario. All queries now execute successfully with correct data and zero crashes.

## 3. ORCA Expected Files 
I have verified the `subselect_gp_optimizer.out` results on a live multi-segment cluster. For this specific shape of Entry-catalog correlated subquery joins, ORCA natively falls back to the Postgres-based planner (producing the explicit `Planner produced plan` state). Updating the optimizer expected file ensures this fallback path remains locked and verified against any future structural modifications.

**ABI Compliance Fix:**
1. **Field Reordering:** Moved the new `force_entry` flag to the very end of `struct PlannerConfig` in `plannerconfig.h`. This preserves the byte offsets of all pre-existing fields, resolving the binary compatibility drop.
2. **Type Whitelisting:** Added `struct PlannerConfig` to `.postgres.types.ignore` for both 7.0.0 and 7.1.0 baselines to whitelist the structural size change (`sizeof`)

SKeyerror:
I found some issues when attempting to perform performance regression testing:


```sql
create table big_dist(id int, v int) distributed by (id);
insert into big_dist select i, i from generate_series(1, 1000000) i;
create table big_dist2(id int, v int) distributed by (id);
insert into big_dist2 select i, i from generate_series(1, 1000000) i;

create table gp_tmp(a int, b int, c int) distributed by (a);
insert into gp_tmp select i, i, i from generate_series(1, 10) i;

```

The below SQL will ERROR out:


```sql
explain (costs on)
select count(*) from big_dist b
where (select count(*)
       from gp_segment_configuration s
       join gp_tmp t2 on s.dbid = t2.b
       where s.dbid = b.id) > 0;
ERROR:  variable not found in subplan target list (setrefs.c:3363)

-- which should be
                                                                  QUERY PLAN                                                                  
----------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=4812.50..4812.51 rows=1 width=8)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=4812.44..4812.49 rows=3 width=8)
         ->  Partial Aggregate  (cost=4812.44..4812.45 rows=1 width=8)
               ->  Seq Scan on big_dist b  (cost=0.00..4534.67 rows=111111 width=0)
                     Filter: ((SubPlan 1) > 0)
                     SubPlan 1
                       ->  Aggregate  (cost=10000003475.63..10000003475.64 rows=1 width=8)
                             ->  Nested Loop  (cost=10000000000.00..10000003280.88 rows=77900 width=0)
                                   ->  Seq Scan on gp_segment_configuration s  (cost=0.00..1.05 rows=1 width=2)
                                         Filter: (dbid = b.id)
                                   ->  Materialize  (cost=0.00..2890.33 rows=77900 width=4)
                                         ->  Result  (cost=0.00..2500.83 rows=77900 width=4)
                                               Filter: (t2.b = b.id)
                                               ->  Materialize  (cost=0.00..1721.83 rows=77900 width=4)
                                                     ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..1332.33 rows=77900 width=4)
                                                           ->  Seq Scan on gp_tmp t2  (cost=0.00..293.67 rows=25967 width=4)
(17 rows)
```


------

`force_entry` is a coarse hammer:


```sql
-- force_entry takes effect at the query-layer level, forcing all base tables in that layer, even if they are unrelated to the correlated parameters.
explain (costs on)
select count(*)
from big_dist b
join big_dist2 c on b.id = c.id
where (select count(*)
       from gp_segment_configuration s
       join gp_tmp t2 on s.dbid = t2.b
       where s.dbid = b.id) > 0;


                                                                QUERY PLAN                                                                 
-------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=48097.11..48097.12 rows=1 width=8)
   ->  Hash Join  (cost=18145.78..47263.78 rows=333333 width=0)
         Hash Cond: (c.id = b.id)
         ->  Materialize  (cost=0.00..22034.67 rows=1000000 width=4)
               ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..17034.67 rows=1000000 width=4)
                     ->  Seq Scan on big_dist2 c  (cost=0.00..3701.33 rows=333333 width=4)
         ->  Hash  (cost=13979.11..13979.11 rows=333333 width=4)
               ->  Result  (cost=0.00..13979.11 rows=333333 width=4)
                     Filter: ((SubPlan 1) > 0)
                     ->  Materialize  (cost=0.00..10645.78 rows=333333 width=4)
                           ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..8979.11 rows=333333 width=4)
                                 ->  Seq Scan on big_dist b  (cost=0.00..4534.67 rows=111111 width=4)
                     SubPlan 1
                       ->  Aggregate  (cost=10000003475.63..10000003475.64 rows=1 width=8)
                             ->  Nested Loop  (cost=10000000000.00..10000003280.88 rows=77900 width=0)
                                   ->  Seq Scan on gp_segment_configuration s  (cost=0.00..1.05 rows=1 width=2)
                                         Filter: (dbid = b.id)
                                   ->  Materialize  (cost=0.00..2890.33 rows=77900 width=4)
                                         ->  Result  (cost=0.00..2500.83 rows=77900 width=4)
                                               Filter: (t2.b = b.id)
                                               ->  Materialize  (cost=0.00..1721.83 rows=77900 width=4)
                                                     ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1332.33 rows=77900 width=4)
                                                           ->  Seq Scan on gp_tmp t2  (cost=0.00..293.67 rows=25967 width=4)
(24 rows)



-- Before this patch
                                                                        QUERY PLAN                                                                        
----------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=12263.84..12263.85 rows=1 width=8)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=12263.78..12263.83 rows=3 width=8)
         ->  Partial Aggregate  (cost=12263.78..12263.79 rows=1 width=8)
               ->  Hash Join  (cost=5923.56..11986.00 rows=111111 width=0)
                     Hash Cond: (c.id = b.id)
                     ->  Seq Scan on big_dist2 c  (cost=0.00..3701.33 rows=333333 width=4)
                     ->  Hash  (cost=4534.67..4534.67 rows=111111 width=4)
                           ->  Seq Scan on big_dist b  (cost=0.00..4534.67 rows=111111 width=4)
                                 Filter: ((SubPlan 1) > 0)
                                 SubPlan 1
                                   ->  Aggregate  (cost=10000003475.63..10000003475.64 rows=1 width=8)
                                         ->  Nested Loop  (cost=10000000000.00..10000003280.88 rows=77900 width=0)
                                               ->  Seq Scan on gp_segment_configuration s  (cost=0.00..1.05 rows=1 width=2)
                                                     Filter: (dbid = b.id)
                                               ->  Materialize  (cost=0.00..2890.33 rows=77900 width=4)
                                                     ->  Result  (cost=0.00..2500.83 rows=77900 width=4)
                                                           Filter: (t2.b = b.id)
                                                           ->  Materialize  (cost=0.00..1721.83 rows=77900 width=4)
                                                                 ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..1332.33 rows=77900 width=4)
                                                                       ->  Seq Scan on gp_tmp t2  (cost=0.00..293.67 rows=25967 width=4)
(21 rows)
```

There is a noticeable performance regression. This is the cost of ensuring correctness, and the trigger condition is quite narrow (it must be an 'Entry-related correlated subquery + distributed tables inside the subquery'). However, once hit, the regression becomes more pronounced as the scale of the parent query grows.

To narrow down the impact scope, the direction would be to force only the specific base table that actually generates the correlated parameters to the Entry node, instead of all base tables across that entire layer. That said, it would require passing the parameter-source relid into `set_rel_pathlist` for per-rel evaluation, which is a much more substantial change than the current approach.

Bonartze:
> I found some issues when attempting to perform performance regression testing:
> 
> ```sql
> create table big_dist(id int, v int) distributed by (id);
> insert into big_dist select i, i from generate_series(1, 1000000) i;
> create table big_dist2(id int, v int) distributed by (id);
> insert into big_dist2 select i, i from generate_series(1, 1000000) i;
> 
> create table gp_tmp(a int, b int, c int) distributed by (a);
> insert into gp_tmp select i, i, i from generate_series(1, 10) i;
> ```
> 
> The below SQL will ERROR out:
> 
> ```sql
> explain (costs on)
> select count(*) from big_dist b
> where (select count(*)
>        from gp_segment_configuration s
>        join gp_tmp t2 on s.dbid = t2.b
>        where s.dbid = b.id) > 0;
> ERROR:  variable not found in subplan target list (setrefs.c:3363)
> 
> -- which should be
>                                                                   QUERY PLAN                                                                  
> ----------------------------------------------------------------------------------------------------------------------------------------------
>  Finalize Aggregate  (cost=4812.50..4812.51 rows=1 width=8)
>    ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=4812.44..4812.49 rows=3 width=8)
>          ->  Partial Aggregate  (cost=4812.44..4812.45 rows=1 width=8)
>                ->  Seq Scan on big_dist b  (cost=0.00..4534.67 rows=111111 width=0)
>                      Filter: ((SubPlan 1) > 0)
>                      SubPlan 1
>                        ->  Aggregate  (cost=10000003475.63..10000003475.64 rows=1 width=8)
>                              ->  Nested Loop  (cost=10000000000.00..10000003280.88 rows=77900 width=0)
>                                    ->  Seq Scan on gp_segment_configuration s  (cost=0.00..1.05 rows=1 width=2)
>                                          Filter: (dbid = b.id)
>                                    ->  Materialize  (cost=0.00..2890.33 rows=77900 width=4)
>                                          ->  Result  (cost=0.00..2500.83 rows=77900 width=4)
>                                                Filter: (t2.b = b.id)
>                                                ->  Materialize  (cost=0.00..1721.83 rows=77900 width=4)
>                                                      ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..1332.33 rows=77900 width=4)
>                                                            ->  Seq Scan on gp_tmp t2  (cost=0.00..293.67 rows=25967 width=4)
>  Optimizer: Postgres-based planner
> (17 rows)
> ```
> 
> `force_entry` is a coarse hammer:
> 
> ```sql
> -- force_entry takes effect at the query-layer level, forcing all base tables in that layer, even if they are unrelated to the correlated parameters.
> explain (costs on)
> select count(*)
> from big_dist b
> join big_dist2 c on b.id = c.id
> where (select count(*)
>        from gp_segment_configuration s
>        join gp_tmp t2 on s.dbid = t2.b
>        where s.dbid = b.id) > 0;
> 
> 
>                                                                 QUERY PLAN                                                                 
> -------------------------------------------------------------------------------------------------------------------------------------------
>  Aggregate  (cost=48097.11..48097.12 rows=1 width=8)
>    ->  Hash Join  (cost=18145.78..47263.78 rows=333333 width=0)
>          Hash Cond: (c.id = b.id)
>          ->  Materialize  (cost=0.00..22034.67 rows=1000000 width=4)
>                ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..17034.67 rows=1000000 width=4)
>                      ->  Seq Scan on big_dist2 c  (cost=0.00..3701.33 rows=333333 width=4)
>          ->  Hash  (cost=13979.11..13979.11 rows=333333 width=4)
>                ->  Result  (cost=0.00..13979.11 rows=333333 width=4)
>                      Filter: ((SubPlan 1) > 0)
>                      ->  Materialize  (cost=0.00..10645.78 rows=333333 width=4)
>                            ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..8979.11 rows=333333 width=4)
>                                  ->  Seq Scan on big_dist b  (cost=0.00..4534.67 rows=111111 width=4)
>                      SubPlan 1
>                        ->  Aggregate  (cost=10000003475.63..10000003475.64 rows=1 width=8)
>                              ->  Nested Loop  (cost=10000000000.00..10000003280.88 rows=77900 width=0)
>                                    ->  Seq Scan on gp_segment_configuration s  (cost=0.00..1.05 rows=1 width=2)
>                                          Filter: (dbid = b.id)
>                                    ->  Materialize  (cost=0.00..2890.33 rows=77900 width=4)
>                                          ->  Result  (cost=0.00..2500.83 rows=77900 width=4)
>                                                Filter: (t2.b = b.id)
>                                                ->  Materialize  (cost=0.00..1721.83 rows=77900 width=4)
>                                                      ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1332.33 rows=77900 width=4)
>                                                            ->  Seq Scan on gp_tmp t2  (cost=0.00..293.67 rows=25967 width=4)
>  Optimizer: Postgres-based planner
> (24 rows)
> 
> 
> 
> -- Before this patch
>                                                                         QUERY PLAN                                                                        
> ----------------------------------------------------------------------------------------------------------------------------------------------------------
>  Finalize Aggregate  (cost=12263.84..12263.85 rows=1 width=8)
>    ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=12263.78..12263.83 rows=3 width=8)
>          ->  Partial Aggregate  (cost=12263.78..12263.79 rows=1 width=8)
>                ->  Hash Join  (cost=5923.56..11986.00 rows=111111 width=0)
>                      Hash Cond: (c.id = b.id)
>                      ->  Seq Scan on big_dist2 c  (cost=0.00..3701.33 rows=333333 width=4)
>                      ->  Hash  (cost=4534.67..4534.67 rows=111111 width=4)
>                            ->  Seq Scan on big_dist b  (cost=0.00..4534.67 rows=111111 width=4)
>                                  Filter: ((SubPlan 1) > 0)
>                                  SubPlan 1
>                                    ->  Aggregate  (cost=10000003475.63..10000003475.64 rows=1 width=8)
>                                          ->  Nested Loop  (cost=10000000000.00..10000003280.88 rows=77900 width=0)
>                                                ->  Seq Scan on gp_segment_configuration s  (cost=0.00..1.05 rows=1 width=2)
>                                                      Filter: (dbid = b.id)
>                                                ->  Materialize  (cost=0.00..2890.33 rows=77900 width=4)
>                                                      ->  Result  (cost=0.00..2500.83 rows=77900 width=4)
>                                                            Filter: (t2.b = b.id)
>                                                            ->  Materialize  (cost=0.00..1721.83 rows=77900 width=4)
>                                                                  ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..1332.33 rows=77900 width=4)
>                                                                        ->  Seq Scan on gp_tmp t2  (cost=0.00..293.67 rows=25967 width=4)
>  Optimizer: Postgres-based planner
> (21 rows)
> ```
> 
> There is a noticeable performance regression. This is the cost of ensuring correctness, and the trigger condition is quite narrow (it must be an 'Entry-related correlated subquery + distributed tables inside the subquery'). However, once hit, the regression becomes more pronounced as the scale of the parent query grows.
> 
> To narrow down the impact scope, the direction would be to force only the specific base table that actually generates the correlated parameters to the Entry node, instead of all base tables across that entire layer. That said, it would require passing the parameter-source relid into `set_rel_pathlist` for per-rel evaluation, which is a much more substantial change than the current approach.

Thanks for the detailed feedback and for pushing toward better approach.

I have implemented the per-relation narrowing via `force_entry_rels` (tracking specific RT indexes using a Bitmapset derived from `parent->plan_params`) and verified the execution plans on a live 3-segment cluster with 1M-row tables. 

The results are optimal and eliminate the performance regression. Here is the structural breakdown of how the fine-grained patch resolves your concerns:

### 1. The Join Case Caveat
In the earlier coarse implementation, the whole query layer was forced to the coordinator, causing a massive regression (`cost=48097.11`). With the new `force_entry_rels` pipeline, the planner finds an elegant distributed path instead of dragging `big_dist2` to the coordinator:

* **Distributed Scan:** `big_dist2` completely escapes the coordinator forcing. Its sequential scan runs fully distributed across the segments:  
  `-> Seq Scan on big_dist2 c (cost=0.00..3701.33 rows=333333 width=4)`
* **Coordinator Filtering:** Only the parameter-source relation (`big_dist b`) is gathered to the coordinator via `Gather Motion 3:1` to safely evaluate `SubPlan 1` where the catalog parameter is available.
* **Slicing back to Segments:** Instead of executing the join on the coordinator, the planner packages the small, filtered result set from the coordinator and streams it back to the segments using a **`Redistribute Motion 1:3`**.
* **Parallel Join:** The heavy `Hash Join` between the two 1M-row relations is executed **parallelly on the segments**, reducing the total query cost from **48,097** down to **26,152** (nearly a 2x performance optimization).

---

### 2. Catalog Visibility on Segments
To settle your fundamental question, I verified the behavior of the catalog on the segment slices using `gp_dist_random`:

```sql
SELECT gp_segment_id, count(*) 
FROM gp_dist_random('gp_segment_configuration') 
GROUP BY gp_segment_id;
-- Result: 0 rows

File: src/test/regress/sql/subselect_gp.sql [hunk 1]
```diff
@@ -1684,3 +1684,65 @@ select count(*) from foo where a > (select regr_count(x, y) from bar where foo.a
...
+        where s.dbid = o.x) > 0
+order by o.x;
+-- A correlated scan of the Entry catalog with no join (no Motion in the
+-- subplan) must keep working unchanged.
+select (select count(*)
+        from gp_segment_configuration s
+        where s.hostname = gp_segment.b) from gp_segment;
+drop table gp_entry_outer;
+drop table gp_segment;
+drop table gp_tmp;
```
SKeyerror, line 1773:
```sql
explain (costs on)
select count(*) from big_dist b
where (select count(*)
       from gp_segment_configuration s
       join gp_tmp t2 on s.dbid = t2.b
       where s.dbid = b.id) > 0;
```

This case should be added in the test file to prevent regression.

Bonartze: Thanks for the suggestion. I've added the regression coverage for both the `WHERE` and `HAVING` clauses.

I kept `costs off` because cost outputs are highly variable and change drastically depending on the local cluster setup, environment or execution parameters. Using `costs off` safely locks the plan shape without causing diff failures for other developers running tests on different machines.

SKeyerror (approved):
LGTM. Thanks for your work!

Before merge, please run a full CI pipeline using this branch at [warehouse-pg-test](https://github.com/EnterpriseDB/warehouse-pg-packaging/actions/workflows/warehouse-pg-test.yaml), make sure every test has passed.

PR #187, opened 2026-07-07
fix(aoco): probe the placeholder column during unique checks

adam8157:
Unique indexes on AOCO tables could miss conflicts and let duplicate keys through, observed in the field as duplicated primary keys after gpfdist loads.

AO unique checks work by probing the block directory for an entry covering the conflicting tid. Within an insert command the real block directory rows are not persisted yet, so the check relies on the placeholder row written at the start of the insert. The writer puts that placeholder on the first non-dropped column, but since 3986f926164 the reader consults the "anchor" column, i.e. the complete column with the smallest eof. Once a bulk load skews the segfile eofs so that the smallest column is no longer the first one (think of a well-compressed constant column), the reader probes a column the placeholder was never written to, finds nothing, and concludes there is no conflict. Both same-command and concurrent-transaction duplicates slip through. AOROW is unaffected: it has a single column group, so writer and reader cannot diverge.

Fix by introducing aocs_unique_check_col(), which picks the first non-dropped complete column (falling back to a dropped complete one), and using it on both sides: the placeholder write and the unique-check lookup. The choice is deterministic and independent of segfile eofs. Completeness matters because a column added in missing mode has no block directory coverage for rows predating the ADD COLUMN. Index-only scans keep using the anchor column, which is fine as they only look at committed rows with real block directory entries on every column.

The existing isolation2 tests only used single-column tables, where the anchor is always column 0 and the divergence cannot happen. Add cases with multi-column tables covering same-command, cross-command and concurrent duplicates, plus dropped-first-column and missing-mode ADD COLUMN variants, and a unit test for the column choice.

### DIFF
--- a/src/backend/access/aocs/aocsam.c
+++ b/src/backend/access/aocs/aocsam.c
@@ -3190,6 +3190,58 @@ get_anchor_col(AOCSFileSegInfo **segInfos, int nseg, int natts,
 	return scancol_proj != -1 ? scancol_proj : (scancol_all == -1 ? 0 : scancol_all);
 }
 
+/*
+ * Choose the column whose block directory entries are consulted during unique
+ * constraint checks, and for which insert commands persist their placeholder
+ * block directory row (see AppendOnlyBlockDirectory_InsertPlaceholder()).
+ *
+ * Writers and readers must agree on this column: a reader looking up a
+ * different column than the one covered by a concurrent (or same-command)
+ * writer's placeholder row would miss in-progress conflicts. The choice
+ * therefore has to be deterministic and independent of mutable state such as
+ * segfile eofs (which is what rules out get_anchor_col() here).
+ *
+ * The column also has to be "complete" (see get_anchor_col()): an
+ * incomplete column's block directory does not cover rows that predate its
+ * missing-mode ADD COLUMN, so consulting it would miss conflicts with those
+ * rows. We prefer the first non-dropped complete column, but fall back to a
+ * dropped one - block directory entries are written even for dropped columns.
+ */
+int
+aocs_unique_check_col(Relation aocsrel)
+{
+	int			natts = RelationGetNumberOfAttributes(aocsrel);
+	bool	   *column_not_complete;
+	int			checkcol = -1;
+
+	column_not_complete = ExistValidLastrownums(RelationGetRelid(aocsrel), natts);
+
+	for (int i = 0; i < natts; i++)
+	{
+		if (column_not_complete[i])
+			continue;
+
+		if (!TupleDescAttr(RelationGetDescr(aocsrel), i)->attisdropped)
+		{
+			checkcol = i;
+			break;
+		}
+
+		if (checkcol == -1)
+			checkcol = i;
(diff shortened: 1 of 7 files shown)


SKeyerror (approved):
LGTM.

```sql
DROP TABLE IF EXISTS aoco_unique_test;

CREATE TABLE aoco_unique_test (
    id      bigint,
    payload text,
    flag    char(1)
) USING ao_column
WITH (compresstype=zstd, compresslevel=7) DISTRIBUTED BY (id);

INSERT INTO aoco_unique_test
SELECT g, md5(g::text), 'N'
FROM generate_series(1, 200000) g;

SELECT column_num, sum(eof) AS total_eof
FROM gp_toolkit.__gp_aocsseg('aoco_unique_test'::regclass)
GROUP BY column_num
ORDER BY total_eof;

-- Add primary key using alter
ALTER TABLE aoco_unique_test ADD PRIMARY KEY (id);

-- We can insert successfully, which is not correct
INSERT INTO aoco_unique_test
SELECT v.x, md5(v.x::text) || '-' || random()::text, 'N'
FROM (VALUES (987654321::bigint), (987654321::bigint)) AS v(x);

-- And we can select 2 rows of 987654321
SELECT id, count(*) FROM aoco_unique_test WHERE id = 987654321 GROUP BY id;
SELECT * FROM aoco_unique_test where id = 987654321;

-- ====================================================================
-- TEST B (control): duplicate PK across TWO separate statements.
-- After statement 1 commits its command, real block-directory entries
-- exist for the anchor column, so statement 2's check should find the
-- conflict regardless of the anchor mismatch -> expected REJECTED.
-- This contrast confirms the defect is specific to the same-command
-- (placeholder-only) window.
-- ====================================================================
INSERT INTO aoco_unique_test VALUES (876543210, md5('a'), 'N');
INSERT INTO aoco_unique_test VALUES (876543210, md5('b'), 'N');

SELECT id, count(*) FROM aoco_unique_test WHERE id = 876543210 GROUP BY id;
```

In this reproduction, the value of `flag char(1)` is entirely 'N' and zstd-7 compression is applied, yielding the smallest EOF after the bulk load. Consequently:

* **Write side** (legacy code in `aocsam_handler.c`): The placeholder block directory entry is written to the first non-dropped column (column 0, i.e., `id`).
* **Read side** (legacy code in `Init_forUniqueChecks` ->  `init_internal_proj(..., use_anchor_column=true)` -> `get_anchor_col`): It queries based on the min-eof anchor column (column 2, i.e., `flag`).

Since the actual block directory entries within the same command have not yet been flushed to disk, the read side fails to find the placeholder in the `flag` column, falsely concluding that there is no conflict. This leads to both '987654321' rows being successfully inserted. This explains exactly why TEST A failed, while TEST B (cross-statement, where committed rows have actual blkdir entries across all columns) passed.

The current fix ensures that both sides call the same deterministic function `aocs_unique_check_col()` (which utilizes the first non-dropped complete column). Being independent of EOF, this divergence between the read and write sides is eliminated, thereby resolving the placeholder issue.

PR #188, opened 2026-07-07
Fix "out of scanning scope" error on AO/AOCS index fetch concurrent with VACUUM

avamingli:
## Problem

A query doing an index fetch on an AO/AOCS table can fail with

```
ERROR:  Row No. 1675 in segment file No. 2 is out of scanning scope for target relfilenode 24898. (aocsam.c:2267)
```

when a `VACUUM` of the same table runs concurrently. Typical trigger: a
long-running `UPDATE ... FROM <aoco_table>` whose inner side is a bitmap index
scan, overlapping a scheduled `VACUUM` of that table. Reported from a
production system; reproduced on a stock demo cluster in under 2 seconds with
the stress script described below.

## Root cause

Commit 61323b178e ("Initialize lastSequence only for the target scanning
segments during index scan.") narrowed the `lastSequence[]` initialization in
the AO/AOCS fetch descriptor from all 128 possible segnos to only the segfiles
visible in pg_aoseg/pg_aocsseg at initialization time, and turned a hit on any
other segno into the XX000 error above.

The fetch descriptor (and its cached segfile list) is initialized at the
*first fetch* of the scan and lives for the whole scan. While the scan is
running, a concurrent VACUUM compaction (or any concurrent insert) may pick a
**brand-new segfile** as its insertion target. The btree immediately contains
physically visible entries for the moved tuples, so a long-running index fetch
(e.g. a nestloop rescanning its inner bitmap index scan) can return a TID
pointing into a segno that did not exist when the descriptor was initialized →
`lastSequence[segno] == InvalidAORowNum` → spurious XX000.

None of those tuples can be visible to the fetch's snapshot: a segno missing
from the cached list means its first-ever use began after the descriptor was
initialized (initial pg_aoseg/pg_aocsseg entries are frozen on insert, so this
is the only way to miss). Erroring out is wrong; the TID should simply be
treated as not found.

## Fix

In `aocs_fetch()` and `appendonly_fetch()`, fold the
`lastSequence[segno] == InvalidAORowNum` case into the existing
"rowNum beyond lastSequence → return not found" fast path instead of raising
the error. This restores the pre-61323b178e semantics (`ReadLastSequence()`
returned 0 for a segno without a gp_fastsequence entry, silently skipping such
TIDs) while keeping that commit's optimization intact.

## Test

New isolation2 test `uao/index_fetch_while_vacuum` (row + column variants)
reproduces the race deterministically without fault injection:

1. session 1 opens a cursor over a nestloop join with a bitmap index scan
   inner side and pulls one row (`MOVE 1`) — pinning the fetch descriptor and
   its cached segfile list;
2. session 2 deletes ~1/3 of the AO table and runs `VACUUM` — compaction moves
   the surviving tuples into a brand-new segfile (the reader's open snapshot
   keeps the old one in `AWAITING_DROP`);
3. session 1 drains the cursor (`MOVE ALL`) — the rescanning inner index scans
   now return TIDs into the new segfile.

Without the fix both variants fail deterministically (row: appendonlyam.c,
column: aocsam.c); with the fix the cursor returns the exact pre-DELETE row
count.

## Verification matrix

| Check | Before fix | After fix |
|---|---|---|
| Stress repro (3 concurrent sessions: index-fetch UPDATE loop + DELETE/INSERT/VACUUM churn + snapshot holder) | error in **1 s**, 1st reader iteration | clean for **300 s** / ~203 VACUUM cycles, insert segno advanced to 29 |
| Deterministic cursor repro (ao_column) | XX000 every run | no error; row count exact (snapshot-correct, deleted rows still visible to old snapshot) |
| New isolation2 test, ao_row + ao_column | both FAIL (appendonlyam.c / aocsam.c) | both pass |
| `uao/limit_indexscan_inits_row/_column` (test added by 61323b178e — guards its optimization) | — | pass |
| Related suites: `uao/` cursor/compaction/vacuum tests, `uao_crash_compaction_*`, brin | — | pass |
| **Full isolation2 schedule** (demo cluster, 3 primaries + 3 mirrors) | — | **All 303 tests passed** |

### DIFF
--- a/src/backend/access/aocs/aocsam.c
+++ b/src/backend/access/aocs/aocsam.c
@@ -2261,18 +2261,19 @@ aocs_fetch(AOCSFetchDesc aocsFetchDesc,
 
 	Assert(segmentFileNum >= 0);
 
-	if (aocsFetchDesc->lastSequence[segmentFileNum] == InvalidAORowNum)
-		ereport(ERROR,
-				(errcode(ERRCODE_INTERNAL_ERROR),
-				 errmsg("Row No. %ld in segment file No. %d is out of scanning scope for target relfilenode %u.",
-				 		rowNum, segmentFileNum, aocsFetchDesc->relation->rd_node.relNode)));
-
 	/*
 	 * skip if the rowNum is not discoverable:
-	 * 1. rowNum is 0 which is invalid.
-	 * 2. rowNum bigger than lastsequence of the segment.
+	 * 1. the segment file was first used after this fetch descriptor was
+	 *    initialized (lastSequence is InvalidAORowNum): e.g. a concurrent
+	 *    VACUUM compaction picked a brand-new segfile as its insertion
+	 *    target and the index already contains entries for the moved
+	 *    tuples.  None of those tuples can be visible to our snapshot,
+	 *    so treat this as "not found" just like the cases below.
+	 * 2. rowNum is 0 which is invalid.
+	 * 3. rowNum bigger than lastsequence of the segment.
 	 */
-	if (rowNum == 0 || rowNum > aocsFetchDesc->lastSequence[segmentFileNum])
+	if (aocsFetchDesc->lastSequence[segmentFileNum] == InvalidAORowNum ||
+		rowNum == 0 || rowNum > aocsFetchDesc->lastSequence[segmentFileNum])
 	{
 		if (slot != NULL)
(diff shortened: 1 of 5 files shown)


haolinw:
Good catch! LGTM. Thanks for fixing this.

haolinw (approved):

PR #189, opened 2026-07-08
fix(orca): avoid duplicated rows for percentiles over a grouped subquery

linwen:
ORCA returns duplicated rows when splittable ordered-set aggregates,
such as percentile_cont and percentile_disc, are the only aggregates in
a scalar aggregation and their input is itself a grouped expression.

This is a global ordered-set aggregate over a grouped input, not the
unsupported per-group ordered-set aggregate case: the outer aggregate
has no grouping columns, so it satisfies the existing gp_percentile
split guard. Only the input subquery is grouped.

For example, this query should return a single row, as the
postgres legacy planner does, but ORCA returned one copy of the row per
group of the subquery:

    CREATE TABLE t_metrics (k1 int, m1 numeric, m2 numeric, m3 numeric)
    DISTRIBUTED BY (k1);

    INSERT INTO t_metrics VALUES
        (1, 0.10, 0.20, 0.30), (1, 0.12, 0.18, 0.25), (1, 0.15, 0.22, 0.28),
        (2, 0.50, 0.60, 0.70), (2, 0.55, 0.58, 0.65), (3, 0.90, 0.10, 0.05);

    SELECT percentile_disc(0.5) WITHIN GROUP (ORDER BY m1) AS p50_m1,
           percentile_disc(0.5) WITHIN GROUP (ORDER BY m2) AS p50_m2,
           percentile_disc(0.5) WITHIN GROUP (ORDER BY m3) AS p50_m3
    FROM (SELECT k1, m1, m2, m3 FROM t_metrics GROUP BY 1,2,3,4) s;

Reason: when the project list contains no regular aggregates besides
the ordered-set aggregates, SplitOrderedAggsPrj() handed back the
original input child as the "remaining aggregates" expression. If that
child is itself a GbAgg, as with a GROUP BY or DISTINCT subquery,
PexprInputAggPrj2Join() mistakes it for the GbAgg that computes
remaining regular aggregates and cross-joins it with the one-row
gp_percentile result. That duplicates the output once per input group.

Fix by returning NULL for the remaining-aggregates expression when
there is none, and by checking for NULL in PexprInputAggPrj2Join().
This preserves the existing path for cases with regular aggregates,
where a real remaining-aggregates GbAgg is still constructed and
joined.

The bug dates back to the introduction of the ordered-set aggregate
split in b90f952808.

Add a minidump test and SQL regress cases covering single and multiple
percentiles over GROUP BY and DISTINCT subqueries, with and without a
regular aggregate alongside.


24nishant (approved):
Changes look good to me.

PR #190, opened 2026-07-08
test: fix atomic_test expectations and make >4GiB workfile unit tests sparse

adam8157:
### 1. `atomic_test` never asserted anything

Both `pg_atomic_sub_fetch_u64` sub-tests computed their expected value with `+` instead of `-`, so two of the five assertions failed on every platform, the summary was always `f` — and the `f` rows were snapshotted into the expected output in the same commit that added the test (2020, 5b147fbe05e). Fix the arithmetic and expect `t`.

### 2. The >4GiB unit tests wrote ~19GB of filler to test offset arithmetic

`buffile_large_file_test` and `logicaltape_test` each wrote 4.83GB per backend (48319 x 100KB records), on the coordinator and every segment at once — ~19GB transient disk on a one-disk demo cluster, enough to ENOSPC a host with less headroom, and minutes of pure filler I/O.

What they guard is int64 offset arithmetic and 1GiB segment rollover, not bulk data volume. `BufFileSeek` cannot move past the last existing segment file, but within existing segments it does not check the physical file size, and pwrite leaves never-written ranges as filesystem holes. So the buffile test now grows the file by writing one record across each 1GiB boundary (exercising the real segment-rollover write path four times), then writes a marked record just past the 4GiB mark and reads it back through both `BufFileSeek` and `BufFileSeekBlock`: logically >4GiB, under 1MB on disk. The tape (strictly sequential block allocation, no sparse path) shrinks to just past the underlying 1GiB segment boundary; its block I/O goes through `BufFileSeekBlock`, which the buffile test now covers past 4GiB.

The target records are also prefixed with a distinct marker — previously every record had identical content, so the post-seek read-back could not have detected a seek landing on the wrong record.

### Verification

On a demo cluster, `buffile_large_file_test`, `logicaltape_test` and `atomic_test` all return `t` on the coordinator and all segments (previously `f` for atomic_test, ENOSPC-prone for the large-file pair).

### DIFF
--- a/src/test/isolation2/workfile_mgr_test.c
+++ b/src/test/isolation2/workfile_mgr_test.c
@@ -118,6 +118,8 @@ gp_workfile_mgr_test_harness(PG_FUNCTION_ARGS)
 		elog(LOG, "No tests match given name: %s", test_name);
 	}
 
+	pfree(test_name);
+
 	PG_RETURN_BOOL(ran_any_tests && result);
 }
 
@@ -153,6 +155,8 @@ gp_workfile_mgr_create_workset(PG_FUNCTION_ARGS)
 		if (closeFile)
 			BufFileClose(buffile);
 	}
+
+	pfree(worksetName);
 
 	PG_RETURN_VOID();
 }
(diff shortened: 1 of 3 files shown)


src/test/isolation2/workfile_mgr_test.c
haolinw, line 535: `buffer` missed pfree.

src/test/isolation2/workfile_mgr_test.c
```diff
@@ -554,28 +594,21 @@ logicaltape_test(void)
...
 		}
 
 	}
 
+	pfree(filler->data);
+	pfree(filler);
+
 	/* Set target LogicalTape */
 	work_tape = test_tape;
 	char *buffer = palloc(nchars * sizeof(char));
```
haolinw, line 621: Both `buffer` and `test_string` missed pfree.

haolinw (approved):
LGTM

PR #194, opened 2026-07-13
Fix ORCA folding NULL scalar-subquery count() results to 0

adam8157:
A scalar subquery returns NULL when a row-eliminating operator (a correlated filter above the aggregate, or a join) eliminates the aggregate's output row; it returns 0 only when the count() aggregate itself ran over empty input.

CSubqueryHandler wrapped the subquery output in COALESCE(count, 0) whenever CUtils::FHasCountAggMatchingColumn() could find a matching count() aggregate anywhere below the subquery root -- the recursion walked through filters, limits and joins, so an eliminated aggregate row was resurrected as 0:

    select v.c, (select cnt
                 from (select count(*) cnt from generate_series(1,4)) s
                 where v.c)
    from (values (false),(true)) v(c);

returned (f,0) instead of (f,NULL) with optimizer=on.

Fix by restricting the coalesce decision to count aggregates reachable through row-preserving operators only, and by requiring correlated execution when the count output sits behind a row-eliminating operator: a single COALESCE at the subquery output cannot represent both "aggregate row eliminated" (NULL) and "aggregate over empty input" (0) in the decorrelated form. Regenerate the two affected minidumps and add regression coverage to bfv_subquery.

### DIFF
--- a/src/backend/gporca/data/dxl/minidump/ScalarSubqueryCountStar.mdp
+++ b/src/backend/gporca/data/dxl/minidump/ScalarSubqueryCountStar.mdp
@@ -277,10 +277,10 @@ EXPLAIN SELECT (SELECT COUNT(*) FROM bar GROUP BY c LIMIT 1) FROM foo;
         </dxl:LogicalGet>
       </dxl:LogicalProject>
     </dxl:Query>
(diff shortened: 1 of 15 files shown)


JunfengYang (approved):
+1

PR #200, opened 2026-07-27
Veto direct dispatch for Strewn-locus Append/MergeAppend children

### LINKED ISSUE #199: Direct dispatch silently drops rows from a Strewn-locus Append/MergeAppend child (gp_dist_random over a FROM-less view)
## Summary

`DirectDispatchUpdateContentIdsFromPlan()`'s per-plan-node-type switch treats `T_Result` and `T_SubqueryScan` as "no change" (silence, meaning "no constraint"). That's only sound for pass-through nodes. A `gp_dist_random(...)` argument gets `CdbPathLocus_MakeStrewn()` applied to its path specifically because there is no known function tying its rows to segments — it must run on every segment, with no qual-based narrowing possible. But once such a child's plan is built, its shape can collapse to a bare `Result` or a plain `SubqueryScan` (e.g. a FROM-less view's contents get constant-folded away), and the walker has no way to tell that collapsed shape apart from a harmless one — both silently contribute "no change". That silence lets a sibling's correctly-computed narrow target (e.g. a plain relation scan with an equality qual on the distribution key) win the whole Append's (or MergeAppend's) direct-dispatch decision outright, silently dropping the Strewn child's rows on every segment the sibling's target excludes.

## Repro (Append)

```sql
set optimizer = off;
create table t(a int) distributed by (a);
insert into t select generate_series(1,100);
create view v_fromless as select 1 as x;

select a from t where a = 1
union all
select x from gp_dist_random('v_fromless');
```

Expected 2 rows (1 from `t` + 1 from the view, once per each of the 3 segments = 3 rows, so actually 4 total); actual 2 — the view's rows on 2 of the 3 segments are silently dropped, because the whole slice gets `Gather Motion 1:1 (segments: 1)` instead of running on all 3 segments.

## Repro (MergeAppend)

When both UNION ALL branches are independently sortable (an inner `ORDER BY` on one side; a single-row `Result` is trivially sorted on the other), the planner picks a `MergeAppend` instead of a plain `Append`, and the identical collapse happens there too:

```sql
set optimizer = off;
create table foo(a int, b int) distributed by (a);
insert into foo select 100, i from generate_series(1,1000) i;
create view foo_fromless as select 1 as x;

(select b from foo where a = 100 order by b)
union all
select x from gp_dist_random('foo_fromless')
order by 1 limit 5;
```

Expected `1,1,1,1,2`; actual `1,1,2,3,4` — two of the view's three per-segment rows are dropped.

## Root cause

Silence in the walker's switch is only safe for genuine pass-through nodes, not for a Strewn-locus leaf whose plan shape happened to collapse to something the walker can't distinguish from a harmless pass-through.

## Fix

Check each Append/MergeAppend child's own Path locus directly, in `create_append_plan()` and `create_merge_append_plan()`, before the child plan is even built: a Strewn locus is an unambiguous, Path-level signal that this child cannot be narrowed under any circumstances, so veto direct dispatch for the whole slice right there.

Verified on a 3-segment demo cluster (`optimizer=off`): both repros now return the correct row counts; a plain two-relation point-lookup UNION ALL with no Strewn child still narrows to a single segment, unaffected.

avamingli:
## Summary

Closes #199

`DirectDispatchUpdateContentIdsFromPlan()`'s per-plan-node-type switch treats `T_Result` and `T_SubqueryScan` as "no change" (silence, meaning "no constraint"). That's only sound for pass-through nodes. A `gp_dist_random(...)` argument gets `CdbPathLocus_MakeStrewn()` applied to its path specifically because there is no known function tying its rows to segments — it must run on every segment, with no qual-based narrowing possible. But once such a child's plan is built, its shape can collapse to a bare `Result` or a plain `SubqueryScan` (e.g. a FROM-less view's contents get constant-folded away), and the walker has no way to tell that collapsed shape apart from a harmless one — both silently contribute "no change". That silence lets a sibling's correctly-computed narrow target win the whole Append's (or MergeAppend's) direct-dispatch decision outright, silently dropping the Strewn child's rows on every segment the sibling's target excludes.

The same collapse happens through a MergeAppend when both UNION ALL branches are independently sortable (an inner ORDER BY on one side; a single-row Result is trivially sorted on the other), since `create_merge_append_plan()` had none of the veto either.

## Fix

Check each child's own Path locus directly, in `create_append_plan()` and `create_merge_append_plan()`, before the child plan is even built: a Strewn locus is an unambiguous, Path-level signal that this child cannot be narrowed under any circumstances, so veto direct dispatch for the whole slice right there.

## Testing

Added regression coverage to `direct_dispatch` (both variants), including the `_optimizer` answer file (GPORCA falls back to the Postgres planner for this construct, so the output is identical). Verified live on a 3-segment demo cluster: both repros return the correct row counts; a plain two-relation point-lookup UNION ALL with no Strewn child still narrows to a single segment, unaffected. Ran the full `greenplum_schedule` — the only failures observed were pre-existing environment issues (a fault-injector left armed by an earlier fault-injection test, cascading into unrelated later tests), unrelated to this change; `direct_dispatch` itself passes cleanly.

### DIFF
--- a/src/backend/optimizer/plan/createplan.c
+++ b/src/backend/optimizer/plan/createplan.c
@@ -1271,9 +1271,23 @@ create_append_plan(PlannerInfo *root, AppendPath *best_path, int flags)
 		 * this child's rows on every segment the sibling's target
 		 * excludes.  Veto direct dispatch here, from the Path's own
 		 * locus, before that ambiguity has a chance to arise downstream.
+		 *
+		 * Exception: set_append_path_locus() also tags a pinned General/
+		 * SegmentGeneral child (the "One-Time Filter:
+		 * gp_execution_segment() = N" trick, see the direct_dispath_
+		 * contentIds comment above) as Strewn -- that child is NOT an
+		 * unconstrained "runs everywhere" leaf, it is already confined to
+		 * one specific, known segment, and create_plan_recurse() (via
+		 * create_gating_plan()'s tail above) already merged that exact
+		 * content id into this slice's direct-dispatch info. Blanket-
+		 * vetoing here would clobber that correct, already-computed
+		 * narrowing, so skip the veto whenever the Path recorded a
+		 * specific pin.
 		 */
 		if (Gp_role == GP_ROLE_DISPATCH && root->config->gp_enable_direct_dispatch &&
-			CdbPathLocus_IsStrewn(subpath->locus))
+			CdbPathLocus_IsStrewn(subpath->locus) &&
+			!(IsA(subpath, ProjectionPath) &&
+			  ((ProjectionPath *) subpath)->direct_dispath_contentIds != NIL))
 		{
 			DirectDispatchInfo dispatchInfo;
 
@@ -1490,10 +1504,14 @@ create_merge_append_plan(PlannerInfo *root, MergeAppendPath *best_path,
 		 * MergeAppend arises here exactly like an Append does -- from a
 		 * UNION ALL whose branches happen to already be sorted (or
 		 * trivially sorted, as a single-row Result is) -- so it needs the
-		 * same veto.
+		 * same veto, with the same exception for an already-pinned
+		 * General/SegmentGeneral child (see the comment in
+		 * create_append_plan()).
 		 */
 		if (Gp_role == GP_ROLE_DISPATCH && root->config->gp_enable_direct_dispatch &&
-			CdbPathLocus_IsStrewn(subpath->locus))
+			CdbPathLocus_IsStrewn(subpath->locus) &&
+			!(IsA(subpath, ProjectionPath) &&
+			  ((ProjectionPath *) subpath)->direct_dispath_contentIds != NIL))
 		{
 			DirectDispatchInfo dispatchInfo;
 
--- a/src/backend/optimizer/plan/createplan.c
+++ b/src/backend/optimizer/plan/createplan.c
@@ -1255,6 +1255,36 @@ create_append_plan(PlannerInfo *root, AppendPath *best_path, int flags)
 		subplan = create_plan_recurse(root, subpath, CP_EXACT_TLIST);
 
 		/*
+		 * A Strewn-locus child (e.g. gp_dist_random(...), which
+		 * CdbPathLocus_MakeStrewn()'s callers use precisely because there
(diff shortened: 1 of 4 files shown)


avamingli:
Pushed a correction (6b4b6d2bc4). While chasing down the earlier CI failure I found the veto was too broad: `set_append_path_locus()` also tags an already-pinned General/SegmentGeneral child (the "One-Time Filter: gp_execution_segment() = N" trick for the earlier pinned-Append-child fix) as `Strewn`. That child is already confined to one known segment and its narrowing was already being merged correctly elsewhere -- the blanket veto clobbered that and widened dispatch to every segment whenever such a child was present.

Fixed by skipping the veto when the Strewn-locus child is a `ProjectionPath` with a non-NIL `direct_dispath_contentIds` (i.e. the Path itself already recorded a specific pin). Re-verified both original repros plus the partition-pruning construct that exposed the gap; ran the full `greenplum_schedule` again -- `direct_dispatch` and `partition_prune` both pass cleanly now.

src/backend/optimizer/plan/createplan.c
```diff
@@ -1254,6 +1254,50 @@ create_append_plan(PlannerInfo *root, AppendPath *best_path, int flags)
...
+		 * create_gating_plan()'s tail above) already merged that exact
+		 * content id into this slice's direct-dispatch info. Blanket-
+		 * vetoing here would clobber that correct, already-computed
+		 * narrowing, so skip the veto whenever the Path recorded a
+		 * specific pin.
+		 */
+		if (Gp_role == GP_ROLE_DISPATCH && root->config->gp_enable_direct_dispatch &&
+			CdbPathLocus_IsStrewn(subpath->locus) &&
+			!(IsA(subpath, ProjectionPath) &&
+			  ((ProjectionPath *) subpath)->direct_dispath_contentIds != NIL))
```
SKeyerror, line 1290: I found this is a typo: "direct_dispath_contentIds", not "direct_dispatch_contentIds" 🤣

This is a historical problem. Not blocking this PR.

SKeyerror (approved):
LGTM.

src/backend/optimizer/plan/createplan.c
avamingli: grep the typo, found one more:)

PR #201, opened 2026-07-28
Fix gp_workfile_entries inflating metrics via gang pid fan-out

SKeyerror:
Fixes https://github.com/warehouse-pg/warehouse-pg/issues/122 (internal ticket: PTT-746)

## Problem

`gp_toolkit.gp_workfile_entries` joins the workfile entries to `gp_stat_activity` on `(sess_id, segid)` only. Every backend of the session's gang on a segment matches that condition, so a single spilling workfile set is reported once per backend of the gang, inflating `SUM(size)` and `SUM(numfiles)` in the downstream views `gp_workfile_usage_per_query` and `gp_workfile_usage_per_segment`.

For example, while a 3-slice query is spilling, the view reports each workfile set 3+ times per segment with identical size/numfiles but different pids (see issue #122 for a full reproduction).

## Fix

Record the pid of the process that creates a `workfile_set` and join on it:

- `workfile_mgr.h`: store the creating process's pid in the reserved ABI field of `workfile_set` (split the 8-byte `abi_reserved` into `int32 pid` + `int32 abi_reserved`), so the struct layout is unchanged.
- `workfile_mgr.c`: set `work_set->pid = MyProcPid` on creation, and add `gp_workfile_mgr_cache_entries_v2_internal()` which returns the same columns as before plus a trailing `pid` column. Both variants share one implementation.
- `gp_workfile_mgr` module: export the new `gp_workfile_mgr_cache_entries_v2` symbol.
- `gp_toolkit` 1.9 (new `gp_toolkit--1.8--1.9.sql`, `default_version` bumped): add `__gp_workfile_entries_f_v2_on_coordinator/segments()` and recreate `gp_workfile_entries` to join with `C.pid = S.pid` as well. The `pid` column now comes from the workfile entry itself, so it identifies the actual spilling process and stays populated even when the left join finds no activity row.

The old 8-column C function is kept unchanged, so views from older extension versions (which spell out the 8-column record in their column definition lists) keep working with new binaries until `ALTER EXTENSION gp_toolkit UPDATE` is run.

## Before / after

While one hash-join query is spilling (session with 4 backends per segment, 2 workfile sets per segment):

Old view (1.8): 24 rows — each workfile set duplicated once per gang process:

```
   pid   | sess_id | slice | segid |  optype  |   size   | numfiles
---------+---------+-------+-------+----------+----------+----------
 1114150 |      10 |     1 |     0 | HashJoin | 14221312 |      127
 1114157 |      10 |     1 |     0 | HashJoin | 14221312 |      127
 1114162 |      10 |     1 |     0 | HashJoin | 14221312 |      127
 1114169 |      10 |     1 |     0 | HashJoin | 14221312 |      127
 ...
(24 rows)
```

New view (1.9): 6 rows — one row per workfile set, attributed to the pid of the spilling process:

```
   pid   | sess_id | slice | segid |  optype  |   size   | numfiles
---------+---------+-------+-------+----------+----------+----------
 1114157 |      10 |     1 |     0 | HashJoin | 20283392 |      254
 1114157 |      10 |     1 |     0 | HashJoin | 14221312 |      127
 ...
(6 rows)
```

`gp_workfile_usage_per_segment` correspondingly reports 381 files per segment instead of 4x that.

## Testing

- Added a regression test to `isolation2/workfile_mgr_test`: create extra gangs in a session with a multi-slice query (and assert more than one backend per segment exists, i.e. the fan-out precondition holds), create an inter-transaction workset on all segments, then assert the view reports exactly one entry per segment with a pid that matches a live backend of the session.
- Verified `ALTER EXTENSION gp_toolkit UPDATE TO '1.9'` on a cluster initialized at 1.8.
- `gpcontrib/gp_toolkit` regress suite passes (the `gp_toolkit_resqueue` failure is pre-existing and unrelated; it fails identically on a clean tree).

### DIFF
--- a/gpcontrib/gp_internal_tools/gp_workfile_mgr.c
+++ b/gpcontrib/gp_internal_tools/gp_workfile_mgr.c
@@ -30,6 +30,7 @@
 PG_MODULE_MAGIC;
 
 Datum gp_workfile_mgr_cache_entries(PG_FUNCTION_ARGS);
+Datum gp_workfile_mgr_cache_entries_v2(PG_FUNCTION_ARGS);
 Datum gp_workfile_mgr_used_diskspace(PG_FUNCTION_ARGS);
 
 PG_FUNCTION_INFO_V1(gp_workfile_mgr_cache_entries);
@@ -43,6 +44,18 @@ Datum
 gp_workfile_mgr_cache_entries(PG_FUNCTION_ARGS)
 {
 	return gp_workfile_mgr_cache_entries_internal(fcinfo);
+}
+
+PG_FUNCTION_INFO_V1(gp_workfile_mgr_cache_entries_v2);
+
+/*
+ * Same as gp_workfile_mgr_cache_entries, but the result also includes the
+ * pid of the process that created each workfile set.
+ */
+Datum
+gp_workfile_mgr_cache_entries_v2(PG_FUNCTION_ARGS)
+{
+	return gp_workfile_mgr_cache_entries_v2_internal(fcinfo);
 }
 
 PG_FUNCTION_INFO_V1(gp_workfile_mgr_used_diskspace);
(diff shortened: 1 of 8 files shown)


adam8157 (approved):

SKeyerror:
CI all passed, merge it.

PR #202, opened 2026-07-28
Fix typo: direct_dispath_contentIds -> direct_dispatch_contentIds

avamingli:
- [ ] Document changes

### DIFF
--- a/.abi-check/7.1.0/postgres.types.ignore
+++ b/.abi-check/7.1.0/postgres.types.ignore
@@ -9,3 +9,4 @@ struct AppendOnlyFetchDescData
 struct AppendOnlyInsertDescData
 struct ResGroupCaps
 struct PlannerConfig
+struct ProjectionPath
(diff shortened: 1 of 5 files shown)


avamingli:
The abi-compare job failure here is a false positive, not a real ABI break.

ProjectionPath.direct_dispath_contentIds got renamed to direct_dispatch_contentIds as part of the typo fix. abi-compliance-checker flags this as a High severity source compatibility problem, since pathnodes.h is one of the installed public headers and any extension source referencing the old field name by string would fail to recompile:

1 Field direct_dispath_contentIds has been renamed to direct_dispatch_contentIds.
Recompilation of a client program may be broken with the error message:
struct ProjectionPath has no member named direct_dispath_contentIds.

Binary compatibility is still 100% (offset/layout unchanged), so already-built extensions keep working. ProjectionPath is a planner-internal Path node though, not something extensions are meant to reach into, so this is the same situation we already carved out for PlannerConfig in .abi-check/7.1.0/postgres.types.ignore.

Suggest adding struct ProjectionPath to that ignore list to unblock this, same precedent as PlannerConfig?

@SKeyerror  what do you think?

SKeyerror:
I'm more in favor of modifying it directly. `direct_dispath_contentIds` doesn't really look like something that would be widely used by other extensions anyway. We can add it to `postgres.types.ignore`.

avamingli:
> I'm more in favor of modifying it directly. `direct_dispath_contentIds` doesn't really look like something that would be widely used by other extensions anyway. We can add it to `postgres.types.ignore`.

Sounds good, done

SKeyerror (approved):

PR #205, opened 2026-08-03
fix: dispatch ALTER COLLATION ... REFRESH VERSION to segments

JunfengYang:
## Problem

`ALTER COLLATION ... REFRESH VERSION` updates `pg_collation.collversion` on the coordinator only. The collation version mismatch check runs on whichever node actually uses the collation, so after a collation provider (ICU) upgrade:

1. every node starts warning `collation ... has version mismatch`;
2. the administrator reindexes and runs `ALTER COLLATION ... REFRESH VERSION`;
3. the warning disappears on the coordinator, **but the segments keep the stale version and keep warning forever** — there is no supported way to clear them cluster-wide.

Reproduction on an ICU build (`153.14` is the ICU 67.1 collator version):

```
-- plant collversion '0.0' on every node, then on the coordinator:
ALTER COLLATION c REFRESH VERSION;
NOTICE:  changing version from 0.0 to 153.14

-- probe pg_collation UNION ALL gp_dist_random('pg_collation'):
-1|153.14   0|0.0   1|0.0   2|0.0     <- catalogs silently diverged
```

## Fix

* `AlterCollation()` now dispatches the statement (same three-flag convention as `DefineCollation()` in the same file). Each node re-derives the version from its **own** collation provider, which is the correct semantics when provider versions differ across hosts.
* `AlterCollationStmt` had never been dispatched before, so it had no serialization support at all — dispatching died with `could not serialize unrecognized node type: 721 (outfast.c)`. Added the missing `_out`/`_read` functions and switch cases in outfuncs/outfast/readfuncs/readfast (the struct carries a single `List *collname`).

## Test

New `gp_collation_refresh` regress test (scheduled in `greenplum_schedule`):

* seeds a consistent stale version on **all** nodes through the already-dispatched `CREATE COLLATION ... (provider = icu, version = '0.0')` — no utility-mode catalog surgery needed;
* asserts cross-node consistency via boolean probes over `pg_collation UNION ALL gp_dist_random('pg_collation')` (no environment-specific ICU version strings in the expected output; node count taken from `gp_segment_configuration`, not hard-coded);
* skips cleanly on non-ICU builds / non-UTF8 databases via a `\if :skip_test \quit` guard plus an alternate expected file (`gp_collation_refresh_1.out`, captured from a real skip-path run).

## Verification

* Unpatched build: repro shows the divergence above; the new test **fails** with `consistent/refreshed = f|f` — it bites the actual bug.
* Patched build: all nodes converge (`153.14` everywhere, per-segment NOTICEs visible); test passes on both the ICU path and the skip path.
* `installcheck-small`: 194/194 pass.

### DIFF
--- a/src/backend/commands/collationcmds.c
+++ b/src/backend/commands/collationcmds.c
@@ -398,6 +398,20 @@ AlterCollation(AlterCollationStmt *stmt)
 	heap_freetuple(tup);
 	table_close(rel, NoLock);
 
+	/*
+	 * Refresh the version recorded in every segment's pg_collation too;
+	 * each node derives the actual version from its own collation
+	 * provider, and collation version mismatch checks run wherever the
+	 * collation is used.
+	 */
+	if (Gp_role == GP_ROLE_DISPATCH)
+		CdbDispatchUtilityStatement((Node *) stmt,
+									DF_CANCEL_ON_ERROR|
+									DF_WITH_SNAPSHOT|
+									DF_NEED_TWO_PHASE,
+									NIL,
+									NULL);
+
 	return address;
 }
 
--- a/src/backend/nodes/outfast.c
+++ b/src/backend/nodes/outfast.c
@@ -1376,6 +1376,10 @@ _outNode(StringInfo str, void *obj)
 				_outAlterObjectDependsStmt(str, obj);
 				break;
 
+			case T_AlterCollationStmt:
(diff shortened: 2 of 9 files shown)


haolinw (approved):
LGTM. This fixes a unrepairable cluster-wide catalog divergence, and helps gpcheckcat to detect a real ICU divergence environment.

PR #208, opened 2026-08-07
fix: treat fixed-size hash keys as binary, and enforce key semantics in hash_create()

JunfengYang:
Three commits: two bug fixes sharing one defect pattern, plus a backport of the upstream hardening that makes the pattern impossible to reintroduce silently.

## The defect pattern

`hash_create()` without `HASH_BLOBS`/`HASH_FUNCTION` silently defaults to **string** hashing and comparison. Callers keying on fixed-size integers then get `strlen()`/`strncmp()` over binary bytes: out-of-bounds reads when no key byte is zero, and truncated comparisons that make different keys alias whenever they agree up to the first zero byte.

## Commit 1 — autovacuum alive-session hash (`do_autovacuum`)

Session ids (4-byte ints) hashed as strings:

* Under AddressSanitizer every autovacuum worker aborts at startup (`stack-buffer-overflow` in `string_hash`), crash-looping the cluster.
* Deterministic aliasing: with a live session id `S < 65536`, an orphaned temp namespace with suffix `S + 2^24` has the identical C-string view (`3D 00 00 00` vs `3D 00 00 01` → both `"\x3D"`) and is never cleaned up.

Fix: `HASH_ELEM | HASH_BLOBS`. New `autovacuum-alive-session` regress test fabricates the aliasing orphan via catalog surgery, waits on the existing `auto_vac_worker_before_do_autovacuum` fault point, and asserts the orphan is swept while the live session's own temp table survives. The test fails without the fix (`swept` → `still present`).

## Commit 2 — resource-group io-limit bdi hash

Same pattern, worse odds: keys are `bdi_t` (`dev_t`), whose first byte is the minor number's low 8 bits. Whole-disk devices typically have minor 0 (`sda` 8:0, `nvme0n1` 259:0, `dm-0` 253:0), so their keys all read as the **empty string** — every such device collapses into a single hash entry, misattributing or losing per-device IO statistics. Found by auditing all 114 `hash_create()` call sites for this pattern.

## Commit 3 — enforce key semantics (PostgreSQL 14 hardening, backported)

`hash_create()` now asserts exactly one of `HASH_STRINGS` / `HASH_BLOBS` / `HASH_FUNCTION`, and that string keys are longer than 8 bytes — so the "integer hashed as a string" mistake fails fast at table creation in assert-enabled builds instead of silently corrupting lookups. All in-tree callers relying on the implicit string default (19 sites) now say `HASH_STRINGS` explicitly. No behavior change in production builds.

The assertion proved itself immediately: it caught one call site the static audit had missed (`InitShmemIndex`, exercised at bootstrap — initdb fails without the flag).

## Verification

* Assert-enabled build: `initdb` clean; full `installcheck-good` (669 tests) runs with **zero assertion failures and zero crashes**; the 8 failing tests are pre-existing environment/build-option diffs (tablespace path length, no-llvm/no-libxml/no-gpfdist build, hba view) unrelated to this branch, and the new `autovacuum-alive-session` test passes within the full schedule.
* Unfixed builds fail the new test with the bug's exact signature; under ASan the unfixed autovacuum worker aborts in `string_hash`.

### DIFF
--- a/src/test/regress/sql/autovacuum-alive-session.sql
+++ b/src/test/regress/sql/autovacuum-alive-session.sql
@@ -1,0 +1,123 @@
+-- The autovacuum orphan-temp-namespace sweep must compare session ids as
+-- fixed-size binary keys.  If the alive-session hash table falls back to
+-- string hashing/comparison, the comparison stops at the first zero byte
+-- of the key: an orphaned namespace whose numeric suffix differs from a
+-- live session id only in bytes past that zero byte is mistaken for the
+-- live session's namespace and never cleaned up.
+--
+-- Scenario: take our own session id S and find the first zero byte below
+-- its top byte.  Adding 256^p (p = 1-based position of that zero byte)
+-- flips a byte that string comparison never reaches, so the orphan suffix
+-- S + 256^p reads as the same C string as S.  Under string comparison the
+-- sweep wrongly considers the orphan alive; under binary comparison the
+-- orphan is swept.  Our real temp table must survive either way.
+--
+-- NOTE: this test must run inside an enable_autovacuum schedule window
+-- (it relies on the launcher being up); it manages only naptime itself,
+-- like the other autovacuum-* tests.
+SET client_min_messages = warning;
+CREATE EXTENSION IF NOT EXISTS gp_inject_fault;
+RESET client_min_messages;
+
+-- speed up test
+ALTER SYSTEM SET autovacuum_naptime = 5;
+SELECT pg_reload_conf();
+
+-- The string-alias offset for this session: 256^p where the p'th byte
+-- (1-based, little-endian) of the session id is the first zero byte below
+-- the top byte.  Such a byte exists for every session id below 65536 (CI
+-- clusters) and for larger ids whose low bytes happen to contain a zero;
+-- if none exists, the alias is mathematically inexpressible and we bail
+-- out with an explanation rather than construct a vacuous test.
+CREATE OR REPLACE FUNCTION alias_offset() RETURNS bigint AS $$
+DECLARE sid int; ofs bigint;
+BEGIN
+  SELECT sess_id INTO sid FROM pg_stat_activity WHERE pid = pg_backend_pid();
+  IF sid <= 0 THEN
+    RAISE EXCEPTION 'invalid session id %', sid;
+  END IF;
+  FOR p IN 1..3 LOOP
+    IF (sid / (256^(p-1))::bigint) % 256 = 0 THEN
+      ofs := (256^p)::bigint;
+      IF sid + ofs > 2147483647 THEN
+        RAISE EXCEPTION 'alias suffix for session id % overflows int4', sid;
+      END IF;
+      RETURN ofs;
+    END IF;
+  END LOOP;
+  RAISE EXCEPTION 'session id % has no zero byte below its top byte; the string-alias cannot be constructed -- rerun on a younger cluster', sid;
+END $$ LANGUAGE plpgsql STABLE;
+
+-- positive control: a live session's real temp table
+CREATE TEMP TABLE alive_probe(a int) DISTRIBUTED BY (a);
+INSERT INTO alive_probe SELECT generate_series(1, 10);
+
+-- fabricate the orphan: a plain table rewritten into a temp table living
+-- in a namespace whose suffix string-aliases our session id.  The catalog
+-- UPDATEs below run on the coordinator only (catalog DML is not
+-- dispatched); the segments keep the schema under its original name, and
+-- cleanup at the end relies on that.
+CREATE SCHEMA orphan_probe_schema;
+CREATE TABLE orphan_probe_schema.t(a int) DISTRIBUTED BY (a);
+SET allow_system_table_mods = on;
+DO $$
+DECLARE sid int;
+BEGIN
+  SELECT sess_id INTO sid FROM pg_stat_activity WHERE pid = pg_backend_pid();
+  UPDATE pg_namespace SET nspname = 'pg_temp_' || (sid + alias_offset())
+   WHERE nspname = 'orphan_probe_schema';
+  UPDATE pg_class SET relpersistence = 't'
+   WHERE oid = ('pg_temp_' || (sid + alias_offset()) || '.t')::regclass;
+END $$;
+RESET allow_system_table_mods;
+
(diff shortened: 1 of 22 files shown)


src/test/regress/sql/autovacuum-alive-session.sql
```diff
@@ -0,0 +1,94 @@
...
+-- fabricate the orphan: a plain table rewritten into a temp table living
+-- in a namespace whose suffix string-aliases our session id
+CREATE SCHEMA orphan_probe_schema;
+CREATE TABLE orphan_probe_schema.t(a int) DISTRIBUTED BY (a);
+SET allow_system_table_mods = on;
+DO $$
+DECLARE sid int;
+BEGIN
+  SELECT sess_id INTO sid FROM pg_stat_activity WHERE pid = pg_backend_pid();
+  UPDATE pg_namespace SET nspname = 'pg_temp_' || (sid + 16777216)
```
adam8157: This will not dispatched to segments, you will get an orphan_probe_schema.t on segments at the end, suggest renaming back then `DROP SCHEMA ... CASCADE` at the sweeping stage.

src/test/regress/sql/autovacuum-alive-session.sql
```diff
@@ -0,0 +1,94 @@
...
+-- positive control: a live session's real temp table
+CREATE TEMP TABLE alive_probe(a int) DISTRIBUTED BY (a);
+INSERT INTO alive_probe SELECT generate_series(1, 10);
+
+-- precondition for the byte-aliasing construction
+DO $$
+DECLARE sid int;
+BEGIN
+  SELECT sess_id INTO sid FROM pg_stat_activity WHERE pid = pg_backend_pid();
+  IF sid <= 0 OR sid >= 65536 THEN
```
adam8157: Why doesn't this fit the precondition? And the sid could be big in a long running cluster.

src/test/regress/sql/autovacuum-alive-session.sql
```diff
@@ -0,0 +1,94 @@
...
+-- fabricate an orphaned temp namespace with suffix S + 16777216 (4th byte
+-- set).  Under string comparison both keys read as the same C string, so
+-- the sweep wrongly considers the orphan alive; under binary comparison
+-- the orphan is swept.  Our real temp table must survive either way.
+SET client_min_messages = warning;
+CREATE EXTENSION IF NOT EXISTS gp_inject_fault;
+RESET client_min_messages;
+
+-- speed up test
+ALTER SYSTEM SET autovacuum = on;
```
adam8157: We use `enable_autovacuum` and `disable_autovacuum` cases to control this, doing RESET at the end of this case might affect others, if this case is placed in the enable window.

src/test/regress/sql/autovacuum-alive-session.sql
```diff
@@ -0,0 +1,94 @@
...
+
+-- cleanup (silent in both outcomes) and restore
+SET allow_system_table_mods = on;
+DO $$
+DECLARE sid int;
+BEGIN
+  SELECT sess_id INTO sid FROM pg_stat_activity WHERE pid = pg_backend_pid();
+  IF EXISTS (SELECT 1 FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace
+             WHERE c.relname = 't' AND n.nspname = 'pg_temp_' || (sid + 16777216)) THEN
+    EXECUTE format('DROP TABLE %I.t', 'pg_temp_' || (sid + 16777216));
```
adam8157: This will error out while reproducing, since catalog DDL is not dispatched to segments, suggest the same fix as above.

adam8157 (commented):
The fix looks great, all my comments are about the test case.

src/test/regress/sql/autovacuum-alive-session.sql
JunfengYang: Fixed in 55bf9b0bc9 as suggested: cleanup now renames the coordinator's copy back to orphan_probe_schema (the segments kept that name all along — the catalog UPDATEs never leave the coordinator) and then runs a single dispatched DROP SCHEMA orphan_probe_schema CASCADE. Measured on a demo cluster: zero leftovers on the coordinator and all three segments in both outcomes. Your prediction also got an involuntary live demo — an aborted mid-test run left orphan_probe_schema.t on every segment, which the old per-object cleanup would never have removed.

src/test/regress/sql/autovacuum-alive-session.sql
JunfengYang: Both points addressed in 55bf9b0bc9. Why the precondition exists: the string-alias requires a zero byte below the session id's top byte — string comparison stops at the first zero byte, so only a byte above it can differ invisibly; sid < 65536 was just one sufficient condition (byte 2 zero). The test now computes the offset generally (new alias_offset() helper finds the first zero byte among the low three and returns 256^p), which covers every sid below 65536 plus any larger sid whose low bytes contain a zero. For a sid with no zero byte below the top byte the alias is mathematically inexpressible — no integer other than sid itself has the same C-string view — so the helper raises with a message saying exactly that, rather than constructing a test that passes vacuously.

src/test/regress/sql/autovacuum-alive-session.sql
JunfengYang: Right — removed in 55bf9b0bc9. The test no longer touches the autovacuum GUC at all; it only sets/RESETs autovacuum_naptime, same as the other autovacuum-* tests. Its schedule slot is already inside the enable window (right after the autovacuum family, which follows the enable_autovacuum bracket), and a header note in the test now states that placement requirement.

src/test/regress/sql/autovacuum-alive-session.sql
JunfengYang: Fixed in 55bf9b0bc9, same rename-back + dispatched DROP SCHEMA CASCADE as the other thread. Validated the reproducing outcome explicitly: fabricating the orphan with a live foreign session's id as suffix drives the sweep to keep it (NOTICE 'orphan temp table still present'), after which the teardown completes with no errors and zero leftovers on any instance. One note on reproducing the original bug directly: after the hash_create() enforcement commit, reverting the fix aborts every autovacuum worker at table creation on assert builds (string keys now require keysize > 8), so the buggy pattern can no longer even start — the live-suffix construction above is the supported way to exercise the still-present path.

adam8157 (approved):

PR #213, opened 2026-08-09
Fix security checks in selectivity estimation functions.

linwen:
Backport of the upstream fix for CVE-2025-8713 — a planner-side information-disclosure flaw where selectivity estimation functions could run a user-supplied (potentially leaky) operator on pg_statistic data belonging to a table the current user is not allowed to read. WarehousePG (PG12 base) still carries the vulnerable code, and PG12 is past upstream EOL, so the fix is backported here from the oldest branch that has it (REL_13).

This PR is two commits: a small prerequisite infrastructure backport, then the CVE fix itself. They are kept separate (rather than squashed) so selfuncs.c matches upstream verbatim and the security change is reviewable in isolation.

The vulnerability

The original CVE-2017-7484 checks were performed against the child RTE in inheritance/partitioning cases. That is unsafe: securityQuals is NULL on an inheritance-child RTE, so those checks silently ignore RLS policies and security-barrier views. A user could learn statistical information (or trigger a leaky operator) about data they cannot select.

What's in this PR

1. Add a reverse-translation column number array to struct AppendRelInfo
Cherry-pick of upstream ce76c0ba53e. Adds AppendRelInfo.parent_colnos[] / num_child_cols, the reverse child→parent column mapping that the CVE fix relies on. Pure infrastructure, no behavioral change. Backporting this (one trivial conflict) lets selfuncs.c match upstream exactly instead of hand-rolling a helper.

2. Fix security checks in selectivity estimation functions
Cherry-pick of upstream 533211ded37 (CVE-2025-8713). The security checks are now done only against the parent RTE — consistent with how RLS policies and the executor's ACL checks work — via a new centralized all_rows_selectable() helper, applied uniformly in both selfuncs.c and the extended-stats code. View permissions (and, in the v13 line, top-parent permissions for inherited UPDATE/DELETE planned via inheritance_planner()) are additionally checked at planner startup, so a permissions error is raised before any estimation function runs.

WarehousePG adaptations
- Conflicts limited to WHPG's ExecutePlan() signature and the older extended_stats.c shape.
- Upstream's foreach_current_index() macro replaced with an explicit counter in planner.c.
- Upstream security semantics preserved exactly: checks go through all_rows_selectable() against the parent RTE, never the inheritance-child RTE.
- Added the GPORCA answer files (*_optimizer.out) for the three touched regress tests.
- Add AppendRelInfo to the ABI type ignore list

### DIFF
--- a/.abi-check/7.1.0/postgres.types.ignore
+++ b/.abi-check/7.1.0/postgres.types.ignore
@@ -10,3 +10,4 @@ struct AppendOnlyInsertDescData
 struct ResGroupCaps
 struct PlannerConfig
 struct ProjectionPath
+struct AppendRelInfo
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -171,7 +171,6 @@ static void ExecutePlan(EState *estate, PlanState *planstate,
 						ScanDirection direction,
 						DestReceiver *dest,
 						bool execute_once);
-static bool ExecCheckRTEPerms(RangeTblEntry *rte);
 static bool ExecCheckRTEPermsModified(Oid relOid, Oid userid,
 									  Bitmapset *modifiedCols,
 									  AclMode requiredPerms);
--- a/src/backend/nodes/copyfuncs.c
+++ b/src/backend/nodes/copyfuncs.c
@@ -2882,6 +2882,8 @@ _copyAppendRelInfo(const AppendRelInfo *from)
 	COPY_SCALAR_FIELD(child_reltype);
 	COPY_NODE_FIELD(translated_vars);
 	COPY_SCALAR_FIELD(parent_reloid);
+	COPY_SCALAR_FIELD(num_child_cols);
+	COPY_POINTER_FIELD(parent_colnos, from->num_child_cols * sizeof(AttrNumber));
 
 	return newnode;
 }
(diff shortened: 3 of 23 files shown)


haolinw (dismissed):
Looks good to me. I didn't find any external dependencies on `AppendRelInfo` so thinking it as an internal ABI change (but better to have one more confirmation, or consider reflecting it in release notes if we have done so previously?)

adam8157 (commented):
Please put new elements of `AppendRelInfo` at its end, to avoid ABI breakage as much as possible.

Even more, we can leverage `parent_colnos` to avoid struct changes at all, but since this is for planner, pg_hint_plan is probably the only consumer, it will be recompiled, and we already have planner-related struct in the ABI ignore list, I think it's safe enough to not do big refactor.

linwen:
> Please put new elements of `AppendRelInfo` at its end, to avoid ABI breakage as much as possible.
> 
> Even more, we can leverage `parent_colnos` to avoid struct changes at all, but since this is for planner, pg_hint_plan is probably the only consumer, it will be recompiled, and we already have planner-related struct in the ABI ignore list, I think it's safe enough to not do big refactor.

Thanks, fixed.

adam8157 (approved):

haolinw (approved):

PR #214, opened 2026-08-11
Collect CREATE PROTOCOL and ALTER TYPE ... SET DEFAULT ENCODING for ddl_command_end event triggers

Bonartze:
# Collect CREATE PROTOCOL and ALTER TYPE ... SET DEFAULT ENCODING for ddl_command_end event triggers

Follow-up to #211, which fixed the same defect for `CREATE EXTERNAL TABLE`.
## Problem

Two WHPG-specific cases in `ProcessUtilitySlow()` set neither `address` nor
`commandCollected`, so they fall through to the tail `EventTriggerCollectSimpleCommand()`
with nothing meaningful to collect and never appear in
`pg_event_trigger_ddl_commands()`:

1. **`CREATE PROTOCOL`** — `case T_DefineStmt` / `OBJECT_EXTPROTOCOL`, which calls void
   `DefineExtProtocol()`.
2. **`ALTER TYPE ... SET DEFAULT ENCODING`** — `case T_AlterTypeStmt`. GPDB's `AlterType()`
   returns void where upstream's returns `ObjectAddress`.

Before #211's `address = InvalidObjectAddress` initialization, both were the same
uninitialised-memory undefined behaviour as the reported external table crash — a silent
skip or a bogus-class error depending on the build. After it, they are deterministic silent
omissions from the audit trail. Correct, but still holes.

## Fix

Return the object address from both functions and assign it in the corresponding
`utility.c` case — the same shape as #211. For `AlterType()` this also brings the signature
back in line with upstream.

Neither needs new catalog plumbing: `OCLASS_EXTPROTOCOL` is already handled in
`objectaddress.c` by `getObjectDescription`, `getObjectTypeDescription` and
`getObjectIdentityParts`, and the type case is ordinary `TypeRelationId`.

## Testing

`src/test/regress/sql/event_trigger_gp.sql` extends the `ddl_command_end` case added in
#211 to cover both commands. Expected output generated by `pg_regress` against a patched
cluster:

```
          tag          |        identity         |      objtype      
-----------------------+-------------------------+-------------------
 CREATE EXTERNAL TABLE | public.endtest_ext      | foreign table
 CREATE TABLE          | public.endtest_heap     | table
 CREATE FUNCTION       | public.write_to_file()  | function
 CREATE FUNCTION       | public.read_from_file() | function
 CREATE PROTOCOL       | demoprot_end_test       | external protocol
 CREATE DOMAIN         | public.endtest_domain   | type
 ALTER TYPE            | public.endtest_domain   | type
(7 rows)
```

The `CREATE PROTOCOL` and `ALTER TYPE` rows are the new ones; on an unpatched build both
are absent while the surrounding `CREATE FUNCTION` / `CREATE DOMAIN` rows are present.

## Backports

Folded into the existing 7.4 / 7.3 fix branches alongside #211's commits, so
each version still needs only one PR.

### DIFF
--- a/src/backend/commands/extprotocolcmds.c
+++ b/src/backend/commands/extprotocolcmds.c
@@ -42,9 +42,10 @@
 /*
  *	DefineExtprotocol
  */
-void
+ObjectAddress
 DefineExtProtocol(List *name, List *parameters, bool trusted)
 {
+	ObjectAddress address;
 	char	   *protName;
 	List	   *readfuncName = NIL;
 	List	   *writefuncName = NIL;
@@ -110,6 +111,10 @@ DefineExtProtocol(List *name, List *parameters, bool trusted)
 									GetAssignedOidsForDispatch(),
 									NULL);
 	}
+
+	ObjectAddressSet(address, ExtprotocolRelationId, protOid);
+
+	return address;
 }
 
 /*
--- a/src/backend/commands/typecmds.c
+++ b/src/backend/commands/typecmds.c
@@ -3855,9 +3855,10 @@ AlterTypeNamespaceInternal(Oid typeOid, Oid nspOid,
  *
  * ALTER TYPE <typname> SET DEFAULT ENCODING (...)
  */
-void
+ObjectAddress
 AlterType(AlterTypeStmt *stmt)
 {
+	ObjectAddress address;
 	TypeName   *typname;
 	Oid			typid;
 	Oid			arrtypid;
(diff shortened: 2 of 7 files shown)


adam8157 (approved):
Verified: no early returns in either function ahead of the address assignment, both have a single caller, and the QE-side collection remains a no-op. The external protocol row in the expected output confirms the existing OCLASS_EXTPROTOCOL plumbing end-to-end. LGTM.

PR #219, opened 2026-08-11
Require superuser to install a non-built-in selectivity estimator (CVE-2026-2004)

linwen:
Summary:

Backport of the upstream fix for CVE-2026-2004 — a flaw where a non-superuser could attach a selectivity estimator that makes data-type assumptions to the wrong operator, causing the estimator to interpret pg_statistic data of an unexpected type. This can rise to the level of arbitrary code execution by any user able to create SQL objects. WarehousePG (PG12 base) still carries the vulnerable code, and PG12 is past upstream EOL, so the fix is backported here from REL_14 (the oldest branch that has it).

This PR is five commits: the main privilege fix, a small prerequisite infrastructure backport, the companion contrib/intarray hardening, and two small follow-up fixes to the same intarray selectivity function. They are kept separate (rather than squashed) so each cherry-pick matches upstream and the security change is reviewable in isolation.

The vulnerability:
Selectivity estimators come in two kinds: those that make assumptions about the data types they operate on, and those that don't. Most built-in estimators are of the latter kind and are safe on any operator. But an estimator that does assume a data type can malfunction if attached to an operator over a different type, since the pg_statistic data it reads won't be what it expects — a memory-safety hazard reachable by a user who can CREATE OPERATOR.

What's in this PR:
1. Require superuser to install a non-built-in selectivity estimator
Cherry-pick of upstream ea3bf349860 (CVE-2026-2004). Establishes two rules: built-in estimators must defend themselves against being called on the wrong data type (fixes a couple that didn't, in ts_selfuncs.c and network_selfuncs.c), and superuser privilege is now required to attach a non-built-in estimator via CREATE/ALTER OPERATOR. Since estimators are written in C — which already requires superuser to create — this has little practical impact on extensions.

2. Add a syscache on pg_extension.oid
Cherry-pick of upstream 9fa38c572a9 (same CVE). Prerequisite for the next commit, which needs the EXTENSIONOID syscache to track pg_extension updates. Upstream added this cache natively in v18 and backpatched it to 14–17; WHPG's PG12 base lacks it, so it is backported here.

3. Harden _int_matchsel() against being attached to the wrong operator
Cherry-pick of upstream 7e82d9a04d0 (same CVE). Prevents further abuse of any already-created operator exposing contrib/intarray's _int_matchsel to the wrong types, by verifying the query Const is actually a query_int. Since query_int is an extension type with no fixed OID, this adds infrastructure to securely look up the OID of a sibling datatype within the calling function's own extension (cached via the syscache from commit 2).

4. Fix incorrect lack of Datum conversion in _int_matchsel()
Cherry-pick of upstream 64dec8060c84. The empty-query path returned (Selectivity) 0.0 from a Datum-returning function instead of PG_RETURN_FLOAT8(0.0). Harmless on 64-bit (USE_FLOAT8_BYVAL) builds like WHPG's, and the path is believed unreachable, but included to keep _int_selfuncs.c byte-aligned with REL_14.

5. Fix missed ReleaseVariableStats() in intarray's _int_matchsel()
Cherry-pick of upstream a96b051a98e3 (bug #19492). The vartype != INT4ARRAYOID bail-out returned without releasing the fetched pg_statistic tuple, leaking a syscache pin (WARNING: cache reference leak) on every planning pass through it. This is exactly the path taken when the estimator is attached to a non-int4[] column — the wrong-operator scenario commit 3 hardens — so it belongs in this series.

WarehousePG adaptations:
- Main fix (ea3bf349860) — clean cherry-pick.
- Syscache (9fa38c572a9) — conflicts limited to the tails of the SysCacheIdentifier enum and the cacheinfo[] array, where WHPG carries its own APPENDONLYOID / ATTENCODINGNUM caches. EXTENSIONOID is appended after them, preserving upstream's intent of not disturbing existing syscache IDs (ABI-safe).
- Companion (7e82d9a04d0) — only conflict was the position of the new #include lines in _int_selfuncs.c.
- Follow-up fixes (64dec8060c84, a96b051a98e3) — clean cherry-picks.

### DIFF
--- a/contrib/intarray/_int_selfuncs.c
+++ b/contrib/intarray/_int_selfuncs.c
@@ -152,7 +152,10 @@ _int_matchsel(PG_FUNCTION_ARGS)
 	 * query_int.
 	 */
 	if (vardata.vartype != INT4ARRAYOID)
+	{
+		ReleaseVariableStats(vardata);
 		PG_RETURN_FLOAT8(DEFAULT_EQ_SEL);
+	}
 
 	/*
 	 * Can't do anything useful if the something is not a constant, either.
--- a/contrib/intarray/_int_selfuncs.c
+++ b/contrib/intarray/_int_selfuncs.c
@@ -190,7 +190,7 @@ _int_matchsel(PG_FUNCTION_ARGS)
 	if (query->size == 0)
 	{
 		ReleaseVariableStats(vardata);
-		return (Selectivity) 0.0;
+		PG_RETURN_FLOAT8(0.0);
 	}
 
 	/*
--- a/contrib/intarray/_int_selfuncs.c
+++ b/contrib/intarray/_int_selfuncs.c
@@ -19,6 +19,7 @@
 #include "catalog/pg_operator.h"
 #include "catalog/pg_statistic.h"
 #include "catalog/pg_type.h"
+#include "commands/extension.h"
 #include "utils/builtins.h"
 #include "utils/selfuncs.h"
 #include "utils/syscache.h"
@@ -171,7 +172,18 @@ _int_matchsel(PG_FUNCTION_ARGS)
 		PG_RETURN_FLOAT8(0.0);
 	}
 
-	/* The caller made sure the const is a query, so get it now */
+	/*
+	 * Verify that the Const is a query_int, else return a default estimate.
+	 * (This could only fail if someone attached this estimator to the wrong
+	 * operator.)
+	 */
+	if (((Const *) other)->consttype !=
+		get_function_sibling_type(fcinfo->flinfo->fn_oid, "query_int"))
+	{
+		ReleaseVariableStats(vardata);
+		PG_RETURN_FLOAT8(DEFAULT_EQ_SEL);
+	}
+
 	query = DatumGetQueryTypeP(((Const *) other)->constvalue);
 
 	/* Empty query matches nothing */
--- a/src/backend/catalog/pg_depend.c
+++ b/src/backend/catalog/pg_depend.c
@@ -22,11 +22,13 @@
 #include "catalog/pg_constraint.h"
 #include "catalog/pg_depend.h"
 #include "catalog/pg_extension.h"
+#include "catalog/pg_type.h"
 #include "commands/extension.h"
 #include "miscadmin.h"
 #include "utils/fmgroids.h"
 #include "utils/lsyscache.h"
 #include "utils/rel.h"
+#include "utils/syscache.h"
 
 
 static bool isObjectPinned(const ObjectAddress *object, Relation rel);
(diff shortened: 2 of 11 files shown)


contrib/intarray/_int_selfuncs.c
```diff
@@ -171,7 +172,18 @@ _int_matchsel(PG_FUNCTION_ARGS)
...
-	/* The caller made sure the const is a query, so get it now */
+	/*
+	 * Verify that the Const is a query_int, else return a default estimate.
+	 * (This could only fail if someone attached this estimator to the wrong
+	 * operator.)
+	 */
+	if (((Const *) other)->consttype !=
+		get_function_sibling_type(fcinfo->flinfo->fn_oid, "query_int"))
+	{
+		ReleaseVariableStats(vardata);
```
adam8157, line 186: This new punt correctly releases the stats, but the pre-existing `vardata.vartype != INT4ARRAYOID` punt a few lines up still returns without `ReleaseVariableStats(vardata)` — and that is exactly the path the wrong-operator scenario hardened here takes when the estimator is attached to a non-int4[] column. Upstream fixed this in a follow-up that landed after the CVE commits this PR picks: a96b051a98e3 ("Fix missed ReleaseVariableStats() in intarray's _int_matchsel().", bug #19492, backpatched through 14). Without it, every planning pass through that punt leaks a pg_statistic syscache pin (`WARNING: cache reference leak` at ResourceOwner cleanup). It's a clean three-line cherry-pick — worth adding as a 4th commit in this series.

adam8157:
One question on the `.abi-check/7.1.0/postgres.types.ignore` addition: `struct ExtensionSiblingCache` is translation-unit-local to extension.c and appears in no installed header, so abi-dumper's `-public-headers` filtering should not ever surface it — and a newly *added* type is not an ABI break in any case. Did abi-compare actually fail without this entry? If not, it might be better to drop it so future baseline directories do not inherit a suppression that silently widens what the ABI gate ignores.

contrib/intarray/_int_selfuncs.c
```diff
@@ -171,7 +172,18 @@ _int_matchsel(PG_FUNCTION_ARGS)
...
+		PG_RETURN_FLOAT8(DEFAULT_EQ_SEL);
+	}
+
 	query = DatumGetQueryTypeP(((Const *) other)->constvalue);
 
 	/* Empty query matches nothing */
```
adam8157, line 192: Just below this hunk, the empty-query path still does `return (Selectivity) 0.0;` from a Datum-returning function. REL_14, which this file otherwise now matches, uses `PG_RETURN_FLOAT8(0.0)` there — fixed by upstream 64dec8060c84 ("Fix incorrect lack of Datum conversion in _int_matchsel()"). The current code is only accidentally correct on `USE_FLOAT8_BYVAL` (64-bit) builds; without float8-by-val, `DatumGetFloat8` on Datum 0 dereferences a NULL pointer. Since this PR aligns the file with REL_14 anyway, consider cherry-picking that one-liner as well.

linwen: Thank you! All suggestion have been accepted.

adam8157 (approved):

PR #220, opened 2026-08-12
build: build on macOS and against libxml2 >= 2.12

adam8157:
Two build-environment differences that surface outside the Linux/glibc
toolchain CI builds with. Neither changes runtime behaviour.

### `s6_addr32` is a glibc extension

`ConvertToIPv4MappedAddr()` writes the mapped IPv4 address through
`in6_addr.s6_addr32`. glibc exposes that union member; macOS's
`<netinet6/in6.h>` names the members `__u6_addr8/16/32` and exposes only
`s6_addr`, so the build stops there:

    ic_udpifc.c:7024:21: error: no member named 's6_addr32' in 'struct in6_addr'

Writing through a `uint32` cast works on both, and matches how the
`0xffff` prefix is already stored two lines above.

### libxml2 2.12 added `const` to the error handler API

libxml2 2.12 changed `xmlStructuredErrorFunc` to take a
`const xmlError *`, so the four `xmlSetStructuredErrorFunc()` calls in
`xml.c` no longer type-check:

    xml.c:1025:45: error: incompatible function pointer types passing
    'void (void *, xmlErrorPtr)' to parameter of type 'xmlStructuredErrorFunc'
    (aka 'void (*)(void *, const struct _xmlError *)')

This is upstream PostgreSQL's fix, cherry-picked unchanged from
REL_12_STABLE (b2fd1dab90240ebb9017cd2fddd731c3641ba434). It carries the
accompanying removal of `contrib/xml2/xpath.c`'s assignment to
`xmlLoadExtDtdDefaultValue`, which 2.12 deprecates and which has had no
effect since external DTDs were disabled at a lower level.

This one is not macOS-specific — any platform shipping libxml2 >= 2.12
runs into it. Homebrew is currently on 2.15.3.

### DIFF
--- a/src/backend/cdb/motion/ic_udpifc.c
+++ b/src/backend/cdb/motion/ic_udpifc.c
@@ -7021,7 +7021,7 @@ ConvertToIPv4MappedAddr(struct sockaddr_storage *sockaddr, socklen_t *o_len)
 
 	((uint16 *) &in6_new->sin6_addr)[5] = 0xffff;
 
-	in6_new->sin6_addr.s6_addr32[3] = in->sin_addr.s_addr;
+	((uint32 *) &in6_new->sin6_addr)[3] = in->sin_addr.s_addr;
 	in6_new->sin6_scope_id = 0;
 
 	/* copy it back */
--- a/contrib/xml2/xpath.c
+++ b/contrib/xml2/xpath.c
@@ -75,7 +75,6 @@ pgxml_parser_init(PgXmlStrictness strictness)
 	xmlInitParser();
 
 	xmlSubstituteEntitiesDefault(1);
-	xmlLoadExtDtdDefaultValue = 1;
 
 	return xmlerrcxt;
 }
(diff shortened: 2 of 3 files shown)


src/backend/cdb/motion/ic_udpifc.c
```diff
@@ -7021,7 +7021,7 @@ ConvertToIPv4MappedAddr(struct sockaddr_storage *sockaddr, socklen_t *o_len)
 
 	((uint16 *) &in6_new->sin6_addr)[5] = 0xffff;
 
-	in6_new->sin6_addr.s6_addr32[3] = in->sin_addr.s_addr;
+	((uint32 *) &in6_new->sin6_addr)[3] = in->sin_addr.s_addr;
```
haolinw, line 7024: I'd prefer the following form for better cross-platform compatibility, though above fix works on macOS.
```c
memcpy(&in6_new->sin6_addr.s6_addr[12], &in->sin_addr.s_addr,
		   sizeof(in->sin_addr.s_addr));
```

haolinw (approved):

src/backend/cdb/motion/ic_udpifc.c
adam8157: ah, it's a bit better, but I decide to not change it, because:
1, line 7022 has the same pattern
2, the comment of this function says "Since this can be called in a signal handler, we avoid the use of async-signal unsafe functions such as memset/memcpy", I doubt it though.

PR #221, opened 2026-08-12
fix: stop gpload runs from failing with no error recorded

adam8157:
Two ways a gpload run could end with an exit code and nothing to explain
it. Both matter most to a caller that sends stdout to /dev/null, such as
an Informatica workflow, where the log file is meant to be the surviving
record.

### The shell wrapper split arguments on whitespace

`gpMgmt/bin/gpload` expanded `$*`, which joins the arguments into a single
string that the shell then splits again. A control file path containing a
space reached `gpload.py` as three arguments, so the option parser
rejected them and printed the usage text, before a session had started or
the log file had been opened:

```
$ gpload -f "/tmp/my configs/load.yml"      # before
COMMAND NAME: gpload

Runs a load job as defined in a YAML formatted control file.

$ gpload -f "/tmp/my configs/load.yml"      # after
2026-08-12 11:34:33|INFO|gpload session started 2026-08-12 11:34:33
2026-08-12 11:34:33|ERROR|A gpload control file processing error occurred. ...
```

`"$@"` keeps the original argument boundaries.

`gpload.bat` needs no equivalent change, because `%*` already preserves
the quoting of the original command line. Its "install python2.7" message
is corrected in passing, since gpload.py has run under Python 3 since
7306abea28. The probe itself stays as it is: the Python 3 installer on
Windows provides `python.exe`, which is what the wrapper looks for.

### An unusable -l path took the log with it

gpload gave up as soon as it could not open the log file named by `-l`,
and the reason only reached stdout. It now reopens the same file name
under `$HOME/gpAdminLogs` instead:

```
stdout   ...|WARN|could not use requested log file /nope/informatica.log, logging to /home/gpadmin/gpAdminLogs/informatica.log instead
stderr   gpload: could not open logfile /nope/informatica.log: [Errno 2] No such file or directory
```

The path given to `-l` is used as it stands: a directory we were not asked
to create is not created for us, so a mistyped argument surfaces as this
warning rather than as a stray directory left behind. The same WARN goes
into the log that was opened, so whoever reads that file later can tell it
is not the one that was asked for. gpload only gives up if both locations
fail.

### Tests

`gpMgmt/bin/gpload_test/gpload2/TEST_local_logging.py`, collected by
`make installcheck`. All three cases are settled before gpload connects to
the database, so none of them needs a running cluster. Each fix carries
its own check, and every check was confirmed to fail against an unpatched
gpload.

### DIFF
--- a/gpMgmt/bin/gpload.py
+++ b/gpMgmt/bin/gpload.py
@@ -1177,18 +1177,46 @@ class gpload:
 
         # default to gpAdminLogs for a log file, may be overwritten
         if self.options.l is None:
-            self.options.l = os.path.join(os.environ.get('HOME', '.'),'gpAdminLogs')
-            if not os.path.isdir(self.options.l):
-                os.mkdir(self.options.l)
-
-            self.options.l = os.path.join(self.options.l, 'gpload_' + \
+            self.options.l = os.path.join(os.environ.get('HOME', '.'), 'gpAdminLogs',
+                                          'gpload_' + \
                                           datetime.date.today().strftime('%Y%m%d') + '.log')
 
-        try:
-            self.logfile = open(self.options.l,'a')
-        except Exception as e:
-            self.log(self.ERROR, "could not open logfile %s: %s" % \
-                      (self.options.l, e))
+        # Make sure we always end up with an open log file. The path given to -l
+        # is used as it stands: a directory we were not asked to create is not
+        # created for us, so a typo does not quietly leave one behind. When that
+        # path cannot be opened, reopen the same file name under
+        # $HOME/gpAdminLogs, which gpload creates for itself anyway, so that the
+        # error output is never silently lost.
+        self.logfile = None
+        requested = self.options.l
+        base = os.path.basename(requested) or \
+               ('gpload_' + datetime.date.today().strftime('%Y%m%d') + '.log')
+        fallback = os.path.join(os.environ.get('HOME', '.'), 'gpAdminLogs', base)
+
+        candidates = [requested]
+        if fallback != requested:
+            candidates.append(fallback)
+
+        for logpath in candidates:
+            try:
+                if logpath == fallback:
+                    os.makedirs(os.path.dirname(fallback), exist_ok=True)
+                self.logfile = open(logpath, 'a')
+                self.options.l = logpath
+                break
+            except Exception as e:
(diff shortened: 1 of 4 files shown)


gpMgmt/bin/gpload.py
huiliang-liu: We don't have to `mkdir` here, it can be handled in the following new codes.

gpMgmt/bin/gpload.py
```diff
@@ -1184,11 +1184,38 @@ def __init__(self,argv):
...
+        self.logfile = None
+        requested = self.options.l
+        base = os.path.basename(requested) or \
+               ('gpload_' + datetime.date.today().strftime('%Y%m%d') + '.log')
+        fallback = os.path.join(os.environ.get('HOME', '.'), 'gpAdminLogs', base)
+        for logpath in [requested, fallback]:
+            try:
+                logdir = os.path.dirname(logpath)
+                if logdir and not os.path.isdir(logdir):
+                    os.makedirs(logdir)
```
huiliang-liu: It is better to use `os.makedirs(logdir, exist_ok=True)` to avoid potentially risk condition of two gploads. One question: should we create the requested path silently if it doesn't exist? I mean if the user use typo arg, should we create it too? or just fallback to default or print a warning message?

gpMgmt/bin/gpload.py
adam8157: Good catch, dropped. The default path now goes straight to `$HOME/gpAdminLogs/gpload_<date>.log`, and the loop below creates that directory when it is the one being opened, so the behaviour on a fresh machine is unchanged.

gpMgmt/bin/gpload.py
adam8157: Took `exist_ok=True` for the fallback, which is now the only place we create anything.

On your question: agreed, and I went a step further — we no longer create the directory named by `-l` at all. The original code never did, and creating it meant that a typo such as `-l /tmp/gplaod/x.log` quietly left a directory behind. Now that path simply fails to open, we warn with both paths, and the log lands in `$HOME/gpAdminLogs`, so the mistake stays visible while the error output is still preserved.

Added `test_missing_log_directory_is_not_created` to pin that down: it asserts the mistyped directory still does not exist afterwards, and that the fallback log was written. It fails against the earlier `makedirs` version.

huiliang-liu (approved):

PR #222, opened 2026-08-12
fix: stop gpload runs from failing with no error recorded (6X)

adam8157:
Backport of #221 to `WHPG_6X_STABLE`. Two ways a gpload run could end
with an exit code and nothing to explain it, which matter most to a
caller that sends stdout to /dev/null, such as an Informatica workflow,
where the log file is meant to be the surviving record.

### The shell wrapper split arguments on whitespace

`gpMgmt/bin/gpload` expanded `$*`, which joins the arguments into a single
string that the shell then splits again. A control file path containing a
space reached `gpload.py` as three arguments, so the option parser
rejected them and printed the usage text, before a session had started or
the log file had been opened. `"$@"` keeps the original argument
boundaries.

### An unusable -l path took the log with it

gpload gave up as soon as it could not open the log file named by `-l`,
and the reason only reached stdout. It now reopens the same file name
under `$HOME/gpAdminLogs`, naming both paths on stdout and in the log that
was opened. The path given to `-l` is used as it stands: a directory we
were not asked to create is not created for us, so a mistyped argument
surfaces as a warning rather than as a stray directory left behind. gpload
only gives up if both locations fail.

### Differences from #221

- `gpload.bat` is left alone. On this branch gpload.py really is a
  Python 2 program, so its "install python2.7" message is correct here.
- The new code uses this file's `except Exception, e` spelling, and
  creates `$HOME/gpAdminLogs` with the `try/except OSError` idiom rather
  than `os.makedirs(exist_ok=True)`, which is Python 3.2 and later.
- The test uses `tempfile.mkdtemp()` rather than `TemporaryDirectory`, so
  that it runs under the `python2 -m pytest TEST_local_*` that
  `gpload_test/Makefile` invokes on this branch.

### Tests

`gpMgmt/bin/gpload_test/gpload2/TEST_local_logging.py`, collected by
`make installcheck`. All three cases are settled before gpload connects to
the database, so none of them needs a running cluster.

Run under Python 2.7.18 on Rocky Linux 9: all three pass against this
branch, and all three fail against `WHPG_6X_STABLE` as it stands.

### DIFF
--- a/gpMgmt/bin/gpload
+++ b/gpMgmt/bin/gpload
@@ -6,5 +6,5 @@ if [ ! -z "$GPHOME_LOADERS" ]; then
     . $GPHOME_LOADERS/greenplum_loaders_path.sh
 fi
 
-exec gpload.py $*
+exec gpload.py "$@"
 
--- a/gpMgmt/bin/gpload.py
+++ b/gpMgmt/bin/gpload.py
@@ -1245,18 +1245,52 @@ class gpload:
 
         # default to gpAdminLogs for a log file, may be overwritten
         if self.options.l is None:
-            self.options.l = os.path.join(os.environ.get('HOME', '.'),'gpAdminLogs')
-            if not os.path.isdir(self.options.l):
-                os.mkdir(self.options.l)
-
-            self.options.l = os.path.join(self.options.l, 'gpload_' + \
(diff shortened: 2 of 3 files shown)


huiliang-liu (approved):
Looks good to me

PR #223, opened 2026-08-12
fix: dispatch ALTER OPERATOR ... SET (RESTRICT/JOIN) to segments

adam8157:
## Summary

`ALTER OPERATOR ... SET (RESTRICT/JOIN)` updated `pg_operator` on the coordinator only, leaving every segment with the stale estimator OIDs. This adds the missing dispatch at the end of `AlterOperator()` (mirroring `DefineOperator()`), the `AlterOperatorStmt` serialization support it needs, and a regress test.

## Background

Found while reviewing #219 (CVE-2026-2004): the new superuser gate on estimator changes runs QD-only for the ALTER path because the statement never reaches segments. Same class of bug as #205 (`ALTER COLLATION ... REFRESH VERSION`), fixed with the same recipe. I audited the remaining utility statements against their dispatch call sites: the only objects left QD-only are event triggers and extended statistics, which is by design (`shouldDispatchForObject`), so no other statement has this gap.

## Test plan

- New `gp_alter_operator` regress test cross-checks `pg_operator` on the QD against `gp_dist_random('pg_operator')` after `SET` and after `SET ... NONE`: it fails with "mismatch on seg 0/1/2" on an unpatched build and passes patched.
- `installcheck-small`: 194/194 passed.

### DIFF
--- a/src/backend/commands/operatorcmds.c
+++ b/src/backend/commands/operatorcmds.c
@@ -571,5 +571,20 @@ AlterOperator(AlterOperatorStmt *stmt)
 
 	table_close(catalog, NoLock);
 
+	/*
+	 * Update every segment's pg_operator too, or QE-local planning (SPI in
+	 * EXECUTE ON ALL SEGMENTS functions, utility-mode connections) keeps
+	 * using the stale oprrest/oprjoin.  The sibling DefineOperator()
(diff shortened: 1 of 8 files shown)


linwen (approved):
Looks good to me!

PR #227, opened 2026-08-17
Fix "ORDER/GROUP BY expression not found in targetlist" for correlated EXISTS sublinks

SKeyerror:
## Problem

```sql
CREATE TABLE foo (a int, b int);
CREATE TABLE bar (c int, d int);
CREATE TABLE buz (x int, y int);

EXPLAIN SELECT * FROM foo
WHERE EXISTS (SELECT max(d) FROM bar JOIN buz ON bar.c = foo.b
              WHERE foo.a = bar.c GROUP BY d);
-- ERROR:  ORDER/GROUP BY expression not found in targetlist (tlist.c:421)
```

## Root cause

Unlike upstream PostgreSQL, GPDB's `simplify_EXISTS_query()` also accepts subqueries with aggregates (the `hasAggs` rejection is `#if 0`-ed out). For a **correlated** sublink it unconditionally throws away the targetlist, but it only removed the GROUP BY clause when the subquery had **no** aggregates. A correlated subquery with both an aggregate and GROUP BY was therefore left in an inconsistent state: `targetList = NIL` while `groupClause` still carried sortgroup references pointing into the discarded targetlist.

This usually goes unnoticed because `convert_EXISTS_sublink_to_join()` pulls the sublink up as a semi-join and drops the groupClause wholesale. But when the pull-up is rejected — here the correlated var `foo.b` appears in a JOIN/ON clause rather than in WHERE — the query falls back to `make_subplan()`, and planning the subquery fails in `standard_qp_callback()` → `make_pathkeys_for_sortclauses(root, groupClause, tlist = NIL)`.

## Fix

When the targetlist of a correlated EXISTS sublink is discarded and the subquery has a GROUP BY, discard the GROUP BY and the aggregation altogether (`groupClause = NIL; hasAggs = false`). This preserves semantics: a grouped query (grouping sets are refused earlier in the function) returns one row per group, so it returns at least one row **iff** its FROM/WHERE clauses yield at least one row — which is all EXISTS cares about. No aggregate expressions can remain anywhere in the query at that point (targetlist is empty; HAVING was demoted to WHERE or we gave up; ORDER BY/DISTINCT/window clauses are discarded too), so clearing `hasAggs` is safe.

The aggregate-without-GROUP-BY case is deliberately left alone: such a subquery always returns exactly one row, so it must keep aggregating (see the existing comment).

### DIFF
--- a/src/backend/optimizer/plan/subselect.c
+++ b/src/backend/optimizer/plan/subselect.c
@@ -1946,15 +1946,36 @@ simplify_EXISTS_query(PlannerInfo *root, Query *query)
 	/*
 	 * Delete GROUP BY if no aggregates.
 	 *
-	 * Note: It's important that we don't clear hasAggs, even though we
-	 * removed any possible aggregates from the targetList! If you have a
-	 * subquery like "SELECT SUM(foo) ...", we don't need to compute the sum,
-	 * but we must still aggregate all the rows, and return a single row,
-	 * regardless of how many input rows there are. (In particular, even
-	 * if there are no input rows).
+	 * Note: If the subquery has aggregates, it's usually important that we
+	 * don't clear hasAggs. For an uncorrelated sublink we kept the
+	 * targetlist above, so any aggregates are still in it; and even for a
+	 * correlated sublink, whose targetlist we threw away, a subquery like
+	 * "SELECT SUM(foo) ..." without GROUP BY must still aggregate all the
+	 * rows and return a single row, regardless of how many input rows there
+	 * are. (In particular, even if there are no input rows). Only for a
+	 * correlated sublink that also has a GROUP BY clause can the aggregation
+	 * be discarded along with it; see below.
 	 */
 	if (!query->hasAggs)
 		query->groupClause = NIL;
+	else if (is_correlated && query->groupClause)
+	{
+		/*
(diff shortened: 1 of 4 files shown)


SKeyerror:
@24nishant Hi, this is for 7X, the same as https://github.com/EnterpriseDB/warehouse-pg-next/pull/331, and updated the comments too, could take a look? Thanks!

src/backend/optimizer/plan/subselect.c
```diff
@@ -1955,6 +1955,24 @@ simplify_EXISTS_query(PlannerInfo *root, Query *query)
 	 */
 	if (!query->hasAggs)
 		query->groupClause = NIL;
+	else if (is_correlated && query->groupClause)
```
24nishant, line 1958: [nit] we should update the comment above

24nishant (dismissed):
LGTM!

src/backend/optimizer/plan/subselect.c
SKeyerror: Done

SKeyerror:
@24nishant Hi, since I pushed the codebase after your approve, please take a look again, thanks!

24nishant (approved):
LGTM!

SKeyerror:
Thanks for the review!

PR #230, opened 2026-08-18
Backport CVE-2026-14677 fix: integer wraparound in pltcl/plperl allocations

SKeyerror:
## Summary

Backport the fix for **CVE-2026-14677** (CVSS 8.8) from PostgreSQL REL_14_STABLE.

On 32-bit builds, memory-allocation size calculations in pltcl and plperl (e.g. `palloc(nargs * sizeof(FmgrInfo))`) can wrap around with crafted function bodies, causing the server to undersize an allocation and write out-of-bounds — potentially executing arbitrary code as the OS user running the database. Upstream fixed this in 14.24 / 15.19 / 16.15 / 17.11 / 18.6 (2026-08-13). PostgreSQL 12 is EOL, so warehouse-pg 7 (PG 12.12 base) needs a manual backport. Note that 64-bit builds are not practically exploitable; this is hardening/compliance, but scanners flag the version regardless.

This is a follow-up to CVE-2026-6473; two of the three commits below are the prerequisites that CVE's fix brought into REL_14/15.

## Cherry-picked commits (in order, each with `-x` upstream SHA)

1. **`b7f37af7c1` — Expand palloc/pg_malloc API for more type safety** (Peter Eisentraut)
   Introduces `palloc_object()` / `palloc_array()` / `palloc0_array()` / `repalloc_array()` etc. Clean apply.

2. **`61a9b4b6e4` — Make palloc_array() and friends safe against integer overflow** (Tom Lane, CVE-2026-6473 series)
   Makes the macros overflow-checked; moves `add_size()`/`mul_size()` from shmem.c into mcxt.c and adds `repalloc_extended()`. Without this commit, the conversion in commit 3 would not actually fix anything.

3. **`aff9dac1c1` — Use palloc_array() in pltcl and plperl to avoid overflow** (Heikki Linnakangas, **the CVE-2026-14677 fix**)
   Converts all multiply-in-palloc sites in `pltcl.c` (12 lines), `plperl.c` (~16 lines), `SPI.xs` (3 lines); `pltcl_quote()` switches to `add_size(mul_size(...))`. **Clean apply, zero conflicts** — these three files carry no GPDB-specific modifications.

## Conflict resolution (commit 2 only)

- **`src/backend/storage/ipc/shmem.c`**: upstream removes `add_size`/`mul_size` here (moved to mcxt.c). The conflict arose only because PG 12 lacks `pg_get_shmem_allocations()` (added in PG 13), which upstream's context expected after those functions. Resolved by deleting the two functions, matching upstream intent exactly.

- **`src/include/utils/palloc.h`**: PG 14 declares `repalloc()` with `pg_nodiscard`, PG 12 has no such macro. Resolved by:
  - adding the upstream `pg_nodiscard` definition to `src/include/c.h` (verbatim from PG 14; needed by the new declarations this commit adds), and
  - **keeping our existing `repalloc()` declaration unmarked** — decorating it would emit new `-Wunused-result` warnings at every existing call site that ignores the return value. All newly added functions carry `pg_nodiscard` as upstream intended.

- Prerequisites `pg_add_size_overflow` / `pg_mul_size_overflow` (`common/int.h`) and `pg_noinline` already exist in this tree — no extra backport needed.

## Verification

- Full incremental build + install with **zero compiler warnings** (the `c.h` change forced a near-full recompile).
- `src/pl/plperl` installcheck against the demo cluster: **all 14 tests passed** (exercises the converted `palloc_array` paths).
- pltcl: local build is configured without `--with-tcl`, so installcheck cannot run here; the modified `pltcl.c` passes a full `gcc -fsyntax-only` compile with tcl 8.6 headers. The change is a line-for-line match of upstream's mechanical conversion; CI/release builds with tcl enabled will provide real coverage.
- Upstream added no regression tests for this fix (the overflow is only constructible on 32-bit).

## CI fix: unit-test mock generator vs `pg_noinline`

The first CI run failed compiling `mcxt_mock.c`: `mocker.py`'s function-parsing regex does not know the `pg_noinline` keyword (used by the backported `add_size_error()`/`mul_size_error()` in mcxt.c). Worse, the regex re-matches mid-keyword at its `inline` suffix, emitting a dangling `static pg_no` into the generated mock. Fixed by stripping the `pg_noinline` token in `CFile.strip()`, the same way `__attribute__((...))` is already removed — this keeps the cherry-picked mcxt.c byte-identical to upstream. `catcache.c`/`float.c`/`expandedrecord.c` also use `pg_noinline` but are not currently mocked, so only mcxt.c was affected. Verified locally with the full `make unittest-check`.

(Local note: `cdbsendonebyte.t` aborts locally on an `interconnect_address` assertion — that is pre-existing main behavior in `--enable-cassert` builds, unrelated to this PR, and compiled out in the RPM CI build.)

## Note

The rest of the CVE-2026-6473 series (tsvector/tsquery/Levenshtein/regexp hardening commits on REL_14_STABLE) is likely also missing from this tree and should be handled as a separate task.


SKeyerror:
CI has passed: https://github.com/EnterpriseDB/warehouse-pg-packaging/actions/runs/32099800362

linwen (commented):
The code changes look good to me.
One suggestion: please fold 68351eed841 ("mocker: strip pg_noinline ...") into
0c043c19b06 ("Make palloc_array() and friends safe against integer
overflow"), to keep every commit self-contained and buildable.

SKeyerror:
> The code changes look good to me. One suggestion: please fold [68351ee](https://github.com/warehouse-pg/warehouse-pg/commit/68351eed8417800b84f97764ace986b5e31ccff7) ("mocker: strip pg_noinline ...") into [0c043c1](https://github.com/warehouse-pg/warehouse-pg/commit/0c043c19b06d89a8d0bcea3a619db0f500ce14ac) ("Make palloc_array() and friends safe against integer overflow"), to keep every commit self-contained and buildable.

Thanks for the advice, already amended.

linwen (approved):

PR #231, opened 2026-08-18
Bounds-check multibyte character lengths (CVE-2025-4207, CVE-2026-2006)

adam8157:
Summary:

Backport of the upstream fixes for CVE-2025-4207 and CVE-2026-2006, two
multibyte-length overreads. They are sent together because the second builds
directly on the first and they touch the same files. WarehousePG's PG12 base is
past upstream EOL, so each is taken from the oldest upstream branch that has it.

Five commits:

1. With GB18030, prevent SIGSEGV from reading past end of allocation.
   Cherry-pick of upstream cbadeaca927 (REL_13, CVE-2025-4207).
2. Avoid fetching one past the end of translate()'s "to" parameter.
   Cherry-pick of upstream b162660d3ac (REL_12).
3. Replace pg_mblen() with bounds-checked versions.
   Cherry-pick of upstream cecedb912ae (REL_14, CVE-2026-2006).
4. Fix overread in JSON parsing errors for incomplete byte sequences.
   Cherry-pick of upstream 377c25d3227 (REL_13).
5. orafce: replace pg_mblen() with bounds-checked versions.
   WarehousePG-specific, no upstream counterpart.

Commit 2 is a prerequisite, not a CVE fix: it is a plain upstream cherry-pick we
never took, and commit 3 rewrites the same translate() loop and does not apply
without the to_end variable it introduces. It also removes a real one-byte
overread we were carrying (translate('abc', 'xb', '') reads to_ptr with tolen
== 0).

Commits 4 and 5 finish the job in the two places the cherry-picks cannot reach.
Neither is a hunk that failed to apply -- the file sets match upstream exactly,
41 files and 280 of its 281 hunks, the one exception being a formatting.c hunk
for datetime "standard mode" separator code that PG12 does not have. But
upstream's *equivalent* code is safe for reasons that never reached PG12, so
without these two commits our tree would still carry 22 unbounded pg_mblen()
call sites that upstream no longer has:

- The JSON lexer is still backend-only in src/backend/utils/adt/json.c here; it
  became src/common/jsonapi.c in PG13. Upstream's copy is safe because of
  CVE-2025-4207 plus 377c25d3227, and neither has a PG12 backpatch. Commit 4
  takes the latter.

- gpcontrib/orafce is a vendored fork of several of the very functions commit 3
  rewrites. orafce/regexp.c's parse_re_flags() and charpad.c's lpad()/rpad()
  loops are copies of src/backend/utils/adt/regexp.c and oracle_compat.c, so
  they are the same CVE-2026-2006 pattern in a file the patch cannot touch.
  Commit 5 converts them, choosing the same variant upstream chose for each
  corresponding site.

Background:

pg_encoding_mblen() has to look at two bytes to decide a GB18030 character's
length, so callers handed unterminated input could read one byte past the
allocation (CVE-2025-4207). More generally, pg_mblen() returns a length that its
callers never checked against the bytes actually remaining, so a corrupted
string could make an iterating caller overrun its buffer (CVE-2026-2006).

The fixes add bounds-checked variants — pg_encoding_mblen_or_incomplete(), and
pg_mblen_cstr() / _range() / _with_len() / _unbounded() — and convert every core
caller to whichever one matches what it actually knows about its input.
pg_mblen(), the COPYCHAR macro and the t_isXXX() functions remain for
compatibility but are deprecated and no longer used by core code.

No catalog, ABI or WAL changes.

WarehousePG adaptations:

The per-file details are in the commit messages. The one deviation worth calling
out here: upstream keeps report_invalid_encoding_db() static in mbutils.c, but
report_invalid_encoding() still lives in wchar.c for us (it moved when wchar.c
became src/common/wchar.c in PG13), while the new pg_mblen_*() functions belong
in mbutils.c. It is therefore extern here, declared in pg_wchar.h next to
report_invalid_encoding(). Everything else is either a plain cherry-pick or a
mechanical re-application caused by PG12 file layout.

Commit 4 carries one hunk beyond upstream's diff, for extract_mb_char(). It has
exactly the overread the commit sets out to remove, one function away from the
one upstream fixed; upstream has no counterpart hunk only because PG13's
refactor deleted the function outright.

Test plan:

- Clean make and make install, no new warnings.
- Full make installcheck against a gpdemo cluster. Everything passes except
  gporca, qp_full_join, dpe and most of contrib/btree_gist, whose diffs are all
  "Server has been compiled without ORCA" / "Optimizer: Postgres-based planner"
  against ORCA-generated expected files — my box has no ORCA libraries, so those
  are down to CI. contrib/pageinspect likewise fails only on GPDB artifacts
  (DISTRIBUTED BY notices, "block number 0 is out of range" because the pages
  live on segments). No data or error-message differences anywhere, and nothing
  touching text_to_bits() or gbt_var_node_cp_len(). contrib dict_xsyn, hstore,
  ltree, pg_trgm and unaccent all pass outright.
- src/test/modules/test_escape: 1031 assertions pass, 0 fail, 208 of them
  GB18030 — including the six new truncated-sequence vectors and the 128 KiB
  page-boundary PQescapeLiteral() case that the CVE-2025-4207 commit adds. The
  run then stops early on mule_internal, because upstream's driver wants a
  SQL_ASCII database and we reject SQL_ASCII as a server encoding; that is a
  pre-existing limitation of this module in our tree, not a regression.
- Guard-page check for the libpq half of CVE-2025-4207: a GB18030 lead byte
  placed as the last byte of a mapped page with the next page PROT_NONE.
  PQescapeStringConn(), PQescapeLiteral() and PQescapeIdentifier() all complete
  without faulting, while PQmblen() — plain pg_encoding_mblen(), i.e. exactly
  what those functions called before the fix — takes SIGSEGV on the same
  pointer. Confirms both that the guard is armed and that the overread is gone.
- Commit 4: the multibyte JSON error paths still render correctly -- a valid
  multibyte document round-trips, '{"k":"\uZZ..."}' reports the hex-digit error
  with its CONTEXT line intact, and a backslash followed by a multibyte
  character reaches extract_mb_char() and prints that character whole in the
  detail.
- Commit 5: orafce's own suite passes 17/17, and an A/B against an unpatched
  orafce build -- multibyte lpad/rpad across every width with multibyte padding,
  plvstr.normalize, plvstr.rvrs/left/right, plvchr._is_kind, oracle.length,
  to_multi_byte/to_single_byte round trips and both regexp option error messages
  -- is byte-identical.
- No discriminating SQL-level test is available for CVE-2026-2006 on PG12: the
  reachable error messages are identical before and after, since the bogus
  length is clamped by Min(mblen, len) either way, and the v14+ conversion.sql
  coverage rides on the noError conversion API that PG12 does not have. Upstream
  ships no new tests for it either.

What commits 4 and 5 actually fix:

Most of these reads are harmless in themselves: for any encoding a database can
use, .mblen() looks only at the leading byte. The damage comes from using the
returned length afterwards. Three sites do worse than overread:

- json.c set token_terminator to s + pg_mblen(s) unclamped, and
  report_json_context() memcpy()s everything up to token_terminator out of a
  buffer that is a bare text payload, not null-terminated. extract_mb_char()
  has the same problem building the "Escape sequence ... is invalid" detail.
- orafce's plvstr_normalize() copies the character out byte by byte into a
  buffer sized palloc(VARSIZE_ANY_EXHDR(str)), so an overshooting length reads
  past the input *and* writes past that allocation.
- orafce's to_single_byte() memcpy()s clen bytes into a buffer sized
  VARHDRSZ + srclen, overrunning both sides likewise.

Commit 4 keeps pg_mblen_unbounded() plus an explicit clamp rather than the
erroring pg_mblen_range(), matching what upstream's lexer does: the point is to
report the JSON syntax error we came here for, not to replace it with "invalid
byte sequence".

charpad.c's two string1 copy loops in commit 5 needed a real guard rather than
just a different call, since the loop counter also covers a half-width space and
so is not by itself a bound on string1. That guard cannot fire for input valid in
the database encoding; it earns its place on truncated input, where it stops
pg_mblen_range() from being called on a pointer already at the end of the datum.

Deliberately left alone:

- src/backend/parser/scan.l's three pg_mblen() calls are correct as they stand.
  They measure a freshly written, deliberately non-terminated UTF-8 sequence, so
  pg_mblen_cstr() there would be a bug. Upstream reads strlen() at the
  equivalent spot only because PG13's a6525588b7c started null-terminating that
  buffer, which is a feature commit unrelated to any of this.

After the series the only bare pg_mblen() references left in the tree are those
three, plus the deprecated function's own definition and declaration.

### DIFF
--- a/contrib/btree_gist/btree_utils_var.c
+++ b/contrib/btree_gist/btree_utils_var.c
@@ -117,36 +117,47 @@ gbt_var_leaf2node(GBT_VARKEY *leaf, const gbtree_vinfo *tinfo, FmgrInfo *flinfo)
 
 /*
  * returns the common prefix length of a node key
+ *
+ * If the underlying type is character data, the prefix length may point in
+ * the middle of a multibyte character.
 */
 static int32
 gbt_var_node_cp_len(const GBT_VARKEY *node, const gbtree_vinfo *tinfo)
 {
 	GBT_VARKEY_R r = gbt_var_key_readable(node);
 	int32		i = 0;
-	int32		l = 0;
+	int32		l_left_to_match = 0;
+	int32		l_total = 0;
 	int32		t1len = VARSIZE(r.lower) - VARHDRSZ;
 	int32		t2len = VARSIZE(r.upper) - VARHDRSZ;
 	int32		ml = Min(t1len, t2len);
 	char	   *p1 = VARDATA(r.lower);
 	char	   *p2 = VARDATA(r.upper);
+	const char *end1 = p1 + t1len;
+	const char *end2 = p2 + t2len;
 
 	if (ml == 0)
 		return 0;
 
 	while (i < ml)
 	{
-		if (tinfo->eml > 1 && l == 0)
+		if (tinfo->eml > 1 && l_left_to_match == 0)
 		{
-			if ((l = pg_mblen(p1)) != pg_mblen(p2))
+			l_total = pg_mblen_range(p1, end1);
+			if (l_total != pg_mblen_range(p2, end2))
 			{
 				return i;
 			}
+			l_left_to_match = l_total;
 		}
 		if (*p1 != *p2)
 		{
 			if (tinfo->eml > 1)
 			{
-				return (i - l + 1);
+				int32		l_matched_subset = l_total - l_left_to_match;
+
+				/* end common prefix at final byte of last matching char */
+				return i - l_matched_subset;
 			}
 			else
 			{
@@ -156,7 +167,7 @@ gbt_var_node_cp_len(const GBT_VARKEY *node, const gbtree_vinfo *tinfo)
 
 		p1++;
 		p2++;
-		l--;
+		l_left_to_match--;
 		i++;
 	}
 	return ml;					/* lower == upper */
--- a/contrib/dict_xsyn/dict_xsyn.c
+++ b/contrib/dict_xsyn/dict_xsyn.c
@@ -48,15 +48,15 @@ find_word(char *in, char **end)
 	char	   *start;
 
 	*end = NULL;
-	while (*in && t_isspace(in))
-		in += pg_mblen(in);
+	while (*in && t_isspace_cstr(in))
+		in += pg_mblen_cstr(in);
 
 	if (!*in || *in == '#')
 		return NULL;
 	start = in;
 
-	while (*in && !t_isspace(in))
-		in += pg_mblen(in);
+	while (*in && !t_isspace_cstr(in))
+		in += pg_mblen_cstr(in);
 
 	*end = in;
 
--- a/contrib/hstore/hstore_io.c
+++ b/contrib/hstore/hstore_io.c
@@ -82,7 +82,7 @@ get_val(HSParser *state, bool ignoreeq, bool *escaped)
 			else if (*(state->ptr) == '=' && !ignoreeq)
 			{
 				elog(ERROR, "Syntax error near \"%.*s\" at position %d",
-					 pg_mblen(state->ptr), state->ptr,
+					 pg_mblen_cstr(state->ptr), state->ptr,
 					 (int32) (state->ptr - state->begin));
 			}
 			else if (*(state->ptr) == '\\')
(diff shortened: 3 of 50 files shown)


linwen (approved):
looks good to me.

PR #234, opened 2026-08-18
Fix volatile multiply-referenced CTEs re-evaluated per reference

### LINKED ISSUE #233: Volatile multiply-referenced CTEs are re-evaluated per reference instead of once
### WarehousePG version or build
7.0.0-beta.0 / 7.2.1 (also affects main as of commit 39ff057ec5)

### Expected behavior
A CTE that is referenced more than once and contains a volatile function is
evaluated exactly once for the whole query, no matter how many times it's
referenced -- matching upstream PostgreSQL behavior.

### Actual behavior
With `gp_cte_sharing` at its default (`off`), a multiply-referenced volatile
CTE is planned as an independent copy per reference, so the volatile
expression is re-evaluated once per reference. Query results silently depend
on how many references happen to exist. Setting `gp_cte_sharing = on` masks
the bug by forcing a shared plan.

### Steps to reproduce the behavior

```sql
SELECT count(*) FROM (
  WITH q1(x) AS (SELECT random() FROM generate_series(1, 5))
    SELECT * FROM q1
  UNION
    SELECT * FROM q1
) ss;
```

Expected: `5` (q1 materialized once, both references see the same five rows).
Actual (default GUCs): `10` (random() re-run once per reference).

Root cause: `set_cte_pathlist()` in `src/backend/optimizer/path/allpaths.c`
only forces a shared scan plan for a volatile, multiply-referenced CTE when
`gp_cte_sharing` is on. Upstream's own CTE-inlining rule never inlines a
volatile CTE, specifically to guarantee single evaluation regardless of any
GUC -- this tree's plan-sharing path doesn't honor that guarantee by default.

avamingli:
## Summary
- `set_cte_pathlist()`'s default case only forced a shared scan plan for a
  CTE when `gp_cte_sharing` was on. With it at its default (off), a CTE
  referenced more than once that also contains a volatile function fell
  through to the unshared branch, which plans an independent copy per
  reference -- re-running the volatile expression once per reference instead
  of once for the whole query, so results silently depended on how many
  references happened to exist.
- Force sharing whenever a CTE is both multiply-referenced and volatile,
  regardless of `gp_cte_sharing`, matching the guarantee upstream's own
  CTE-inlining rule already relies on (a volatile CTE is never inlined, so
  it's always evaluated exactly once). The `gp_cte_sharing`-gated condition
  is kept as-is for every other case, so existing sharing behavior is
  unchanged.

Fixes #233

## Test plan
- [x] Added a regression case in `with.sql` covering the repro with
      `gp_cte_sharing` explicitly off; blessed both `with.out` (Postgres
      planner) and `with_optimizer.out` (ORCA).
- [x] Verified the repro returns `5` (not `10`) with `gp_cte_sharing` at its
      default and with it explicitly `on`.
- [x] Verified a singly-referenced volatile CTE is unaffected -- same plan
      as before, no spurious Shared Scan introduced.
- [x] Verified `subselect.sql`'s existing `gp_cte_sharing=on` forced-share
      block (singly-referenced volatile CTE) still blesses the same plan.
- [x] Ran the full dependency-chain prefix of `parallel_schedule` through
      `with`/`subselect` under both `optimizer=off` and `optimizer=on`
      (ORCA) -- all tests pass, zero diffs.

### DIFF
--- a/src/backend/optimizer/path/allpaths.c
+++ b/src/backend/optimizer/path/allpaths.c
@@ -3095,8 +3095,17 @@ set_cte_pathlist(PlannerInfo *root, RelOptInfo *rel, RangeTblEntry *rte)
 			is_shared = true;
 			break;
 		default:
-			/* if plan sharing is enabled and contains volatile functions in the CTE query, also generate a shared scan plan */
-			is_shared =  root->config->gp_cte_sharing && (cte->cterefcount > 1 || contain_volatile_function);
+			/*
+			 * Force sharing when the CTE is both multiply-referenced and
+			 * volatile, regardless of gp_cte_sharing: a volatile CTE with
+			 * more than one reference must be evaluated exactly once, or
+			 * results would depend on how many references there happen to
+			 * be. Otherwise, fall back to the original gp_cte_sharing-gated
+			 * behavior.
+			 */
+			is_shared = (cte->cterefcount > 1 && contain_volatile_function) ||
+				(root->config->gp_cte_sharing &&
+				 (cte->cterefcount > 1 || contain_volatile_function));
 
 	}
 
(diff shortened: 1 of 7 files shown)


avamingli:
Fixed several issues found by @linwen

PR #236, opened 2026-08-19
Fix out-of-bound reads in ascii(), and complete the multibyte-length CVE fix set

adam8157:
## Summary

Backports the upstream fix for CVE-2026-18024 (`ascii()` out-of-bound reads on
invalid multibyte input) together with the four upstream CVE-2026-2006 commits
that the earlier backport did not carry, and the upstream test coverage that all
of them share.

Three of those four are live defects on `main` today:

- EUC_CN's `pg_encoding_max_length()` is 2 while its mb<->wchar routines accept
  3-byte SS2/SS3 sequences, so a round trip through a `max_length * nchars + 1`
  buffer overruns it.  Reproducible as
  `WARNING: problem in alloc set ExprContext: detected write past chunk end`.
- `mb2wchar` produces garbage wchars for input ending in an incomplete
  character.
- `pgp_sym_decrypt()` / `pgp_pub_decrypt()` return `text` without encoding
  validation.  Every other route into the server validates -- the lexer,
  `pg_any_to_server()` for client input, and `COPY` -- so this is a way for
  invalid text to enter a query from plain SQL, and `ascii()` is one of the
  functions that trusts the invariant.

Separately, `SUBSTRING()` on a toasted non-ASCII column has been failing since
the bounds-checking backport landed (upstream bug #19406):

```sql
CREATE TABLE t (c text);
INSERT INTO t VALUES (repeat(U&'\2026', 4000));   -- 4000 3-byte characters
SELECT SUBSTRING(c FROM 1 FOR 1) FROM t;
ERROR:  invalid byte sequence for encoding "UTF8": 0xe2 0x80
```

## Background

Upstream fixed CVE-2026-2006 across seven REL_14 commits between January and
February 2026.  WarehousePG took one of them, plus the GB18030 fix for
CVE-2025-4207.  This PR takes the remaining four, in upstream order, and then
CVE-2026-18024 on top.

The test-coverage commit has to come before the `ascii()` fix, because the
latter's regression tests are written against the `test_bytea_to_text()` helper
it introduces -- and porting it is what surfaced the two mb<->wchar defects
above, since its expected output encodes the fixed behaviour.

Commits are plain upstream cherry-picks from `REL_14_STABLE`; each records its
adaptations for the PG12 base in its own message.  Of note:

- `construct_array_builtin()` / `deconstruct_array_builtin()` are *not*
  backported.  Upstream's hunks automerge here but do not compile, because PG12
  has no `TYPALIGN_*` macros.  The two test-only call sites spell out int4's
  `elmlen`/`elmbyval`/`elmalign` instead, so `arrayfuncs.c` and `array.h` are
  untouched.
- `test_encoding()` returns its report as a row instead of `RAISE NOTICE`, and
  the driving query sorts the rows: the function is evaluated on the segments,
  so upstream's notices would arrive in an arbitrary order.  All 56 reports are
  byte-identical to upstream's notices.
- `SUBSTRING(x SIMILAR pattern ESCAPE escape)` is PG13 syntax; PG12 spells the
  same thing `SUBSTRING(x FROM pattern FOR escape)`.

No catalog, ABI or WAL changes.

## Test plan

- `src/test/regress`: the new `encoding` and `euc_kr` tests pass; full
  `parallel_schedule` is 195/196.  `disable_autovacuum` fails on `main` too
  (A/B confirmed) -- a 129-second timing test, unrelated.
- `contrib/pgcrypto` 25/25, `contrib/pg_trgm` 4/4 (including the new
  `pg_utf8_trgm`), `gpcontrib/orafce` 17/17, `ltree` / `citext` / `unaccent`
  green.
- A/B on an assert-enabled build: the eight `ascii()` cases abort the backend
  before the fix and raise clean errors after, while `ascii('e-acute')`,
  `ascii('rocket')` and `ascii('')` are unchanged; `SUBSTRING()` on a toasted
  3-byte-UTF-8 column errors before and returns the character after; the
  mb<->wchar round-trip reports match upstream byte for byte, and the EUC_CN
  chunk-overrun warning is gone.


linwen (approved):
looks good to me.

PR #238, opened 2026-08-19
fix: don't error in pg_event_trigger_ddl_commands() for objects dropped by the same statement

Bonartze:
## Summary

Two unexpected issues splitting default partitions in WHPG7. Investigation of the report (repro doc + live testing) turned up three claimed defects; after code review, one turned out to already be correct, documented behavior, and the fix set is now:

1. **Event-trigger crash (fixed)** — with a `ddl_command_end` event trigger installed (as extensions like `pgaa` use to track per-relation storage/catalog metadata), `SPLIT DEFAULT PARTITION` (and, as review caught, regular non-default `SPLIT PARTITION` too) on a partition with explicit storage crashed the whole statement with `ERROR: cache lookup failed for relation ...`. This is a latent gap in `pg_event_trigger_ddl_commands()` itself, confirmed via the existing `event_trigger` test's own `ALTER TABLE ... DROP IDENTITY, SET DATA TYPE` case, which hit the identical bug and had a hand-written exception handler silently tolerating it.
2. **Storage inheritance — investigated, not a bug** — the new (non-default) partition of a `SPLIT DEFAULT PARTITION` inherits the storage of the partition being split (the default partition), not the root's. This matches WHPG's own documentation ("partitions created from a `SPLIT PARTITION` inherit the properties of the split child"), WHPG6 precedent, and an existing regression test. An earlier version of this PR changed this to inherit from the root instead; that was reverted after review as a behavior change contradicting documented/tested semantics, not a bug fix.
3. **Naming — investigated, not a bug (see History)** — `SPLIT DEFAULT PARTITION` names the new partition `<root>_1_prt_<name>` (legacy convention) rather than the literal name given. An earlier version of this PR changed this too; CI caught that the legacy name is load-bearing well beyond partition maintenance (see History), so it was also reverted.

A fourth reported symptom — `CREATE TABLE ... PARTITION OF <table> FOR VALUES ...` failing with `"... is not partitioned"` — turned out **not** to be a bug either: the repro pointed the new partition at the default partition itself (a plain leaf table) rather than the root partitioned table. No code change; traced through `DefineRelation()` (`tablecmds.c:1210-1319`) to confirm adding a partition alongside an existing, populated default partition already works correctly for heap/`ao_row`/`ao_column` when the root is named correctly.

**Net effect: this PR now makes exactly one behavioral change** — fixing the `pg_event_trigger_ddl_commands()` crash — plus the regression test that proves it.

## The fix (`tablecmds.c`, `event_trigger.c`)

`AtExecGPSplitPartition()` renames the partition being split to a temp name, then drops that same (now temp-named) relation later in the *same* top-level statement — this happens for both `SPLIT DEFAULT PARTITION` and regular `SPLIT PARTITION`, unconditionally. The rename's `ddl_command_end` entry is collected with a valid address at the time, but `pg_event_trigger_ddl_commands()` only resolves the object's type/identity lazily, when the event trigger function actually runs — by which point the relation is already dropped, so `getObjectTypeDescription()`/`getObjectIdentity()` hit a hard `elog(ERROR, "cache lookup failed for relation %u", ...)` instead of degrading gracefully.

This isn't specific to GPDB's SPLIT machinery — it's a latent gap in `pg_event_trigger_ddl_commands()` itself. The real, complete upstream fix (PostgreSQL commits `2a10fdc4307a` + `2d689babe3cb`) threads a `missing_ok` parameter through `getObjectDescription()`/`getObjectTypeDescription()`/`getObjectIdentityParts()` themselves. I attempted that full backport; it's larger than even initially estimated — this PG12-era tree is missing prerequisite infrastructure upstream's patch assumes exists (`format_procedure_extended`, `format_operator_extended`, `FORMAT_TYPE_INVALID_AS_NULL`), on top of ~1850 lines threading the parameter through three large per-object-class switch statements and ~90 call-site updates tree-wide (`contrib/sepgsql`, `dependency.c`, `pg_depend.c`, `pg_shdepend.c`, `extension.c`, `tablecmds.c`, `regproc.c`). Rather than ship a partial port of this security- and dependency-critical code, this PR instead closes the two concrete, correctness-affecting gaps in a targeted fix:

- **No more duplicated catalog lookup** — the existence check and the schema-name lookup a few lines below now share a single `table_open()`/`get_catalog_object_by_oid()`/`table_close()`, instead of the existence check duplicating that pattern and leaving the schema-lookup's own `elog(ERROR, "cache lookup failed for object %u/%u", ...)` unreachable dead code.

An earlier version of this fix also added an `objectSubId` check (for a dropped column of an otherwise-surviving relation), guarded by a lookup against the `ATTNUM` syscache. Review caught that it wasn't guarding against anything real: `RemoveAttributeById()` only deletes the `pg_attribute` row for system columns; a normal user column drop just sets `attisdropped` and leaves the row in place, so `getObjectIdentity()` already reports it gracefully on its own. Removed rather than keep speculative code with no concrete case behind it.

**Remaining, documented limitation**: object classes outside `is_objectclass_supported()`'s `ObjectProperty[]` (e.g. user mappings) can still hit the same error — closing that fully requires the complete upstream backport, called out explicitly in the code comment as a larger, separate follow-up rather than silently left unmentioned.

**Independent confirmation this is a real, pre-existing, general bug**: the existing `event_trigger` regression test's own `event_trigger_report_end()` function had an `EXCEPTION WHEN SQLSTATE 'XX000' THEN RAISE NOTICE 'END: got internal exception'` handler, written specifically to tolerate this exact crash for `ALTER TABLE ... DROP IDENTITY, SET DATA TYPE int` (an unrelated operation that does its own internal rewrite-via-temp-relation dance). With this fix, that crash no longer happens and the catch branch never fires — review traced its origin to upstream commit `ac78cec9811`, a PG12-only backpatch shim added because the real fix wasn't backpatchable at the time (master/14+ never had it). Removed as dead code rather than left in place, since keeping it would silently swallow the one remaining documented gap above into a bare notice instead of a test failure. `expected/event_trigger.out`/`sql/event_trigger.sql` updated accordingly.

This is the most direct, code-confirmed explanation found so far for the customer's reported `pgaa` failure (`"Get table mapping: Either storage_location and storage_path..."`) — `pgaa` almost certainly hits this same lazy-resolution gap via its own `ddl_command_end` handler.

## Regression test (`event_trigger_gp.sql`)

Folded into the existing `event_trigger_gp` test (which already runs alone on its own schedule line, with a `ddl_command_end` trigger calling `pg_event_trigger_ddl_commands()`) rather than adding a new standalone test + schedule slot. Covers both:

- `SPLIT DEFAULT PARTITION` on a default partition with explicit storage different from the root's;
- regular (non-default) `SPLIT PARTITION` — confirmed this hits the identical crash pre-fix (the rename-then-drop is unconditional, not default-partition-specific), and now completes cleanly, with all 19 expected rows present in `regress_ddl_history` for both split forms.

Generated and verified against a real build via `make installcheck-tests TESTS='event_trigger event_trigger_gp object_address'` — all three pass.

## History — what changed after review

The first version of this PR also (a) made the new non-default partition inherit storage from the **root** instead of the partition being split, and (b) kept its **literal** user-given name instead of the legacy `<root>_1_prt_<name>` convention. Both were reverted after review:

- **Storage**: found to contradict documented WHPG semantics and an existing test (see Summary #2 above) — a deliberate product decision, not a bug, if ever wanted.
- **Naming**: CI (`installcheck-good`, both `optimizer=on`/`optimizer=off`) caught exactly the risk flagged during review — the legacy naming convention is load-bearing well beyond `GpFindTargetPartition`'s by-name lookup. It broke `partition`, `partition1`, `partition_indexing`, `partition_storage`, `partition_locking`, `qp_misc_jiras`, `qp_dropped_cols`, `incremental_analyze`, and the `uao_ddl/alter_ao_part_tables*` tests — all either reference a split-off default partition by its generated legacy name directly, or rely on it being prefixed with `<root>_` for `LIKE`-based catalog queries.

An `ObjectAddress` fix in `tablecmds.c`'s `AT_PartSplit` case was also dropped entirely after review established it never affected `pg_event_trigger_ddl_commands()`'s actual output (that reads `cmd->d.alterTable.objectId`, populated independently) and didn't cover the nested `ALTER TABLE ... ALTER PARTITION ... SPLIT DEFAULT PARTITION` form either.

A second review pass caught two more issues, both fixed: a code comment in `tablecmds_gp.c` had the naming behavior backwards (naming was never inherited from the split partition — `ChoosePartitionName()` always uses the root table's name — only storage/reloptions are inherited from the split partition), and a comment in `event_trigger_gp.sql` misdescribed *why* the test catches the bug (each SQL statement is its own implicit transaction, so a crashing `SPLIT` only rolls back that one statement, not the rest of the script — the test still catches a regression via that statement's own error text and a shorter row count in the final `SELECT`). Both comments corrected; no behavioral changes.


src/backend/commands/tablecmds_gp.c
```diff
@@ -991,8 +992,61 @@ AtExecGPSplitPartition(Relation rel, AlterTableCmd *cmd)
 		elem->colencs = p_colencs;
 		elem->options = p_reloptions;
 
+		elem_new = elem;
+		if (defaultpartname)
+		{
+			/*
+			 * Splitting the DEFAULT partition: the new sibling is a brand
+			 * new range, unrelated to the default partition's own storage
```
adam8157: This is a behaviour change, not a bug fix. The pre-PR behaviour is what WHPG documents, tests and has always done:

- gpdb-doc/markdown/admin_guide/ddl/about-part-changes.html.md:103: "The partitions created from a SPLIT PARTITION inherit the properties of the split child". The default partition is the split child here.
- WHPG 6 (ATPExecPartSplit) copies reloptions / column encodings / orientation from the partition being split for both halves, default included.
- The existing case at src/test/regress/sql/partition.sql:1766-1774 was written to assert exactly this: heap root, `default partition def with(appendonly=true)`, and after the split foo_p_1_prt_p10_20 is expected to be ao_row. This PR flips that expectation to heap.

Concretely, in the common classic layout where the root is heap and every leaf (default included) is declared AO/AOCO with compression, the range split off the default now silently becomes uncompressed heap, while ADD PARTITION on the same table still creates AOCO leaves. The same happens to tablespace: the new half now lands on the parent's tablespace instead of the default partition's, which the PR description doesn't mention.

Are you sure this is the change we want? If so it needs the doc updated and a product-level sign-off, since it contradicts documented semantics.

src/backend/commands/event_trigger.c
```diff
@@ -2109,6 +2109,33 @@ pg_event_trigger_ddl_commands(PG_FUNCTION_ARGS)
...
+					 * this function runs -- e.g. GPDB's SPLIT DEFAULT
+					 * PARTITION renames a partition out of the way and later
+					 * drops that same relation within the same top-level
+					 * statement. getObjectTypeDescription()/getObjectIdentity()
+					 * do a hard catalog lookup and error out for a vanished
+					 * object, so skip such rows instead, the same way the
+					 * IF-NOT-EXISTS/invalid-OID case above is skipped: there
+					 * is nothing coherent left to report about it.
+					 */
+					if (is_objectclass_supported(addr.classId))
```
adam8157, line 2136: The crash is real and pre-existing, but upstream already fixed this exact case (the same DROP IDENTITY + SET DATA TYPE combination) in PG14:

- 2a10fdc4307a "Eliminate cache lookup errors in SQL functions for object addresses" -- adds a missing_ok argument to getObjectIdentity() / getObjectTypeDescription() / getObjectDescription().
- 2d689babe3cb "Improve handling of dropped objects in pg_event_trigger_ddl_commands()" -- uses it here:

```c
identity = getObjectIdentity(&addr, true);
if (identity == NULL)
    continue;
type = getObjectTypeDescription(&addr, true);
```

I'd rather backport those two than carry a hand-rolled probe, because the probe has gaps the upstream version doesn't:

- It only checks (classId, objectId) and ignores objectSubId, so a dropped column of a surviving relation still errors inside getObjectIdentity().
- It is gated on is_objectclass_supported(), so classes outside ObjectProperty[] (e.g. user mappings, which are collected as SCT_Simple) still hit the elog.
- It repeats the table_open/get_catalog_object_by_oid/table_close that the schema-name block ~25 lines below already does for the same object, never frees checktup, and turns that block's `elog(ERROR, "cache lookup failed for object %u/%u")` into dead code.

Note upstream deliberately did not back-patch 2a10fdc because of its size, so this is a bigger diff than the current hunk, but it lands us on the upstream code path and lets expected/event_trigger.out match upstream too.

src/backend/commands/tablecmds.c
```diff
@@ -5908,10 +5908,27 @@ ATExecCmd(List **wqueue, AlteredTableInfo *tab, Relation rel,
 				 (int) cmd->subtype);
 			break;
 
+		case AT_PartSplit:
```
adam8157: This hunk doesn't do what the comment says, and only covers half of the SPLIT forms:

- pg_event_trigger_ddl_commands() builds the ALTER TABLE row from cmd->d.alterTable.objectId and never reads CollectedATSubcmd.address, so there is no "extra, malformed entry" to fix there; the only reader of that field in-tree is src/test/modules/test_ddl_deparse. This is also why the PR's own test output is unchanged by this hunk.
- The nested form `ALTER TABLE t ALTER PARTITION p SPLIT DEFAULT PARTITION ...` is wrapped by gram.y (line ~3979) into an AT_PartAlter whose def->arg holds the AT_PartSplit. It is only unwrapped inside ATExecGPPartCmds() (tablecmds_gp.c:1117), i.e. after this switch has already dispatched on AT_PartAlter. So that form still takes the shared arm and is still collected with InvalidObjectAddress.

If we want the address populated for deparse-style consumers, do it once for all eight partition subtypes in the existing shared arm:

```c
case AT_PartAdd:
case AT_PartDrop:
case AT_PartAlter:
case AT_PartSplit:
case AT_PartRename:
case AT_PartTruncate:
case AT_PartExchange:
case AT_PartSetTemplate:
    ATExecGPPartCmds(rel, cmd);
    ObjectAddressSet(address, RelationRelationId, RelationGetRelid(rel));
    break;
```

and reword the comment accordingly (it's about CollectedATSubcmd consumers, not pg_event_trigger_ddl_commands). Otherwise I'd drop this hunk.

src/test/regress/expected/partition.out
```diff
@@ -2589,8 +2589,8 @@ into (partition p10_20, default partition);
 select c.oid::regclass, relkind, amname, reloptions from pg_class c left join pg_am am on am.oid = relam where c.oid in ('foo_p_1_prt_p10_20'::regclass, 'foo_p_1_prt_def'::regclass);
         oid         | relkind | amname |                            reloptions                             
 --------------------+---------+--------+-------------------------------------------------------------------
- foo_p_1_prt_p10_20 | r       | ao_row | {blocksize=32768,compresslevel=0,compresstype=none,checksum=true}
  foo_p_1_prt_def    | r       | ao_row | {blocksize=32768,compresslevel=0,compresstype=none,checksum=true}
+ foo_p_1_prt_p10_20 | r       | heap   |
```
adam8157: The query at sql/partition.sql:1774 has no ORDER BY, and this re-recorded output now lists foo_p_1_prt_def (the later-created, higher-OID row) before foo_p_1_prt_p10_20, so the row order is coming from pg_class tuple placement. atmsort will paper over it, but the same query in the new split_default_partition.sql already uses `ORDER BY 1::text`; please add the same here and regenerate partition.out / partition_optimizer.out so the expected files stop depending on physical order.

src/test/regress/sql/split_default_partition.sql
```diff
@@ -0,0 +1,109 @@
+--
+-- Tests for ALTER TABLE ... SPLIT DEFAULT PARTITION:
+--  1) the new (non-default) partition inherits storage from the root
+--     table, not from the default partition being split (naming is
+--     intentionally left as the legacy "<root>_1_prt_<name>" pattern --
+--     see the comment in AtExecGPSplitPartition() for why)
+--  2) regular (non-default) SPLIT PARTITION is unaffected by (1): it
+--     keeps inheriting storage from the partition being split
+--  3) SPLIT DEFAULT PARTITION on a default partition with explicit
```
adam8157: Two suggestions on scope:

- Parts 1 and 2 duplicate coverage that partition.sql:1757-1776 and partition_storage.sql already provide. Part 3 is the only new coverage, and event_trigger_gp already runs alone on its own schedule line (greenplum_schedule:94) with a ddl_command_end trigger calling pg_event_trigger_ddl_commands(). Folding part 3 into event_trigger_gp avoids adding another serial slot to the schedule.
- The rename-to-pg_temp_<oid> (tablecmds_gp.c:696) and the DropStmt (tablecmds_gp.c:1094) are unconditional, so a regular `SPLIT PARTITION` under a ddl_command_end trigger hit the same "cache lookup failed" crash. Please add that case too; today nothing in the suite exercises the non-default split with an event trigger installed.

adam8157 (changes requested):
Thanks for the thorough write-up. Of the two code changes that remain, the event-trigger crash is a real pre-existing bug, but upstream already fixed it and I'd rather we backport that (see inline). The storage-inheritance change alters documented SPLIT PARTITION semantics that the existing regression suite asserts, so I'd like an explicit confirmation that this is intended before it goes in (see inline).

Two leftovers from the reverted naming change:

- The PR title and the first commit message still say "naming", but the final diff contains no naming code; naming is byte-identical to main.
- The 14-line NOTE comment in AtExecGPSplitPartition() (tablecmds_gp.c ~1031) describes the literal-name behaviour that e90aa15f53ab reverted, cites a "PR discussion" a future reader can't locate, and sits inside a block that only handles storage. It should go, along with the "see the comment in AtExecGPSplitPartition() for why" sentence in the test header.

Inline comments cover the AT_PartSplit case in ATExecCmd (only covers the top-level form, and pg_event_trigger_ddl_commands() never reads that address), the ORDER BY in partition.sql, and folding the trigger test into event_trigger_gp plus adding a regular SPLIT case.

src/test/regress/expected/partition.out
Bonartze: Moot now — the storage-inheritance change that touched this section is reverted (25c2dc9585), so partition.out/partition_optimizer.out are back to their original, untouched content and this PR no longer modifies that query's output at all. The underlying row-order fragility you flagged is real and your ORDER BY suggestion is the right fix, but it's a pre-existing issue this PR doesn't create or need to fix — happy to file it separately if useful rather than pull it into this diff.

src/backend/commands/tablecmds_gp.c
Bonartze: You're right — this was a behavior change rather than a bug fix. Confirmed and reverted

src/test/regress/sql/split_default_partition.sql
Bonartze: Done both

src/test/regress/expected/partition.out
Bonartze: The storage-inheritance change that touched this section is reverted so partition.out/partition_optimizer.out are back to their original

src/backend/commands/tablecmds.c
Bonartze: You are right on both, dropped entirely

src/backend/commands/event_trigger.c
Bonartze: I closed the two concrete, correctness-affecting gaps you identified instead: objectSubId
is now checked (a dropped column of a surviving relation no longer crashes), and the existence check now shares one table_open/get_catalog_object_by_oid/table_close with the schema-name lookup instead of duplicating it — that lookup's own `elog(ERROR, "cache lookup failed for object %u/%u") `is no longer reachable in the case this check already handles

adam8157 (commented):
Core change looks right. Skipping rows for vanished objects matches what upstream settled on in `2d689babe3cb`, and the new `event_trigger.out` line is byte for byte what master produces. Left a few inline comments.

One non-inline point: the title still says "naming, storage inheritance, and event-trigger reporting", but naming and storage were both reverted, so the net change is just the event trigger fix. Half the 11 commits revert each other too. Suggest squashing on merge with something like "fix: don't error in pg_event_trigger_ddl_commands() for objects dropped by the same statement".

src/test/regress/expected/event_trigger.out
```diff
@@ -467,7 +467,7 @@ NOTICE:  END: command_tag=ALTER SEQUENCE type=sequence identity=evttrig.id_col_d
 NOTICE:  END: command_tag=ALTER TABLE type=table identity=evttrig.id
 ALTER TABLE evttrig.id ALTER COLUMN col_d DROP IDENTITY,
   ALTER COLUMN col_d SET DATA TYPE int;
-NOTICE:  END: got internal exception
+NOTICE:  END: command_tag=ALTER TABLE type=table identity=evttrig.id
```
adam8157, line 468: Since this no longer throws, the `EXCEPTION WHEN SQLSTATE 'XX000'` clause in `event_trigger_report_end()` (`sql/event_trigger.sql:317`) is dead and worth removing here too.

It came from upstream `ac78cec9811`, which backpatched these test cases to PG12 and added the clause only because the fix itself wasn't backpatchable. Master and 14+ don't have it. Keeping it means the remaining gap this PR documents (classes outside `ObjectProperty[]`, e.g. user mappings) would be swallowed into a bare notice instead of failing the test. No expected churn either way.

src/backend/commands/event_trigger.c
```diff
@@ -2109,51 +2109,88 @@ pg_event_trigger_ddl_commands(PG_FUNCTION_ARGS)
...
+									addr.objectId);
+						vanished = !HeapTupleIsValid(objtup);
+
+						/*
+						 * The object's parent row still exists, but if this
+						 * address is for one of its columns (objectSubId), that
+						 * specific column may itself have been dropped since the
+						 * command was collected.
+						 */
+						if (!vanished && addr.objectSubId != 0 &&
```
adam8157: I don't think the dropped-column case this guards against can actually happen. `RemoveAttributeById()` only deletes the `pg_attribute` tuple for system columns (`attnum < 0`); for a user column it just sets `attisdropped`, so the tuple survives and `SearchSysCacheExists2` still returns true. `getObjectIdentity()` then reports `........pg.dropped.N........` without erroring, same as upstream with `missing_ok`.

So this may only be reachable for a system-column address, where it drops a row rather than preventing a crash. Do you have a repro that hits it? If not, maybe drop the block or reword the comment.

Either way, worth gating on `addr.classId == RelationRelationId`. Nothing outside pg_class sets a non-zero subId today, but if something ever does we'd probe `pg_attribute` with a foreign OID and silently drop a valid row.

src/backend/commands/tablecmds_gp.c
```diff
@@ -991,6 +991,23 @@ AtExecGPSplitPartition(Relation rel, AlterTableCmd *cmd)
 		elem->colencs = p_colencs;
 		elem->options = p_reloptions;
 
+		/*
+		 * NOTE: both the new (non-default) half and the recreated default
+		 * half intentionally inherit storage/reloptions/naming from the
```
adam8157: Naming is backwards here. With `partcomp.tablename` NULL, `makePartitionCreateStmt()` calls `ChoosePartitionName(RelationGetRelationName(parentrel), ...)`, so the name comes from the root, not from the child being split. Storage is correct.

src/backend/commands/tablecmds_gp.c
```diff
@@ -991,6 +991,23 @@ AtExecGPSplitPartition(Relation rel, AlterTableCmd *cmd)
...
+		 * DEFAULT partition being split (via the shared "elem" below and
+		 * leaving partcomp.tablename unset for the new half), not from the
+		 * root table, even though the new half covers a brand new range.
+		 * This matches documented WHPG semantics ("partitions created from
+		 * a SPLIT PARTITION inherit the properties of the split child",
+		 * gpdb-doc/markdown/admin_guide/ddl/about-part-changes.html.md),
+		 * WHPG6's ATPExecPartSplit behavior, and the existing regression
+		 * test at partition.sql:1766-1774. An earlier version of this
```
adam8157: Two small things. `partition.sql:1766-1774` will drift, naming the test case reads better. And the "earlier version of this fix ... reverted after review" part won't mean anything to someone reading the file later, that belongs in the commit message.

src/test/regress/sql/event_trigger_gp.sql
```diff
@@ -59,11 +87,16 @@ ALTER TYPE endtest_domain SET DEFAULT ENCODING (compresstype=zlib);
 
 drop event trigger regress_event_trigger_end;
 
--- both the external and the regular table must be here
+-- the external table, the regular table, and both kinds of split must all
+-- be here -- if pg_event_trigger_ddl_commands() had crashed partway
+-- through either SPLIT above, this statement never would have completed
```
adam8157: This isn't quite why the test works. Each regress statement runs in its own implicit transaction, so a failing SPLIT only rolls back that statement; the later CREATEs, this SELECT and the DROPs all still run. The test does catch the bug, via the error text and the shorter row list.

src/backend/commands/event_trigger.c
Bonartze: Removed the objectSubId sub-check entirely

Bonartze:
> Core change looks right. Skipping rows for vanished objects matches what upstream settled on in `2d689babe3cb`, and the new `event_trigger.out` line is byte for byte what master produces. Left a few inline comments.
> 
> One non-inline point: the title still says "naming, storage inheritance, and event-trigger reporting", but naming and storage were both reverted, so the net change is just the event trigger fix. Half the 11 commits revert each other too. Suggest squashing on merge with something like "fix: don't error in pg_event_trigger_ddl_commands() for objects dropped by the same statement".

Thank you for this round of review, I have addressed all suggessted issues. whenever you have time, please, take another  look.

src/backend/commands/event_trigger.c
```diff
@@ -2109,51 +2109,82 @@ pg_event_trigger_ddl_commands(PG_FUNCTION_ARGS)
...
-					identity = getObjectIdentity(&addr);
-
 					/*
-					 * Obtain schema name, if any ("pg_temp" if a temp
-					 * object). If the object class is not in the supported
-					 * list here, we assume it's a schema-less object type,
-					 * and thus "schema" remains set to NULL.
+					 * A command collected earlier in this transaction may
+					 * reference an object that no longer exists by the time this
+					 * function runs -- e.g. GPDB's SPLIT DEFAULT PARTITION
```
adam8157: Nit: please say WHPG rather than GPDB in new comments.

src/backend/commands/tablecmds_gp.c
```diff
@@ -991,6 +991,25 @@ AtExecGPSplitPartition(Relation rel, AlterTableCmd *cmd)
 		elem->colencs = p_colencs;
 		elem->options = p_reloptions;
 
+		/*
+		 * NOTE: both the new (non-default) half and the recreated default
+		 * half intentionally inherit storage/reloptions from the DEFAULT
```
adam8157: Nit: the shared `elem` path is taken for regular splits too, so this reads better as "inherit storage/reloptions from the partition being split (default or not)".

src/backend/commands/tablecmds_gp.c
```diff
@@ -991,6 +991,25 @@ AtExecGPSplitPartition(Relation rel, AlterTableCmd *cmd)
...
+		 * partition being split (via the shared "elem" below), not from
+		 * the root table, even though the new half covers a brand new
+		 * range. This matches documented WHPG semantics ("partitions
+		 * created from a SPLIT PARTITION inherit the properties of the
+		 * split child", gpdb-doc/markdown/admin_guide/ddl/about-part-changes.html.md),
```
adam8157: Nit: please drop the gpdb-doc path here. The docs are maintained in a separate repo, so the in-tree copy and this path can go stale; the quoted sentence stands on its own.

adam8157 (dismissed):
All points from both rounds are addressed and CI is green on the squashed commit. Approving.

Two small wording nits inline on the new comments, fine to fold in before merge.

adam8157 (approved):

PR #239, opened 2026-08-20
[WHPG6]pgcrypto: reset debug handler after PGP pipeline errors

linwen:
After CVE-2026-14663, cipher initialization failures can ereport() from inside the PGP pipeline.  With debug=1, that could bypass the regular px_set_debug_handler(NULL) cleanup and leave the global pgcrypto debug handler installed in the backend, making later non-debug calls emit "dbg: ..." NOTICEs.

Wrap the encrypt and decrypt pipelines in PG_TRY()/PG_CATCH() to reset the handler before rethrowing, and keep the normal-path reset after PG_END_TRY() since PG_FINALLY() is not available on the PG 9.4 base.

Add a regression test that verifies a failed debug call does not leak debug NOTICEs into a subsequent non-debug call.

Reported-by: Zhenglong Li <zhenglong.li@enterprisedb.com>
(cherry picked from commit af4b3f3b4eafc239abdc75cc5f467ca5de5bff2e)


SKeyerror (approved):
LGTM.

PR #240, opened 2026-08-20
[CVE-2026-6464] psql: skip in-line COPY ... FROM STDIN data after a failure

SKeyerror:
## Summary

Backport of the fix for **CVE-2026-6464** (CVSS 8.1): if a `COPY ... FROM STDIN` command failed before the server sent `PGRES_COPY_IN`, psql did not consume the in-line data that follows and executed that data as SQL — a SQL-injection vector from a hostile server or a crafted plain-text dump.

Fixes https://enterprisedb.atlassian.net/browse/PTT-1571

## Backported commits (in order)

1. **Partial e717a9a18b2** ("SQL-standard function body") — only the `psqlscan.l` / `psqlscan_int.h` lexer parts. Adds the `begin_depth` BEGIN...END tracking machinery that PG14 has and PG12 lacks; prerequisite so the CVE fix applies cleanly and we stay textually aligned with upstream. The server-side BEGIN ATOMIC feature itself is NOT backported.
2. **029c5ac03db** ("psql: Refine lexing of BEGIN...END blocks in CREATE FUNCTION statements") — whole commit, plus the upstream follow-up `if` → `else if` correction so the file matches REL_14_STABLE final state.
3. **7215c7e9643** ("Save/restore more lexer state when skipping text due to \if") — applied clean, zero conflicts.
4. **5c51ae4556f** ("Teach psql to skip in-line COPY ... FROM STDIN data after a failure") — the main fix. With the prerequisites in place the `psqlscan.l` conflicts disappeared; remaining conflicts were resolved against WHPG's PG12-era psql (include ordering in `common.c`, `PQputCopyEnd` call site in `copy.c`) and WHPG's diverged test scripts (`copy2.sql`, `copyselect.sql`, `generated.sql`).
5. **WHPG-specific test adjustments** — psql now consumes in-line data up to `\.` even when the COPY command fails, so every `COPY FROM STDIN` needs a `\.` terminator:
   - `gpcopy.source`: added `\.` after three GPDB-specific COPYs expected to fail before COPY_IN (invalid NEWLINE option, `ON SEGMENT` with STDIN, copying to a catalog table).
   - `psql` test: the upstream lexer test uses `BEGIN ATOMIC`, which the PG12-based server rejects; the test is kept (it still verifies psql sends the whole construct as one statement across `\if`/`\endif`) with the expected output adjusted to the single server-side syntax error.

## Verification

- The five CVE-relevant lexer functions (`psqlscan_record_initial_keyword`, `psqlscan_is_create_routine`, `psqlscan_is_copy_from_stdin`, `psqlscan_track_identifier`, `psql_scan_count_copy_from_stdin`) are byte-identical with current REL_14_STABLE.
- Manual checks against a demo cluster: injected SQL between a failed COPY and `\.` is discarded; `\copy` failure skips its data; multiple COPYs in one query string via `\;` are counted and consumed correctly; an unexpected `PGRES_COPY_IN` from the server now aborts the connection.
- Swept all pg_regress-style test inputs (`sql/*.sql`, `input/*.source`) repo-wide for `COPY FROM STDIN` without a `\.` terminator; the three `gpcopy.source` sites above were the only real hits (isolation2 talks libpq directly and is unaffected; pgbench usage matches upstream and needs no change).
- Regression tests covering the touched scripts pass against a demo cluster: `copy`, `copy2`, `copyselect`, `generated`, `privileges`, `rowsecurity`, `psql`, `gpcopy` all ok in schedule context.


SKeyerror:
Superseded by #241 (same commits, opened from the contributor fork as intended).

PR #241, opened 2026-08-20
[CVE-2026-6464] psql: skip in-line COPY ... FROM STDIN data after a failure

SKeyerror:
## Summary

Backport of the fix for **CVE-2026-6464** (CVSS 8.1): if a `COPY ... FROM STDIN` command failed before the server sent `PGRES_COPY_IN`, psql did not consume the in-line data that follows and executed that data as SQL — a SQL-injection vector from a hostile server or a crafted plain-text dump.

Fixes https://enterprisedb.atlassian.net/browse/PTT-1571
(Supersedes #240, which was opened from a branch pushed directly to this repo by mistake.)

## Backported commits (in order)

Each commit is a `git cherry-pick -x` of the upstream PostgreSQL commit: the original upstream message, author, and date are preserved, a `(cherry picked from commit ...)` line records the upstream hash, and any conflict resolution is documented in a `WarehousePG adaptation:` note in the commit message.

1. **e717a9a18b2** ("SQL-standard function body") — partial pick, `psqlscan.l` / `psqlscan_int.h` lexer parts only. Adds the `identifier_count`/`begin_depth` BEGIN...END tracking machinery that PG14 has and PG12 lacks; prerequisite so the CVE fix applies cleanly and we stay textually aligned with upstream. The server-side BEGIN ATOMIC feature itself is NOT backported.
2. **029c5ac03db** ("psql: Refine lexing of BEGIN...END blocks in CREATE FUNCTION statements") — whole commit, plus the folded upstream follow-up **d9a9f4b4b92a** ("psql: Fix line continuation prompts for unbalanced parentheses"), the `if` → `else if` correction, so the file matches REL_14_STABLE final state.
3. **7215c7e9643** ("Save/restore more lexer state when skipping text due to \if") — clean pick, zero conflicts, no adaptation needed.
4. **5c51ae4556f** ("Teach psql to skip in-line COPY ... FROM STDIN data after a failure") — the main fix. Conflicts were resolved against WHPG's PG12-era psql (include ordering in `common.c`, `PQputCopyEnd` call site in `copy.c`) and WHPG's diverged test scripts (`copy2.sql`, `copyselect.sql`, `generated.sql`). The WHPG-specific test adjustments are folded into this commit so it passes tests on its own:
   - `gpcopy.source`: added `\.` after three GPDB-specific COPYs expected to fail before COPY_IN (invalid NEWLINE option, `ON SEGMENT` with STDIN, copying to a catalog table) — psql now consumes in-line data up to `\.` even when the COPY command fails.
   - `psql` test: the upstream lexer test uses `BEGIN ATOMIC`, which the PG12-based server rejects; the test is kept (it still verifies psql sends the whole construct as one statement across `\if`/`\endif`) with the expected output adjusted to the single server-side syntax error.

## Verification

- The five CVE-relevant lexer functions (`psqlscan_record_initial_keyword`, `psqlscan_is_create_routine`, `psqlscan_is_copy_from_stdin`, `psqlscan_track_identifier`, `psql_scan_count_copy_from_stdin`) are byte-identical with current REL_14_STABLE.
- Manual checks against a demo cluster: injected SQL between a failed COPY and `\.` is discarded; `\copy` failure skips its data; multiple COPYs in one query string via `\;` are counted and consumed correctly; an unexpected `PGRES_COPY_IN` from the server now aborts the connection.
- Swept all pg_regress-style test inputs (`sql/*.sql`, `input/*.source`) repo-wide for `COPY FROM STDIN` without a `\.` terminator; the three `gpcopy.source` sites above were the only real hits (isolation2 talks libpq directly and is unaffected; pgbench usage matches upstream and needs no change).
- Regression tests covering the touched scripts pass against a demo cluster: `copy`, `copy2`, `copyselect`, `generated`, `privileges`, `rowsecurity`, `psql`, `gpcopy` all ok in schedule context.


SKeyerror:
CI: https://github.com/EnterpriseDB/warehouse-pg-packaging/actions/runs/32327161314

linwen (changes requested):
The CVE-2026-6464 upstream commits are all incorporated correctly and conflicts are resolved. Two small things before merge:

1. Please use git cherry-pick -x for the upstream commits. That keeps each commit's original message and adds a (cherry picked from commit <sha>) line, so we retain the upstream traceability. For the ones that weren't a clean pick, consider add a short WarehousePG adaptation: note on the main conflicts resolved (common.c include order, the copy.c PQputCopyEnd call site, the diverged test scripts) would help future readers.

2. Please fold commit 5 into commit 4, and drop regression.diffs / regression.out. Commit 5 has fixed cases that commit 4's behavior change introduces, so keeping them separate leaves commit 4 failing on its own — folding them in keeps each commit self-consistent.

SKeyerror:
Sorry, I didn't notice that, please take a look again, thanks.

linwen (approved):
Thanks, looks good to me.

PR #242, opened 2026-08-21
[6X] Cross-check the type of a portal running EXECUTE or FETCH

avamingli:
## Summary
- CVE-2026-16239 (CVSS 8.8): `FillPortalStore()` never verified that the outer portal's tuple descriptor agreed with the actual output of the query it ran into the tuplestore, which upstream says can be leveraged to disclose server memory / achieve arbitrary code execution.
- Backports the prerequisite (REL_14_STABLE `2f48ede080f`) that lets `SetTuplestoreDestReceiverParams()` take a target tupdesc + failure message, and the fix itself (REL_14_STABLE `1f49beef27`, applied byte-identical to `pquery.c`).
- WHPG6 predates the pluggable `TupleTableSlot` ops introduced later upstream, so the tuple-conversion path in `tstoreReceiver.c` is adapted onto `HeapTuple` + `do_convert_tuple()` instead of the newer slot-based `execute_attr_map_slot()`.
- Also updates the sibling call site in `portalcmds.c` (`PersistHoldablePortal`, the WITH HOLD cursor path) for the new signature.
- Checked the other GPDB-forked portal paths (`_SPI_pquery`, dispatch-side portals): none of them bypass `FillPortalStore()`; parallel retrieve cursors don't exist on this branch.
- Adds a dedicated regression test (`bfv_cve_2026_16239`) exercising both call sites' ordinary usage.

## Test plan
- [x] `bfv_cve_2026_16239` passes under both `optimizer=off` and `optimizer=on` (ORCA)
- [x] `returning`, `with`, `portals`, `prepare`, `insert`, `update`, `delete`, `rpt_returning` all still pass (broad pre-existing coverage of the touched `FillPortalStore()`/`PersistHoldablePortal()` paths)
- [x] Diffed the adapted files against the `REL_14_STABLE` tip to confirm every remaining difference is a pre-existing WHPG6/PG9.4 API divergence, not an error

### DIFF
--- a/.abi-check/6.27.5-WHPG/postgres.symbols.ignore
+++ b/.abi-check/6.27.5-WHPG/postgres.symbols.ignore
@@ -1,0 +1,1 @@
+SetTuplestoreDestReceiverParams
--- a/src/backend/commands/portalcmds.c
+++ b/src/backend/commands/portalcmds.c
@@ -493,7 +493,9 @@ PersistHoldablePortal(Portal portal)
 		SetTuplestoreDestReceiverParams(queryDesc->dest,
 										portal->holdStore,
 										portal->holdContext,
-										true);
+										true,
+										NULL,
+										NULL);
 
 		/* Fetch the result set into the tuplestore */
 		ExecutorRun(queryDesc, ForwardScanDirection, 0);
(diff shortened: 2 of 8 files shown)


SKeyerror:
Same as 7X, ad99227 and ad99227 needs to be folded.

avamingli:
> Same as 7X, [ad99227](https://github.com/warehouse-pg/warehouse-pg/commit/ad99227966108ac7bdf0621d492c380704703ba5) and [ad99227](https://github.com/warehouse-pg/warehouse-pg/commit/ad99227966108ac7bdf0621d492c380704703ba5) needs to be folded.

done

SKeyerror (approved):
LGTM.

PR #243, opened 2026-08-21
Cross-check the type of a portal running EXECUTE or FETCH

avamingli:
## Summary
- CVE-2026-16239 (CVSS 8.8): `FillPortalStore()` never verified that the outer portal's tuple descriptor agreed with the actual output of the query it ran into the tuplestore. Upstream says this can be leveraged to disclose server memory / achieve arbitrary code execution.
- Backports the prerequisite (REL_14_STABLE `2f48ede080f`) that lets `SetTuplestoreDestReceiverParams()` take a target tupdesc + failure message, and the fix itself (REL_14_STABLE `1f49beef27`), applied to `pquery.c`'s `FillPortalStore()`.
- Updates the two other WHPG7-only callers of `SetTuplestoreDestReceiverParams()` (`PersistHoldablePortal` in `portalcmds.c`, and the ANALYZE sample-rows dispatcher in `analyze.c`) for the new signature, passing `NULL, NULL` to match upstream's own non-security callers.
- Checked GPDB's other portal paths: RETRIEVE (parallel retrieve cursor) goes through `UtilityTupleDescriptor()` → the same `FillPortalStore()` code path, so it's covered without any separate change.
- Adds a regression test to `prepare.sql` reproducing the vulnerability class via a PL/pgSQL cursor `OPEN ... FOR EXECUTE` of a prepared statement that's deallocated and re-prepared with a mismatched type before the cursor is fetched.
- See also the sibling WHPG6 backport in #242.

## Test plan
- [x] Reproduced the vulnerability live on the unpatched build: fetching a `text` value through a portal that expects `int4` returns the value's raw bytes reinterpreted as an integer. Confirmed the patched build raises `query result type does not match portal result type` instead.
- [x] New `prepare` test passes cleanly (both `prepare.out` and the ORCA-enabled `prepare_optimizer.out`) through the real `pg_regress` harness.
- [x] Manually exercised every `FillPortalStore()`-driven strategy (EXECUTE, FETCH, modifying-CTE) to confirm no behavior change on non-mismatched input.
- [x] Diffed the adapted files against the `REL_14_STABLE` tip to confirm every remaining difference is a pre-existing WHPG7/PG12 API divergence, not an error.

### DIFF
--- a/.abi-check/7.1.0/postgres.symbols.ignore
+++ b/.abi-check/7.1.0/postgres.symbols.ignore
@@ -13,3 +13,4 @@ MainLWLockNames
 choose_setop_type
 gp_resgroup_print_operator_memory_limits
 pqFunctionCall3
+SetTuplestoreDestReceiverParams
--- a/src/backend/commands/analyze.c
+++ b/src/backend/commands/analyze.c
@@ -2313,7 +2313,9 @@ static QueryDesc *build_querydesc(Portal portal, char *sql)
 	SetTuplestoreDestReceiverParams(destReceiver,
 									portal->holdStore,
 									portal->holdContext,
-									false);
+									false,
+									NULL,
+									NULL);
 
 	/*
 	 * Parse the SQL string into a list of raw parse trees.
--- a/src/backend/commands/portalcmds.c
+++ b/src/backend/commands/portalcmds.c
@@ -533,7 +533,9 @@ PersistHoldablePortal(Portal portal)
 		SetTuplestoreDestReceiverParams(queryDesc->dest,
 										portal->holdStore,
 										portal->holdContext,
-										true);
+										true,
+										NULL,
+										NULL);
 
 		/* Fetch the result set into the tuplestore */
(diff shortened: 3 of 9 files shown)


SKeyerror:
This cherry-pick is pretty clean, just a minor suggestion:

1cc80c4d should be amended into 1c5a452e to keep every commit self-contained and buildable, since we will choice "rebase and merge" not "squash and merge"

avamingli:
Folded it in, but split it: the ABI-ignore entry went into 1c5a452e (the commit that actually changes SetTuplestoreDestReceiverParams()'s signature), and the regression test went into fd5fd80ed4 (the actual fix) instead — the test exercises behavior that only exists after the fix, so putting it in 1c5a452e alone would've left that commit with a test that can't pass yet. Down to 2 commits now, each buildable and self-contained.

SKeyerror (approved):
LGTM.

avamingli:
> LGTM.

Thanks for review!

PR #245, opened 2026-08-22
[WHPG6] pgcrypto: Fix buffer overflow in pgp_pub_decrypt_bytea()

linwen:
pgp_pub_decrypt_bytea() was missing a safeguard for the session key length read from the message data, that can be given in input of pgp_pub_decrypt_bytea().  This can result in the possibility of a buffer overflow for the session key data, when the length specified is longer than PGP_MAX_KEY, which is the maximum size of the buffer where the session data is copied to.

A script able to rebuild the message and key data that can trigger the overflow is included in this commit, based on some contents provided by the reporter, heavily editted by me.  A SQL test is added, based on the data generated by the script.

Reported-by: Team Xint Code as part of zeroday.cloud
Author: Michael Paquier <michael@paquier.xyz>
Reviewed-by: Noah Misch <noah@leadboat.com>
Security: CVE-2026-2005
Backpatch-through: 14
(cherry picked from commit 01de2e32df7b5b46720f64068a2674c3bd28ae3f)

WarehousePG adaptation:
Renumbered the new PXE_PGP_KEY_TOO_BIG error code from -111 to -124: WHPG6 (PostgreSQL 9.4 base) still carries PXE_PGP_RSA_UNSUPPORTED at -111, a 9.4-era pgcrypto error code that upstream removed in later versions, and -112..-123 are likewise in use, so -124 is the next free slot. The code fix, error string, and pgp-pubkey-session regression test applied otherwise unchanged.


avamingli (approved):
LGTM.

PR #246, opened 2026-08-23
[WHPG6] Require superuser to install non-built-in selectivity estimators

linwen:
In WHPG6/PostgreSQL 9.4, CREATE OPERATOR is the only path that can attach restriction and join selectivity estimators to an operator. Reject non-superusers when they specify a non-built-in estimator function for either RESTRICT or JOIN.  Built-in estimators keep the existing EXECUTE privilege check.

Also harden tsmatchsel() so it verifies that the variable side is a tsvector before using tsvector statistics.  If the estimator is attached to an unexpected operator, fall back to the default selectivity instead.

Only the portions applicable to the WHPG6 code base are included: network_selfuncs remains unchanged because WHPG6's network estimators are constant-returning stubs; intarray _int_matchsel and the related pg_extension syscache changes do not exist in this branch; and ALTER OPERATOR cannot change estimator functions in WHPG6.

Add regression coverage for the CREATE OPERATOR superuser gate for both restriction and join estimators.

Based on upstream commit ea3bf34986032ba2760538dacbc144b1f6e300fd (CVE-2026-2004), by Tom Lane. Because the surrounding 9.4 code has diverged from upstream, this was reimplemented rather than cherry-picked directly.

Reported-by: Daniel Firer as part of zeroday.cloud
Security: CVE-2026-2004

- [ ✅] Add tests for the change
- [ ✅ ] Pass `make installcheck`

### DIFF
--- a/src/backend/commands/operatorcmds.c
+++ b/src/backend/commands/operatorcmds.c
@@ -250,11 +250,32 @@ DefineOperator(List *names, List *parameters)
 					 errmsg("restriction estimator function %s must return type \"float8\"",
 							NameListToString(restrictionName))));
 
-		/* Require EXECUTE rights for the estimator */
-		aclresult = pg_proc_aclcheck(restrictionOid, GetUserId(), ACL_EXECUTE);
-		if (aclresult != ACLCHECK_OK)
-			aclcheck_error(aclresult, ACL_KIND_PROC,
-						   NameListToString(restrictionName));
+		/*
+		 * If the estimator is not a built-in function, require superuser
+		 * privilege to install it.  This protects against using something
+		 * that is not a restriction estimator or has hard-wired assumptions
+		 * about what data types it is working with.  (Built-in estimators are
+		 * required to defend themselves adequately against unexpected data
+		 * type choices, but it seems impractical to expect that of
+		 * extensions' estimators.)
+		 *
+		 * If it is built-in, only require EXECUTE rights.
+		 */
+		if (restrictionOid >= FirstNormalObjectId)
+		{
+			if (!superuser())
+				ereport(ERROR,
+						(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+						 errmsg("must be superuser to specify a non-built-in restriction estimator function")));
+		}
+		else
+		{
+			/* Require EXECUTE rights for the estimator */
+			aclresult = pg_proc_aclcheck(restrictionOid, GetUserId(), ACL_EXECUTE);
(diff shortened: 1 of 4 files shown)


huiliang-liu (approved):

avamingli:
The built-in boundary differs from upstream. ea3bf349860 gates on FirstGenbkiObjectId (10000), but this PR uses FirstNormalObjectId (16384). WHPG6 doesn't have FirstGenbkiObjectId, but FirstBootstrapObjectId (10000) in src/include/catalog/pg_magic_oid.h is the same boundary.

With 16384 as the line, any function created during initdb (OID 10000-16383, e.g. information_schema helpers, cdb_schema/system_views functions) counts as built-in and skips the superuser gate, getting only the EXECUTE check. Upstream deliberately drew the line at 10000 so only bootstrap pg_proc entries qualify, since those are the only ones held to the "must defend itself" standard. I don't see an initdb-created function matching the estimator signature today, so this is likely not exploitable, but it's a needless weakening, and the merged main fix (#219) uses FirstGenbkiObjectId.

Suggest changing both the RESTRICT and JOIN checks to >= FirstBootstrapObjectId, maybe with a comment noting it's the 6X equivalent of upstream's FirstGenbkiObjectId. All real built-in estimators have OIDs well below 10000, and the new regression tests pass unchanged since the test functions get OIDs >= 16384 either way.

linwen:
> The built-in boundary differs from upstream. ea3bf349860 gates on FirstGenbkiObjectId (10000), but this PR uses FirstNormalObjectId (16384). WHPG6 doesn't have FirstGenbkiObjectId, but FirstBootstrapObjectId (10000) in src/include/catalog/pg_magic_oid.h is the same boundary.
> 
> With 16384 as the line, any function created during initdb (OID 10000-16383, e.g. information_schema helpers, cdb_schema/system_views functions) counts as built-in and skips the superuser gate, getting only the EXECUTE check. Upstream deliberately drew the line at 10000 so only bootstrap pg_proc entries qualify, since those are the only ones held to the "must defend itself" standard. I don't see an initdb-created function matching the estimator signature today, so this is likely not exploitable, but it's a needless weakening, and the merged main fix (#219) uses FirstGenbkiObjectId.
> 
> Suggest changing both the RESTRICT and JOIN checks to >= FirstBootstrapObjectId, maybe with a comment noting it's the 6X equivalent of upstream's FirstGenbkiObjectId. All real built-in estimators have OIDs well below 10000, and the new regression tests pass unchanged since the test functions get OIDs >= 16384 either way.

thanks, will check it.

PR #247, opened 2026-08-24
Fix reversed coordinator/segment CPUSET split in resource groups

SKeyerror:
`getCpuSetByRole()` gave the coordinator the part after ';' and segments the part before it, the reverse of the documented CPUSET='<coordinator_cores>;<segment_cores>' syntax, since the refactoring in 8588bc98803. 

Swap the branches back so validation and cgroup cpuset binding apply to the right hosts.

### DIFF
--- a/src/backend/commands/resgroupcmds.c
+++ b/src/backend/commands/resgroupcmds.c
@@ -1753,17 +1753,15 @@ getCpuSetByRole(const char *cpuset)
 		splitcpuset = (char *)cpuset;
 	else
 	{
-		char *scpu = first + 1;
-
 		/* Get result cpuset by IS_QUERY_DISPATCHER(), on coordinator or segment */
 		if (IS_QUERY_DISPATCHER())
-			splitcpuset = scpu;
-		else
 		{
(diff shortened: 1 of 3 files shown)


linwen (commented):
[pr247-resgroupcmds-unittest.patch](https://github.com/user-attachments/files/31378735/pr247-resgroupcmds-unittest.patch)

The code change to resgroupcms.c looks good to me.
I suggest we add a unit test to cover this in case it was wrongly modified again. 
I have attached a patch, please take a look.
It can be covered by "make unittest-check"

SKeyerror:
> [pr247-resgroupcmds-unittest.patch](https://github.com/user-attachments/files/31378735/pr247-resgroupcmds-unittest.patch)
> 
> The code change to resgroupcms.c looks good to me. I suggest we add a unit test to cover this in case it was wrongly modified again. I have attached a patch, please take a look. It can be covered by "make unittest-check"

Thanks for this!

No need to overcomplicate it, though. If you think adding a unit test is necessary, feel free to modify the original commit directly. I'm completely fine with that.

linwen (approved):
thanks, looks good to me.

PR #248, opened 2026-08-25
Resolve dispatch addresses from a topology file for DR read replica

JunfengYang:
Add whpg_dispatch_topology_file (PGC_SIGHUP, default empty). When set on
a dispatcher, the component table is built from the file's rows —
exactly one primary row per content, coordinator row included — so
dispatch, interconnect peer lists, and QE options all resolve addresses
from the file instead of gp_segment_configuration row selection. With
the GUC empty (the default) behavior is unchanged everywhere.

Audits, all ERROR-not-fallback:
- The file's content set must equal the catalog's; a missing or extra
  content fails the query with a named error, never a silent fall back
  to the possibly stale catalog.
- dbid handshake: while the topology is active, gpqeid carries the
  expected dbid as an optional seventh field and the QE refuses the
  connection pre-auth when its own gp_dbid differs, turning a wrong or
  swapped topology entry into an immediate error instead of silently
  wrong results. Gang creation classifies this failure as permanent and
  surfaces it without burning the gang-creation retries.
- An unparsable or invalid file fails closed; the read-only
  whpg_dispatch_topology_state GUC reports inactive/active/error.

The cached component table is rebuilt at transaction start whenever the
topology configuration changes, tracked by a content-hash signature of
the file (file mtimes come from the kernel's coarse clock — 10ms at
HZ=100 — so a same-size same-tick swap would be invisible to a
stat-based signature).

Covered by the new isolation2 test dispatch_topology.

Reclassify it into WAL_RECOVERY_TARGET alongside recovery_target_name:
pausing replay at a named restore point is a supported operation for
read replicas, not a developer knob. Visibility flags and behavior are
unchanged.

- [ ] Document changes


src/backend/cdb/dispatcher/cdbgang.c
```diff
@@ -570,6 +616,27 @@ cdbgang_parse_gpqeid_params(struct Port *port pg_attribute_unused(),
...
+	 * wrong data directory; refuse it instead of running queries as the
+	 * wrong segment.
+	 */
+	if (cp && np)
+	{
+		if (gpqeid_next_param(&cp, &np))
+		{
+			int			expected_dbid = (int) strtol(cp, NULL, 10);
+
+			if (expected_dbid != GpIdentity.dbid)
```
huiliang-liu, line 709: It is better to check contentID too.  The file's content<->dbid pairing and dbid uniqueness are never validated, so a consistent-but-misassigned row silently dispatches one content's work to another  content's segment.

src/backend/cdb/cdbdispatchtopology.c
```diff
@@ -0,0 +1,367 @@
...
+				   &content, &dbid, hostname, address, &port, datadir,
+				   extra) != 6)
+		{
+			snprintf(errbuf, errbufsz,
+					 "line %d: expected 6 fields \"content dbid hostname address port datadir\"",
+					 lineno);
+			goto fail;
+		}
+
+		if (content < -1)
```
huiliang-liu: content has no upper-bound check before the (int16) cast, and the duplicate check compares the raw int against already-truncated stored values, so out-of-range contents alias real ones and bypass every validation.

src/backend/cdb/cdbutil.c
```diff
@@ -333,6 +334,101 @@ readGpSegConfigFromCatalog(int *total_dbs)
...
+
+	configs = (GpSegConfigEntry *)
+		palloc0(topology->nentries * sizeof(GpSegConfigEntry));
+
+	for (j = 0; j < topology->nentries; j++)
+	{
+		DispatchTopologyEntry *entry = &topology->entries[j];
+		GpSegConfigEntry *config = &configs[j];
+
+		config->dbid = entry->dbid;
```
huiliang-liu, line 474: applyDispatchTopology substitutes the file's (replica) dbids into component rows, but the FTS shared-memory status array stays keyed by catalog dbids, so all dispatch-path health checks read the wrong slot.

src/backend/cdb/cdbdispatchtopology.c
```diff
@@ -0,0 +1,367 @@
...
+ * demand so a fresh session (e.g. a monitoring connection that never
+ * dispatched) still reports the truth.
+ */
+const char *
+show_whpg_dispatch_topology_state(void)
+{
+	if (!dispatch_topology_enabled())
+		return "inactive";
+
+	if (topology_state == DISPATCH_TOPOLOGY_UNKNOWN)
```
huiliang-liu: show_whpg_dispatch_topology_state caches ACTIVE/ERROR in a per-backend global forever after the first evaluation, so sessions that never dispatch — the monitoring consumers this GUC exists for — report stale state indefinitely.

src/backend/cdb/cdbutil.c
```diff
@@ -570,6 +678,9 @@ getCdbComponentInfo(void)
 
 	hash_destroy(hostPrimaryCountHash);
 
+	/* remember which topology configuration this table was built with */
+	component_databases->topology_signature = dispatch_topology_signature();
```
huiliang-liu: The topology signature is computed from a second, independent read of the file rather than from the bytes dispatch_topology_load parsed, and the 'unreadable' sentinel matches itself — both defeat the rebuild-on-change mechanism.

src/backend/cdb/dispatcher/cdbgang_async.c
```diff
@@ -157,7 +158,9 @@ cdbgang_createGang_async(List *segments, SegmentType segmentType)
 									 segdbDesc->isWriter,
 									 segdbDesc->identifier,
 									 segdbDesc->segment_database_info->hostPrimaryCount,
-									 totalSegs * 2);
+									 totalSegs * 2,
+									 dispatch_topology_enabled() ?
```
huiliang-liu: dispatch_topology_enabled() is re-evaluated live at gang creation instead of being captured with the component table, so a SIGHUP disabling the GUC mid-transaction dispatches to file-derived addresses without the dbid handshake.

src/backend/cdb/dispatcher/cdbgang.c
JunfengYang: Fixed in d4145bd9ff, with one deliberate deviation from the suggestion. The handshake now carries the expected CONTENT next to the dbid and the QE refuses on either mismatch — that closes the consistent-but-misassigned case (content A with content B's dbid+address passes the dbid check by construction). dbid uniqueness within the file is now enforced at parse.

What I did NOT add is validating the file's content↔dbid pairing against the catalog: after a source-side failover the restored node keeps the dbid of the basebackup it came from (internal.auto.conf), while the replayed catalog may already name the promoted mirror's dbid as that content's primary — the two legally diverge in a configuration the replica explicitly supports. Content identity is the invariant that survives failovers, so the runtime content check is the enforcement point instead. There's a comment in applyDispatchTopology recording this.

New isolation2 case: a file with contents 0/1 carrying each other's dbid AND address — dbid handshake passes, content handshake refuses.

src/backend/cdb/cdbdispatchtopology.c
JunfengYang: Fixed in d4145bd9ff: content is bounded at PG_INT16_MAX before the cast (symmetric with the dbid check), so the duplicate check now compares values that are all in-range — no truncation aliasing, and the catalog cross-check can no longer be reached with a truncated value. isolation2 grows an oversized-content case (65537 refused at parse with "invalid content", instead of aliasing content 1).

src/backend/cdb/cdbutil.c
JunfengYang: Fixed in d4145bd9ff, but not by re-keying: FtsIsSegmentDown() now returns false for rows of a topology-built component table (detected via the table's signature) instead of indexing ftsProbeInfo->status[] with the file's dbid. Rationale: the FTS array describes the catalog world (probe populates it under catalog dbids, source addresses); a topology-built table lives outside that world, and on the hot-standby dispatcher this feature serves, no FTS probe runs at all — health is arbitrated by the connection attempts themselves. Enforcing file-dbid==catalog-dbid instead would have broken the legal post-failover divergence described in the thread above, so the array is bypassed rather than re-aligned.

src/backend/cdb/cdbdispatchtopology.c
JunfengYang: Fixed in d4145bd9ff: the show_hook re-validates on every call — the per-backend ACTIVE/ERROR cache is gone (the file is a handful of lines; a parse per SHOW is nothing against a monitoring cadence). One subtlety kept: a file that parses but failed the catalog cross-check during a component-table build keeps reporting "error" until its bytes change — the show_hook can't re-run the cross-check (no catalog access), so dispatch_topology_set_error() now records the offending file's signature and the verdict sticks exactly as long as that file does. isolation2 covers the monitoring shape: the same utility session reads error, the file is fixed behind it, the next read says active — no dispatch in between.

src/backend/cdb/cdbutil.c
JunfengYang: Fixed in d4145bd9ff: dispatch_topology_parse() now accumulates the signature from the very bytes it reads (same rolling hash, same byte stream) and returns it on the DispatchTopology; the component table stores that, so a file swap between parse and table build can no longer stamp old entries with a new file's identity. And an unreadable current file never matches any stored signature — the comparison forces the rebuild, which then reports the real error instead of silently keeping a stale table (the signature function's comment always promised that; now the code does it).

src/backend/cdb/dispatcher/cdbgang_async.c
JunfengYang: Fixed in d4145bd9ff: whether gang creation demands the handshake is now a property of the component table the addresses came from — its stored topology signature (empty = catalog-built) — rather than the GUC value at that instant. A SIGHUP disabling the feature between table build and gang creation can no longer dispatch to file-derived addresses with the handshake disarmed; the GUC change takes effect at the next table rebuild, where addresses and handshake flip together.

src/backend/cdb/cdbfts.c
```diff
@@ -138,6 +138,22 @@ FtsIsSegmentDown(CdbComponentDatabaseInfo *dBInfo)
...
+	 * some other segment's slot, or garbage past the used range.  The
+	 * file declares its members dispatchable; their health is arbitrated
+	 * by the connection attempts themselves.  (On the hot-standby
+	 * dispatcher this feature serves, no FTS probe runs at all.)
+	 * Threadsafe: reads only memory owned by the component table.
+	 */
+	if (dBInfo->cdbs != NULL &&
+		dBInfo->cdbs->topology_signature != NULL &&
+		dBInfo->cdbs->topology_signature[0] != '\0')
+		return false;
```
huiliang-liu, line 157: **FTS is disabled wholesale for topology-built component tables, and nothing restricts the GUC to hot-standby dispatchers.**

Returning false here unconditionally turns off every FTS consumer on the dispatch path: the pre-gang-creation down check (`cdbgang_async.c:106`), mid-dispatch `checkSegmentAlive` (`cdbdisp_async.c:981`), the idle-QE recycling gate (`cdbutil.c:1011`), and `FtsTestSegmentDBIsDown` via `cdbfts.c:176`. The comment's justification ("no FTS probe runs on the hot-standby dispatcher") is sound *there*, but `whpg_dispatch_topology_file` is PGC_SIGHUP and nothing prevents enabling it on an ordinary live primary QD — where a file generated from the current catalog passes the dbid/content handshake everywhere.

On such a QD: a segment host that powers off without sending RST no longer breaks the dispatch, so the query hangs until TCP keepalive timeout (potentially hours) instead of failing within an FTS probe interval; idle QEs to the dead host are recycled instead of destroyed; and after an FTS mirror promotion, new gangs keep connecting to the deposed-but-still-alive old primary (same dbid/content, handshake passes), committing writes that are lost when the file is later repointed at the promoted mirror.

Suggest refusing to activate the feature where an FTS probe actually runs (e.g. error in `applyDispatchTopology` unless the QD is a hot-standby dispatcher), or at minimum guarding/documenting the GUC accordingly.

src/backend/cdb/cdbutil.c
```diff
@@ -680,7 +813,8 @@ cdbcomponent_updateCdbComponents(void)
 			cdb_component_dbs->expand_version = GetGpExpandVersion();
 		}
 		else if ((cdb_component_dbs->fts_version != ftsVersion ||
-				 cdb_component_dbs->expand_version != expandVersion))
+				 cdb_component_dbs->expand_version != expandVersion ||
+				 !dispatch_topology_signature_matches(cdb_component_dbs->topology_signature)))
 		{
 			if (TempNamespaceOidIsValid())
```
huiliang-liu, line 904: ** A session holding temp tables detects the topology-signature mismatch but silently keeps dispatching to the stale topology.**

On a signature mismatch this branch does nothing — no rebuild, no warning — so a session with a valid temp namespace keeps using the old file's addresses for its entire lifetime, while re-reading and re-hashing the file at every transaction start without ever converging.

Concretely: a session creates a temp table while file F1 is active; the operator swaps in F2 redirecting contents to new DR addresses (or removes the file expecting fail-closed behavior). Every subsequent transaction in that session detects the mismatch, skips the rebuild, and dispatches to F1's stale addresses — the gpqeid handshake passes because the stale table's dbids match the stale segments' own identities. That defeats the feature's explicit no-silent-fallback guarantee, and `whpg_dispatch_topology_state` reports `active` the whole time.

For a topology change (as opposed to an fts_version bump) an ERROR — or at least a one-time WARNING — seems better than silent stale routing; alternatively remember the mismatch and rebuild as soon as the temp namespace is gone.

src/backend/cdb/cdbutil.c
```diff
@@ -333,6 +334,113 @@ readGpSegConfigFromCatalog(int *total_dbs)
...
+	/*
+	 * Note what is deliberately NOT validated here: that a row's dbid
+	 * matches the catalog's dbid for that content.  After a source-side
+	 * failover the restored node keeps the dbid of the basebackup it came
+	 * from (internal.auto.conf), while the replayed catalog may already
+	 * name the promoted mirror's dbid as the content's primary — the two
+	 * legally diverge.  Misassignment within the file (content A carrying
+	 * content B's dbid and address) is caught at connection time by the
+	 * content half of the gpqeid handshake instead.
+	 */
```
huiliang-liu, line 465: **The entry-database identity check contradicts this invariant for the coordinator row.**

This comment correctly defers dbid validation to the content half of the gpqeid handshake, because file dbids and the replica's own identity legally diverge after a source-side failover — but the unchanged check at `cdbutil.c:616-628` still requires the content=-1 row to carry `dbid == GpIdentity.dbid`. The deferral exists for segments but not for the coordinator row.

Scenario: the DR replica QD runs with local gp_dbid=1 (kept from its basebackup). The source cluster fails over its coordinator to the standby (dbid 8); the replicated catalog now shows content -1, role='p', dbid 8. The operator regenerates the topology file from `gp_segment_configuration` where role='p' — the same recipe this PR's test uses — producing a coordinator row with dbid 8. Every component-table build then fails with the internal-looking "cannot locate entry database", while `SHOW whpg_dispatch_topology_state` still reports `active`.

When the table is topology-built, the entry-db check should match on segindex only (or be relaxed equivalently).

src/backend/cdb/cdbdispatchtopology.c
```diff
@@ -0,0 +1,406 @@
...
+
+		/*
+		 * Accumulate the signature from the very bytes being parsed, so
+		 * the identity stored with the resulting table can never describe
+		 * a different file than the one the entries came from.  Same
+		 * function as dispatch_topology_signature() over the same byte
+		 * stream.
+		 */
+		for (k = 0; k < linelen; k++)
+			hash = hash * 33 + (unsigned char) line[k];
```
huiliang-liu: **The two signature computations hash different byte streams; any NUL byte in the file makes them permanently disagree.**

The stored signature is accumulated here over `fgets` chunks measured with `strlen()`, while `dispatch_topology_signature()` — recomputed at every transaction start — hashes raw `fread` bytes. These are two hand-rolled djb2 loops (plus a duplicated `psprintf("%s|%llu|%llu", ...)` format) whose byte-identity is enforced only by a comment.

If a tooling accident leaves an embedded NUL in a comment line (`# note\0junk\n`), the file parses fine and queries succeed, but the stored signature omits everything after the NUL. Every transaction start then sees a mismatch and silently destroys and rebuilds the component table — dropping all cached gangs and idle QEs, paying full gang re-creation per transaction, forever, with no error anywhere and the state GUC showing `active`.

Suggest one shared byte-accumulating helper used by both paths (or `hash_bytes()` from common/hashfn.h, or the pg_crc32c API already used elsewhere in cdb code) so the two can never drift.

src/backend/cdb/cdbdispatchtopology.c
```diff
@@ -0,0 +1,406 @@
...
+	char		errbuf[TOPOLOGY_MAX_LINE];
+
+	if (!dispatch_topology_enabled())
+		return "inactive";
+
+	dispatch_topology_resolve_path(path, sizeof(path));
+	if (dispatch_topology_parse(path, errbuf, sizeof(errbuf)) == NULL)
+		return "error";
+
+	if (crosscheck_error_signature != NULL &&
```
huiliang-liu: **This branch is dead code: no backend that can execute SHOW can ever have `crosscheck_error_signature` set.**

The latch is a per-backend static and its only writers are the crosscheck failure paths in `applyDispatchTopology` (`cdbutil.c:375`, `cdbutil.c:399`), which run only in GP_ROLE_DISPATCH sessions. During a crosscheck outage those sessions abort every transaction at transaction start (component-table rebuild → ERROR) before any SHOW can execute; and utility sessions — the only ones that still work, and per the test's own comment the intended monitoring path — never run `applyDispatchTopology`, so their latch is always NULL.

Net effect: with a file that parses cleanly but, say, omits content 0, every dispatch session fails every query, yet `whpg_dispatch_topology_state` polled from a utility session reports `active` — the exact failure this GUC exists to surface reports healthy. (The only reachable way for this branch to fire is the stale-latch false positive described in the `dispatch_topology_set_error` comment.)

Suggest recording the failed signature in shared memory (the `ftsProbeInfo` pattern) so any backend's show_hook can see it.

src/backend/cdb/cdbdispatchtopology.c
```diff
@@ -0,0 +1,406 @@
...
+		for (k = 0; k < linelen; k++)
+			hash = hash * 33 + (unsigned char) line[k];
+		total += linelen;
+
+		while (*p == ' ' || *p == '\t')
+			p++;
+		if (*p == '\0' || *p == '\n' || *p == '#')
+			continue;
+
+		if (sscanf(p, "%d %d %255s %255s %d %1023s %7s",
```
huiliang-liu: **Hostnames the parser accepts here are rejected later with an internal error.**

The `%255s` widths (and the TOPOLOGY_MAX_LINE comment above, which explicitly claims wider-than-MAXHOSTNAMELEN support) accept hostname/address up to 255 bytes, but `getCdbComponentInfo()` (`cdbutil.c:505`) still `elog(ERROR)`s on any hostname longer than MAXHOSTNAMELEN (64 on Linux) with the internal-looking "Invalid length (%d) of hostname" message. The catalog path enforced that limit at insertion time; the file path re-establishes it nowhere.

So a topology file listing a legitimate 80-char FQDN parses fine, `SHOW whpg_dispatch_topology_state` reports `active`, and every dispatching transaction dies with the internal error — monitoring sees a healthy feature during a total dispatch outage.

Validate the length at parse time (rejecting the file with a proper `errbuf` message) so the state GUC and the load error agree.

src/backend/cdb/cdbdispatchtopology.c
```diff
@@ -0,0 +1,406 @@
...
+/*
+ * Called by the component-table build when the file parsed but failed a
+ * cross-check against the catalog.  The show_hook cannot re-run the
+ * cross-check (no catalog access there), so remember which file bytes
+ * failed: status keeps reporting "error" until the file changes.
+ */
+void
+dispatch_topology_set_error(void)
+{
+	char	   *sig = dispatch_topology_signature();
```
huiliang-liu: **The error latch is stamped from a re-read of the file and is never cleared on a later successful build.**

(a) The signature is computed here from a *second read* of the file — violating this module's own "never from a second look at the file" rule (see the `dispatch_topology_load` comment). If the crosscheck fails on file A and the operator swaps in corrected file B before this read happens, the error is stamped with B's bytes, so this backend reports `error` for the good file until its bytes change again.

(b) Nothing clears `crosscheck_error_signature` when the same bytes later pass the crosscheck — e.g. the standby replays a catalog change (segment addition) that makes the untouched file valid: dispatch works, but this backend's state GUC reports `error` indefinitely.

Suggest passing the failing topology's already-computed `topology->signature` into `dispatch_topology_set_error()`, and clearing the latch when `applyDispatchTopology` succeeds.

huiliang-liu (commented):
Re-review of the updated branch. Three high-severity and four medium-severity findings below as inline comments. High: the blanket FTS bypass for topology-built tables (cdbfts.c), the temp-namespace branch that silently keeps stale routing (cdbutil.c), and the entry-database dbid check that breaks after a source-side coordinator failover (cdbutil.c). Medium: the two divergent signature hash implementations, the dead crosscheck-error branch in the show_hook, hostnames accepted at parse time but rejected at table-build time, and the error latch being stamped from a re-read and never cleared.

src/backend/cdb/cdbfts.c
JunfengYang: Fixed in a1c00dc312 with the hard refusal you suggested: applyDispatchTopology errors outright unless IS_HOT_STANDBY_QD(), naming the exit (whpg-dr promote's retirement removes the GUC before any node reaches normal operation), the state GUC reports error for that configuration, and the GUC description documents the envelope. All three harms close at the root — on a primary the file can never build a component table, so FTS stays fully authoritative there; the FtsIsSegmentDown bypass survives only for the one window it still covers (a session's topology table outliving an out-of-band promotion while the fresh probe fills the array), and its comment now says exactly that.

Testability drove the follow-through in c75f3e158b: the positive suite moves to the hot_standby schedule (-1S sessions on the standby coordinator — the real deployment shape, which the old primary-QD test never exercised), and the old location keeps a deployment-envelope negative suite: established sessions error, new connections are refused at connect time, the state GUC reads error, and everything recovers when the GUC is removed.

src/backend/cdb/cdbutil.c
JunfengYang: Fixed in a1c00dc312: a topology-signature mismatch with a valid temp namespace now fails closed — ERROR naming the way out (reconnect / drop the temp tables) — instead of silently keeping the stale table; fts_version/expand_version mismatches keep the historical skip. With the hard refusal in place this corner is normally unreachable (a hot-standby session cannot create temp tables; a primary session can never own a topology-built table), but a session can outlive an out-of-band pg_promote — and in exactly that corner the skip used to short-circuit the rebuild that would have refused, so the fail-closed branch stays. Covered in the negative suite: a session creates a temp table, the GUC appears, its next query errors with the reconnect hint, and the temp table survives the episode.

src/backend/cdb/cdbutil.c
JunfengYang: Fixed in ddebaddd4f, but by validating rather than relaxing — the entry-db locate check is untouched. Relaxing it to segindex-only would only defer the failure: the coordinator row's dbid flows into the entry-gang gpqeid handshake, so a row carrying the replayed catalog's post-failover identity would FATAL mid-query on exactly the queries that use an entry slice — intermittent instead of deterministic, and still internal-looking.

The file's contract is that every row carries the node's OWN identity (gp_dbid from internal.auto.conf — what whpg-dr's harvest emits; the coordinator row it produces is always the local dispatcher). For content=-1 that identity is verifiable synchronously at load, so applyDispatchTopology now refuses a mismatched coordinator row with both dbids named and a hint stating the contract. The catalog-recipe in the old test — the scenario's likely source — now carries a warning that it only holds without failover history, and the hot-standby suite has the negative case (coordinator row with another node's dbid → load refusal).

src/backend/cdb/cdbdispatchtopology.c
JunfengYang: Fixed in 33d0b29ebc by deleting the content hash rather than unifying the two loops. Stepping back, no product flow ever changes the file under a live dispatcher (enable writes it before nodes start, step never touches it, re-enable/retirement run with nodes stopped), so hashing the whole file at every transaction start was oversized for the need — and the dual-loop drift you found grew out of exactly that machinery. The signature is now the file's stat identity (dev, inode, size, mtime, ctime) through one shared formatter: the stored side is fstat'ed from the very descriptor the parse read (atomically consistent with the parsed bytes — the parse-vs-stamp race is gone by construction), and the compare side is open+fstat with no content read. Every legitimate change moves the tuple (rename → inode; overwrite → size/mtime/ctime); forged-timestamp in-place rewrites are declared out of scope in the comment. NUL bytes stop being a signature concern — inside a data row they already fail the field count.

src/backend/cdb/cdbdispatchtopology.c
JunfengYang: You're right that it was dead code — fixed in fe8b559b8e by removing the latch instead of promoting it to shared memory. Looking at consumers: whpg-dr never reads this GUC (it probes with live queries — enable's dispatch smoke, status's per-node SQL), and a crosscheck failure is the loudest state the system has: the build fails closed and every dispatching query carries the real message. There was no quiet observer left to serve, so the honest resolution was deleting the subsystem and stating the GUC's scope explicitly: inactive / active (feature applies here and the file is well-formed) / error (parse failure, or enabled outside hot standby). The show_hook's comment documents that catalog-level failures surface through the queries themselves. If a genuine SHOW-based consumer materializes later, the shmem design you sketched is the right shape for it.

src/backend/cdb/cdbdispatchtopology.c
JunfengYang: Fixed in ddebaddd4f: the parse now enforces MAXHOSTNAMELEN for both hostname and address with a message naming the limit, so the state GUC and the load error agree. The buffers deliberately stay at 256 — a tighter sscanf width would silently truncate an overlong FQDN into a different, plausible-looking name; reading it in full and refusing is the safer shape — and the TOPOLOGY_MAX_LINE comment now says that instead of advertising wider-than-MAXHOSTNAMELEN support. Covered in the hot-standby suite with a 70+ char hostname variant. (Worth noting: Kubernetes-style FQDNs exceed 64 routinely, so the component-table limit itself may deserve a kernel-wide revisit some day — but that's a both-gates-together change, out of scope here; this fix guarantees there's a single validation to relax when it happens.)

src/backend/cdb/cdbdispatchtopology.c
JunfengYang: Both halves are moot as of fe8b559b8e — the latch is removed entirely (see the thread above), so there is no set_error to stamp from a re-read and no stale latch to clear. For the record you were right on both counts: the re-read violated the module's own single-read rule from the very same commit that introduced it (the correct value, topology->signature, was already in the caller's hand), and no clear path existed. Had the latch survived, the fix would have been exactly what you outlined — stamp from the parsed signature, clear on successful build.

huiliang-liu (dismissed):
LGTM

JunfengYang:
Follow-up 1bf2234 on the standby demand (b7b9bb8): the check keyed on standby.signal alone, but a whpg-dr replica runs TARGETED recovery — recovery.signal, held paused at its restore point for the replica's whole life — so on the feature's primary deployment shape every dispatch died with "dispatch topology standby mismatch" (verified live: enable's dispatch smoke failed on a healthy replica; with the fix the same cluster reaches SERVING and fans out to all contents).

The impostor the check rejects is a live primary of the source cluster, which carries NEITHER signal file, so accepting either recovery entrance (standby.signal or recovery.signal) keeps the defense intact while admitting both legitimate shapes. Still a plain stat(), usable from ProcessStartupPacket. The end-of-recovery corner is covered by ordering: promote retires the topology file before any node leaves recovery, so a node that lost recovery.signal is no longer a topology-dispatch target.

huiliang-liu (approved):

PR #249, opened 2026-08-25
Fix CVE-2026-2006 and CVE-2025-4207 (multibyte length over-reads)

adam8157:
Backport the two multibyte-length over-read CVEs to `WHPG_6X_STABLE`.

## Summary

- **CVE-2026-2006** (CVSS 8.8) — `pg_mblen()` has no bounds check, so a corrupted
  string makes every loop that iterates with it overrun its buffer. Upstream
  replaced all core callers with `pg_mblen_cstr()`, `pg_mblen_with_len()`,
  `pg_mblen_range()` or `pg_mblen_unbounded()`; this branch had ~135 unbounded
  call sites. The full upstream REL_14 series is taken, including the EUC_CN
  length fix (a real heap overrun), the `mb2wchar` short-input fix, the pgcrypto
  change that stops `pgp_sym_decrypt()` handing back unvalidated text, and the
  `SUBSTRING()` fix for bug #19406 that the main patch would otherwise regress.
- **CVE-2025-4207** (CVSS 5.9) — with GB18030, `pg_gb18030_mblen()` reads
  `*(s + 1)` before anything has established that a second byte exists.
  Reachable from SQL via `convert()`/`convert_from()`, and from libpq via
  `PQescapeLiteral()`, `PQescapeIdentifier()`, `PQescapeStringConn()` and
  `PQescapeString()`.
- `gpcontrib/orafce` is a vendored fork of several of the same string functions
  and carried the same pattern; it is hardened to match.

## Background

Both CVEs were fixed on `main` earlier (#231 and #236); these are the 6X
counterparts. The port is against PostgreSQL 9.4, so it starts from the
PG12-adapted commits on `main` rather than upstream REL_13/REL_14, whose
`wchar.c` has already moved to `src/common/`. Each commit message records what
had to change for 9.4; the notable ones are `ts_locale.h`'s
`USE_WIDE_UPPER_LOWER` conditional (removed upstream in PG10), `ISOPERATOR`
still living in `tsvector_parser.c`, and the many upstream hunks that only
exist because a later release started printing the whole offending character
instead of its first byte — those files have no `pg_mblen()` call here at all
and are left untouched.

One commit is not a plain backport of the CVE fixes: "Fix insufficiently-paranoid
GB18030 encoding verifier" is an upstream master-only commit from 2015 that was
never back-patched to 9.4. Every branch the CVE-2025-4207 fix was back-patched
to already contains it, and without it `pg_gb18030_verifier()` performs the very
over-read the CVE is about.

The coverage tests (`encoding`, `euc_kr`, `pg_utf8_trgm`) are introduced already
adapted to this branch's psql, which has no `\if` (PG10+); the skip guards are
replaced by comments and the tests simply run, since gpinitsystem/gpdemo create
the cluster as UTF-8.

## Test plan

- [x] `installcheck-good` on a gpdemo cluster (failures limited to the
      pre-existing ORCA-disabled / role-name diffs, identical to a
      `WHPG_6X_STABLE` baseline run)
- [x] `contrib`: pgcrypto, pg_trgm, ltree, unaccent, citext, hstore, dict_xsyn
- [x] `gpcontrib/orafce` (12/12)
- [x] `conversion` / `encoding` / `euc_kr` / `pg_utf8_trgm` green; regenerated
      6X expected output matches the PG12 port byte-for-byte apart from
      documented version differences
- [x] GB18030 `convert_from()` rejects truncated input instead of reading past it


linwen (approved):
looks good to me.

PR #250, opened 2026-08-25
[WHPG6] Use FirstBootstrapObjectId for estimator boundary

linwen:
WHPG6 does not define upstream's FirstGenbkiObjectId, but FirstBootstrapObjectId is the equivalent 10000 boundary.  Use it for the RESTRICT and JOIN estimator checks so initdb-created functions with OIDs 10000-16383 do not skip the superuser gate.

This is a defense-in-depth follow-up to
f4c19e561d8d902bfdc9a80a697a4b02f4810b4f.

Existing create_operator tests still cover the user-created estimator path because their helper functions receive normal OIDs.

Suggested-by: Zhang Mingli <avamingli@gmail.com>

Security: CVE-2026-2004

### DIFF
--- a/src/backend/commands/operatorcmds.c
+++ b/src/backend/commands/operatorcmds.c
@@ -260,8 +260,14 @@ DefineOperator(List *names, List *parameters)
 		 * extensions' estimators.)
 		 *
 		 * If it is built-in, only require EXECUTE rights.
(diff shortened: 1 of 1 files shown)


avamingli (approved):
LGTM.

PR #251, opened 2026-08-25
Fix ORCA crash in extended-stats estimation after incremental group stats

JunfengYang:
When a memo group's cached statistics lack columns needed by a later
derivation request (e.g. a ROLLUP expanded into a CTE whose consumers
request different column sets), the missing columns are derived
separately and merged via CStatistics::AppendStats(). That merge copied
histograms and widths only, so the colid -> attno mapping (added for
extended statistics) diverged from the histogram map: a column could
have a histogram but no attno entry.

ApplyCorrelatedStatsToScaleFactorFilterCalculation() then dereferenced
the result of the attno lookup without a null check and crashed the
coordinator with SIGSEGV whenever the table had any extended statistics
object and a filter referenced one of the appended columns (e.g.
TPC-DS Q67 after CREATE STATISTICS (ndistinct) on item/date_dim,
PTT-1682). This is the same class of bug as the CTE remapping fix in
upstream commit https://github.com/warehouse-pg/warehouse-pg/commit/124a048e8c6b561a0bdaba1c529ff26cdebfb746 (gpdb#16212), which repaired
CopyStatsWithRemap() but not AppendStats().

Fix both layers:

* CStatistics::AppendStats() now also merges the colid -> attno mapping
  (copy-on-write, since the mapping is shared by refcount across stats
  objects, see ScaleStats()) and adopts the input's extended-stats info
  when none is set, restoring the invariant that every column with a
  histogram has an attno entry.

* CExtendedStatsProcessor now null-checks the attno and histogram
  lookups and skips clauses it cannot translate, falling back to the
  independence assumption, matching what
  ApplyCorrelatedStatsToNDistinctCalculation() already does. Also bail
  out gracefully if no ndistinct item matches the attribute set.

* CExtendedStatsProcessor also skips system columns (attno <= 0):
  extended statistics never cover them, and feeding a negative attno
  into CBitSet::ExchangeSet() would convert it to a huge unsigned bit
  index. The backend rejects these clauses too
  (dependency_is_compatible_clause()).

Add a regression test with the minimized reproduction: a ROLLUP over a
join where the dimension table has ndistinct extended statistics and
the filter column is appended to the scan group's stats after the
initial derivation.

- [ ] Document changes
- [x] Review a PR in return to support the community

### DIFF
--- a/src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
+++ b/src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
@@ -34,7 +34,29 @@ using namespace gpopt;
 static BOOL
 IsDependencyCapablePredicate(CStatsPred *child_pred GPOS_UNUSED)
 {
+	/*
+	 * TODO: EsptPoint covers all constant comparisons (Eq/G/GEq/L/LEq, see
+	 * CStatsPredUtils::StatsCmpType()), so range predicates are let into
+	 * functional-dependency estimation where they get equality semantics.
+	 * The backend analogue, dependency_is_compatible_clause(), accepts
+	 * equality-to-pseudoconstant only; this should gate on the point
+	 * predicate's comparison type being EstatscmptEq.
+	 */
 	return child_pred->GetPredStatsType() == CStatsPred::EsptPoint;
+}
+
+/*
+ * A colid -> attno mapping entry is usable for extended-statistics estimation
+ * only if it exists and refers to a user column. Unlike the backend, ORCA has
+ * to translate its column ids back to attnos and the mapping is not
+ * guaranteed to cover every column. System columns (attno <= 0) are never
+ * covered by extended statistics and cannot be represented in a CBitSet; the
+ * backend rejects them too (dependency_is_compatible_clause()).
+ */
+static BOOL
+FUsableAttno(const INT *attnum)
+{
+	return nullptr != attnum && 0 < *attnum;
 }
 
 /*
@@ -277,6 +299,14 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
 		{
 			ULONG colid = child_pred->GetColId();
 			INT *attnum = colid_to_attno_mapping->Find(&colid);
+			if (!FUsableAttno(attnum))
+			{
+				/*
+				 * Skip such clauses; they fall back to the independence
+				 * assumption.
+				 */
+				continue;
+			}
 			clauses_attnums->ExchangeSet(*attnum);
 		}
 	}
@@ -353,6 +383,12 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
 
 			ULONG colid = child_pred->GetColId();
 			INT *attnum = colid_to_attno_mapping->Find(&colid);
+			if (!FUsableAttno(attnum))
+			{
+				/* no usable attno for this clause; it was never collected
+				 * into clauses_attnums by the pre-processing loop above */
+				continue;
+			}
 
 			/*
 			 * Technically we could find more than one clause for a given
@@ -365,7 +401,24 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
 			 */
 			if (dependency_implies_attribute(dependency, *attnum))
 			{
-				s2 = 1 / result_histograms->Find(&colid)->GetFrequency().Get();
+				const CHistogram *histogram = result_histograms->Find(&colid);
+				if (nullptr == histogram)
+				{
+					/*
+					 * No histogram to estimate the implied clause with; give
+					 * up on the dependency for this attribute. The attnum
+					 * bit must still be cleared: the outer loop terminates
+					 * only once no dependency is fully matched by the
+					 * remaining attnums, so leaving the bit set would make
+					 * find_strongest_dependency() return the same dependency
+					 * forever. The clause stays unestimated here and is later
+					 * given the default selectivity by the regular per-column
+					 * path (see MakeHistHashMapConjFilter()).
+					 */
+					clauses_attnums->ExchangeClear(*attnum);
+					continue;
+				}
+				s2 = 1 / histogram->GetFrequency().Get();
 
 				/* mark this one as done, so we don't touch it again. */
 				child_pred->SetEstimated();
@@ -428,10 +481,21 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToNDistinctCalculation(
 		ULONG colid = *(*src_grouping_cols)[ul];
 
 		INT *attnum = colid_to_attno_mapping->Find(&colid);
-		if (!attnum)
+		if (nullptr == attnum)
 		{
+			/* no colid -> attno mapping; extended stats are unusable */
 			attnums->Release();
 			return false;
+		}
+		if (0 >= *attnum)
+		{
+			/*
+			 * System column: extended statistics never cover it and CBitSet
+			 * cannot represent it. Skip just this column; it never enters
+			 * 'attnums', so it stays unmatched below and is kept for the
+			 * regular per-column ndistinct path.
+			 */
+			continue;
 		}
 		attnums->ExchangeSet(*attnum);
 	}
@@ -512,6 +576,14 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToNDistinctCalculation(
 				item = tmpitem;
 				break;
 			}
+		}
+
+		if (nullptr == item)
+		{
+			/* there should be an item for every attribute combination */
+			matched->Release();
+			attnums->Release();
+			return false;
 		}
 
 		/* Form the output varinfo list, keeping only unmatched ones */
--- a/src/backend/gporca/libnaucrates/src/statistics/CStatistics.cpp
+++ b/src/backend/gporca/libnaucrates/src/statistics/CStatistics.cpp
@@ -552,6 +552,33 @@ CStatistics::AppendStats(CMemoryPool *mp, IStatistics *input_stats)
 	CStatisticsUtils::AddWidthInfo(mp, stats->m_colid_width_mapping,
 								   m_colid_width_mapping);
 	GPOS_CHECK_ABORT;
+
+	// Also merge the colid -> attno mapping so that every column that has a
+	// histogram keeps an attno entry; extended-stats estimation relies on
+	// this invariant. The mapping may be shared with other stats objects
+	// (see ScaleStats()), so merge copy-on-write instead of mutating it.
+	if (0 < stats->m_colid_to_attno_mapping->Size())
+	{
+		UlongToIntMap *attnos_new = GPOS_NEW(mp) UlongToIntMap(mp);
+		AddAttnoInfo(mp, m_colid_to_attno_mapping, attnos_new);
+		AddAttnoInfo(mp, stats->m_colid_to_attno_mapping, attnos_new);
+		m_colid_to_attno_mapping->Release();
+		m_colid_to_attno_mapping = attnos_new;
+	}
+
+	// Attnos are only meaningful relative to m_ext_stats' base relation, so
+	// merging is only valid for stats objects describing the same relation;
+	// all callers merge stats of the same memo group / expression today.
+	GPOS_ASSERT(nullptr == m_ext_stats || nullptr == stats->m_ext_stats ||
+				m_ext_stats == stats->m_ext_stats);
+
+	if (nullptr == m_ext_stats)
+	{
+		// non-owning pointer, kept alive by the MD cache (never released by
+		// ~CStatistics and never AddRef'd)
+		m_ext_stats = stats->m_ext_stats;
+	}
+	GPOS_CHECK_ABORT;
 }
 
 // copy statistics object
@@ -764,6 +791,27 @@ CStatistics::AddAttnoInfoWithRemap(CMemoryPool *mp, UlongToIntMap *src_attno,
 		{
 			colid = new_colref->Id();
 		}
+
+		if (nullptr == dest_attno->Find(&colid))
+		{
+			const INT *attno = col_attno_map_iterator.Value();
+			INT *attno_copy = GPOS_NEW(mp) INT(*attno);
+			BOOL result GPOS_ASSERTS_ONLY =
+				dest_attno->Insert(GPOS_NEW(mp) ULONG(colid), attno_copy);
+			GPOS_ASSERT(result);
+		}
+	}
+}
+
+// add attno information without remapping
+void
+CStatistics::AddAttnoInfo(CMemoryPool *mp, UlongToIntMap *src_attno,
+						  UlongToIntMap *dest_attno)
+{
+	UlongToIntMapIter col_attno_map_iterator(src_attno);
+	while (col_attno_map_iterator.Advance())
+	{
+		ULONG colid = *(col_attno_map_iterator.Key());
 
 		if (nullptr == dest_attno->Find(&colid))
 		{
--- a/src/test/regress/sql/gporca.sql
+++ b/src/test/regress/sql/gporca.sql
@@ -3713,6 +3713,29 @@ create table cte_test(a int);
 insert into cte_test select i from generate_series(1,10)i;
 analyze cte_test;
 explain (analyze, costs off, summary off, timing off) with cte as (select * from cte_test) select * from cte union all select * from cte;
+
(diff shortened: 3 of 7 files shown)


src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
```diff
@@ -428,8 +466,13 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToNDistinctCalculation(
...
+		if (nullptr == attnum || 0 >= *attnum)
 		{
+			/*
+			 * No usable attno for this grouping column (mapping missing, or
+			 * a system column, which extended statistics never cover and
+			 * CBitSet cannot represent); fall back to per-column ndistinct.
+			 */
 			attnums->Release();
 			return false;
 		}
```
linwen: **Blocking — behavior regression for grouping lists that contain a system column** (e.g. `GROUP BY gp_segment_id, a, b, c`).

Pre-PR, only `nullptr == attnum` bailed out here; a system column's negative attno became a stray bit in the CBitSet that the intersection with `GetStatKeys()` (line 501) discarded, so the ndistinct item still applied to the user columns. With this change, the first system column abandons multivariate ndistinct for the *whole* grouping list, and the caller falls back to multiplying independent per-column NDVs — a group-count overestimate for queries that previously estimated correctly (and never crashed here).

Suggest splitting the two cases: keep `return false` for a missing mapping (the pre-existing behavior — note the output loop below re-looks-up and dereferences the attnum at line 575, so a plain `continue` for nullptr would just move the crash there), and `continue` for a system column: it never enters `attnums`, so it can never appear in `matched`, and the output loop keeps it for the regular per-column path — exactly the pre-PR net semantics.

```suggestion
		if (nullptr == attnum)
		{
			/* no colid -> attno mapping; extended stats are unusable */
			attnums->Release();
			return false;
		}
		if (0 >= *attnum)
		{
			/*
			 * System column: extended statistics never cover it and CBitSet
			 * cannot represent it. Skip just this column; it never enters
			 * 'attnums', so it stays unmatched below and is kept for the
			 * regular per-column ndistinct path.
			 */
			continue;
		}
```

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
```diff
@@ -365,7 +387,23 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
...
+					 * only once no dependency is fully matched by the
+					 * remaining attnums, so leaving the bit set would make
+					 * find_strongest_dependency() return the same dependency
+					 * forever. The clause stays unestimated and is handled
+					 * by the regular per-column path later.
+					 */
+					clauses_attnums->ExchangeClear(*attnum);
+					continue;
+				}
+				s2 = 1 / histogram->GetFrequency().Get();
```
linwen, line 421: Pre-existing defect, **not** introduced by this PR — flagging it because this PR rewrites exactly this line. `result_histograms` at the call site (CFilterStatsProcessor.cpp:327-332) is a `CopyHistHashMap` of the *unfiltered* input, whose histograms are normalized with total frequency ~1.0. So `s2 = 1 / GetFrequency()` is ~1.0 and the damping factor `degree + (1-degree)*s2` is algebraically a no-op, while `SetEstimated()` removes the implied clause from per-column estimation — its selectivity vanishes entirely. ORCA thus estimates ~sel(a=1) for `WHERE a=1 AND b=2` instead of the backend's `sel(a=1) * (deg + (1-deg)*sel(b=2))`; overestimation up to ~1/degree.

The repo's own expected output demonstrates it: `Correlated-Stat-Function-Dependency-3.mdp` (degree 0.989901) expects Rows = 99.999999 = 10001 * P(b=1), i.e. zero dependency contribution. (Separately, applying selectivity-space damping to an inverse scale factor is the wrong algebra even with a correctly filtered frequency P: `deg + (1-deg)/P != 1/(deg + (1-deg)*P)`.)

No action needed in this PR. Recommend filing a dedicated follow-up PR for the fix, using this thread as the tracking note.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
```diff
@@ -365,7 +387,23 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
 			 */
 			if (dependency_implies_attribute(dependency, *attnum))
 			{
-				s2 = 1 / result_histograms->Find(&colid)->GetFrequency().Get();
+				const CHistogram *histogram = result_histograms->Find(&colid);
+				if (nullptr == histogram)
```
linwen, line 405: Not blocking: this guard covers `nullptr == histogram`, but a *present* dummy histogram with ~zero total frequency passes it — `s2` becomes ~5e249, the scale factor caps at 1e250, and the row estimate silently collapses to MinRows=1 regardless of the true selectivity. A reachable (if narrow) trigger: `ALTER COLUMN TYPE` drops the column's pg_statistic row but deliberately keeps dependencies extended stats, after which ORCA builds a zero-bucket dummy histogram while the attno mapping entry remains.

Worth a follow-up: also bail out (same ExchangeClear + continue handling) when the histogram is empty or its total frequency is below `CStatistics::Epsilon`.

src/test/regress/sql/gporca.sql
```diff
@@ -3713,6 +3713,29 @@ create table cte_test(a int);
...
+insert into extstats_dim select g, g % 20, g % 3, g % 4, g % 5 from generate_series(1, 100) g;
+insert into extstats_fact select g % 100 + 1, g from generate_series(1, 1000) g;
+create statistics extstats_dim_nd (ndistinct) on d_a, d_b, d_c from extstats_dim;
+analyze extstats_dim;
+analyze extstats_fact;
+set optimizer = on;
+select count(*) from (
+  select d_a, d_b, d_c, sum(f_val) s
+  from extstats_fact, extstats_dim
+  where f_sk = d_sk and d_filter between 5 and 10
```
linwen, line 3734: Pre-existing, not blocking (anchoring here because `IsDependencyCapablePredicate` itself is outside the diff): `EsptPoint` covers **all** constant comparisons (Eq/G/GEq/L/LEq — see CStatsPredUtils.cpp:77-88), so `IsDependencyCapablePredicate` (CExtendedStatsProcessor.cpp:35) lets range predicates into functional-dependency estimation, where they get equality semantics and `SetEstimated()` drops their restriction entirely (e.g. `WHERE a=1 AND c>100` with dependency a=>c estimates ~sel(a=1) even when nothing satisfies c>100). The backend analogue cited in that file, `dependency_is_compatible_clause()`, accepts equality-to-pseudoconstant only. Suggest a TODO there and gating on `EstatscmptEq` in a follow-up.

Related: this `d_filter between 5 and 10` is collected as two dependency-capable range preds — exactly the shape the backend rejects. Harmless in this test (ndistinct-only statistics, so no dependency is ever applied), but worth keeping in mind for the follow-up: a dependencies-based case with `=` clauses would pin down the intended semantics.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
```diff
@@ -365,7 +387,23 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
...
+					 * bit must still be cleared: the outer loop terminates
+					 * only once no dependency is fully matched by the
+					 * remaining attnums, so leaving the bit set would make
+					 * find_strongest_dependency() return the same dependency
+					 * forever. The clause stays unestimated and is handled
+					 * by the regular per-column path later.
+					 */
+					clauses_attnums->ExchangeClear(*attnum);
+					continue;
+				}
```
linwen, line 420: The last sentence ("handled by the regular per-column path later") doesn't hold today: that path does `result_histograms->Find(&colid)->CopyHistogram()` at CFilterStatsProcessor.cpp:377-378, with the `GPOS_ASSERT` only *after* the deref. So if this guarded state (attno mapped, histogram missing) ever becomes reachable, the SIGSEGV this PR fixes is merely relocated ~50 lines downstream. ("Falls back to the independence assumption" would be equally untrue — the histogram for this exact colid is the thing that's missing.)

Recommendation: add a null guard at CFilterStatsProcessor.cpp:377 in this same commit — one line, it makes this comment's fallback claim real and actually closes the crash class this PR targets. Alternatively, reword the comment to state that the branch is purely defensive and the state is currently unreachable.

src/backend/gporca/libnaucrates/src/statistics/CStatistics.cpp
```diff
@@ -552,6 +552,25 @@ CStatistics::AppendStats(CMemoryPool *mp, IStatistics *input_stats)
...
+		UlongToIntMap *attnos_new = GPOS_NEW(mp) UlongToIntMap(mp);
+		AddAttnoInfo(mp, m_colid_to_attno_mapping, attnos_new);
+		AddAttnoInfo(mp, stats->m_colid_to_attno_mapping, attnos_new);
+		m_colid_to_attno_mapping->Release();
+		m_colid_to_attno_mapping = attnos_new;
+	}
+
+	if (nullptr == m_ext_stats)
+	{
+		m_ext_stats = stats->m_ext_stats;
```
linwen, line 579: Not blocking: this adoption (and the attno merge above) relies on two unenforced contracts — (1) both stats objects describe the same base relation (attnos are only meaningful relative to `m_ext_stats`'s relation and collide across tables), and (2) `m_ext_stats` is a non-owning pointer kept alive by the MD cache (`~CStatistics` never releases it, nothing AddRefs it). Both hold today because the only call sites (CGroup.cpp:870, CExpression.cpp:611) merge same-group stats, but a future caller merging different relations would get silently wrong selectivity. Cheap hardening:

```cpp
GPOS_ASSERT(nullptr == m_ext_stats || nullptr == stats->m_ext_stats ||
            m_ext_stats == stats->m_ext_stats);
```

plus a one-line ownership note here.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
```diff
@@ -277,6 +277,20 @@ CExtendedStatsProcessor::ApplyCorrelatedStatsToScaleFactorFilterCalculation(
 		{
 			ULONG colid = child_pred->GetColId();
 			INT *attnum = colid_to_attno_mapping->Find(&colid);
+			if (nullptr == attnum || 0 >= *attnum)
```
linwen: Nit: `nullptr == attnum || 0 >= *attnum` plus a near-identical rationale comment now appears at three sites (lines 280, 370, 469). A file-local helper, e.g. `static BOOL FUsableAttno(const INT *attnum)`, carrying the rationale once would match this file's existing static-predicate idiom (`IsDependencyCapablePredicate`, `dependency_implies_attribute`, ...).

linwen (changes requested):

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Good catch — you are right that the pre-PR net semantics for system columns were correct (the stray bit was discarded by the intersection and the column stayed on the per-column path), and folding them into the bail-out regressed that. Fixed as suggested in 80f38368a0: nullptr keeps the `return false` (the output loop re-dereferences the attnum, as you noted), and system columns just `continue`. Verified with correlated columns (ndistinct stats on x,y with 50 real combinations vs 2500 under independence): `GROUP BY x, y` estimates 50 and `GROUP BY gp_segment_id, x, y` estimates 85, i.e. the multivariate ndistinct still applies.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Agreed — the unfiltered copy has total frequency ~1.0, so the damping term degenerates to a no-op while SetEstimated() drops the implied clause entirely, and the Correlated-Stat-Function-Dependency-3 expected output indeed encodes exactly that. Also agreed the inverse-scale-factor algebra is wrong even with a correctly filtered frequency. Left untouched in this PR as suggested; I will file a dedicated follow-up PR referencing this thread.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Agreed — a present-but-dummy histogram with ~zero total frequency sails past the nullptr check and collapses the estimate to MinRows via the 1e250 cap. Will handle it in a follow-up together with the frequency-semantics fix from the other thread (same ExchangeClear + continue treatment when the histogram is empty or its total frequency is below CStatistics::Epsilon).

src/test/regress/sql/gporca.sql
JunfengYang: Right — EsptPoint covers all five constant comparisons, so range predicates get equality semantics here, unlike the backend's dependency_is_compatible_clause(). Added a TODO on IsDependencyCapablePredicate() in 80f38368a0 spelling this out; gating on EstatscmptEq (plus a dependencies-based test with = clauses) goes to the follow-up. The BETWEEN in the new test is deliberate only in the sense that it reproduces the crash — thanks for flagging that it is also two dependency-capable range preds; will keep that shape in mind for the follow-up test.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Fixed in 80f38368a0, taking the first option: MakeHistHashMapConjFilter() now null-checks the histogram lookup and estimates the clause with the default selectivity instead of dereferencing null (matching how unsupported predicates and empty input histograms are handled), so the crash class is actually closed rather than relocated. Reworded the comment here accordingly ("is later given the default selectivity by the regular per-column path").

src/backend/gporca/libnaucrates/src/statistics/CStatistics.cpp
JunfengYang: Added in 80f38368a0: the suggested GPOS_ASSERT (any side may be null, but two non-null ext-stats must be the same object) plus a comment on both contracts — attnos are only meaningful relative to m_ext_stats' base relation, and the pointer is non-owning, kept alive by the MD cache.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Done in 80f38368a0 — extracted a file-local FUsableAttno() carrying the rationale once, used at the two filter sites. The ndistinct site no longer uses the combined check since the two cases diverge there per the blocking thread (nullptr bails out, system columns are skipped per-column).

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Filed as PTT-1728 (High). Ticket carries the full reproduction: with dependencies stats present, ORCA estimates a=1 AND b=1 identically to a=1 alone (measured 100 vs planner's 53 vs 11 with the stats object dropped), plus both layers of the root cause (unfiltered-copy frequency ~1.0 making the damping a no-op while SetEstimated() drops the clause; inverse-space algebra) and the note that the Correlated-Stat-Function-Dependency mdp family fossilizes the buggy output and needs a hand-verified re-bless.

src/backend/gporca/libnaucrates/src/statistics/CExtendedStatsProcessor.cpp
JunfengYang: Filed as PTT-1729 (High) — severity upgraded from this thread's assessment: the trigger turns out to be natural, not narrow. Measured on a demo cluster: ALTER TABLE ... ALTER COLUMN ... TYPE deletes the column's pg_statistic row but keeps pg_statistic_ext_data, and from that moment until the next ANALYZE, ORCA estimates rows=1 for equality filters on the covered columns (true count 100; the zero-frequency placeholder histogram passes the nullptr guard, s2 = 1/1e-250, scale factor ~5e249, MinRows floor). Fix approach per this thread: extend the guard to empty/near-zero-frequency histograms with the same ExchangeClear + continue handling.

linwen (approved):
Looks good to me.

PR #252, opened 2026-08-26
Fix ssh error reporting, a ~/.bashrc deletion hazard, and PAM lockout in CI sshd setup

adam8157:
Three independent fixes in the management-utility and CI shell layer, found
while hardening the behave suite. One commit each.

## 1. `gpMgmt: report ssh failures from REMOTE_EXECUTE_AND_GET_OUTPUT`

The function captured `$?` after piping ssh output through the banner filter
(`awk | tail`), so a failed ssh was reported as success with empty output and
every caller proceeded on an empty string. Its exit status was also always 0
(the function ends with `echo "$OUTPUT"`), leaving the rc checks in
`POSTGRES_VERSION_CHK`, `CHK_FILE`, `CHK_DIR` and gpinitsystem's standby-IP
`ERROR_CHK` dead.

Fix: run ssh separately from the filter so the real status is captured and
returned. `[FATAL]` is reserved for rc 255 (ssh itself failed; the remote
command never ran) and displayed on stderr so it cannot leak into callers'
`$( )` capture; other non-zero statuses are legitimate answers from remote
probes (`grep -c` etc.) and are logged as INFO and passed through.

Caller audit: the function is only ever invoked inside `$( )`; no consumer
runs `set -e`; the one `trap RETRY ERR` (SED_PG_CONF) is dismantled before
returning and its call sites are ERR-exempt `if [ ... ]` conditions. The
revived rc checks all take sane paths.

## 2. `behave: never delete ~/.bashrc when no backup was taken`

`restore_bashrc()` inferred "no backup file = there was no `.bashrc` before
the test" and removed `~/.bashrc` on that guess. When the before hook raises
before reaching `backup_bashrc()` (e.g. cluster unreachable), the after hook
still runs restore — deleting the developer's real `~/.bashrc`. A module flag
now records whether the backup actually ran, and restore acts only on that.

## 3. `ci: disable PAM in sshd via a config drop-in`

The `UsePAM yes` / `PasswordAuthentication yes` seds match nothing on
EL9-family images (stock `sshd_config` has only Include / AuthorizedKeysFile /
Subsystem lines), so UsePAM stays at its default of yes and sshd accepts
gpadmin's key but refuses the session ("Access denied for user gpadmin by PAM
account configuration"). Every bare ssh/rsync exits 255 while gpssh (different
PAM path) works. State the settings in an `sshd_config.d` drop-in when the
stock config includes that directory, append otherwise; also create
`/var/empty/sshd` and `/run/sshd` for privilege separation.

## Verification

- Fix 1: function-level matrix — healthy command rc 0 with output intact;
  remote exit 3 passed through without FATAL; unreachable host rc 255 with
  empty stdout and FATAL on stderr and in the log; banner stripping unchanged
  (rc propagated through a banner-printing stub shell).
- Fix 2: isolated harness over a temporary HOME — existing `.bashrc`
  round-trips through backup/edit/restore; a scenario-created `.bashrc` is
  removed when none existed before; with the backup step skipped, a
  pre-existing `.bashrc` survives restore.
- Fix 3: `bash -n` clean; drop-in heredoc content byte-verified; the same
  approach is running green in CI on Rocky Linux 9 images.

### DIFF
--- a/gpMgmt/bin/lib/gp_bash_functions.sh
+++ b/gpMgmt/bin/lib/gp_bash_functions.sh
@@ -261,11 +261,23 @@ REMOTE_EXECUTE_AND_GET_OUTPUT () {
   LOG_MSG "[INFO]:-Start Function $FUNCNAME"
   HOST="$1"
   CMD="echo 'GP_DELIMITER_FOR_IGNORING_BASH_BANNER';$2"
-  OUTPUT=$( $TRUSTED_SHELL "$HOST" "$CMD" | $AWK '/^GP_DELIMITER_FOR_IGNORING_BASH_BANNER/ {seen = 1} seen {print}' | $TAIL -n +2 )
+  # Run the remote command on its own so RETVAL is the ssh/remote status;
+  # a pipeline into the banner filter would leave $? holding the filter's
+  # status instead, hiding connection failures.
+  OUTPUT=$( $TRUSTED_SHELL "$HOST" "$CMD" )
   RETVAL=$?
-  if [ $RETVAL -ne 0 ]; then
- 
-     LOG_MSG "[FATAL]:- Command $CMD on $HOST failed with error status $RETVAL" 2
+  OUTPUT=$( $ECHO "$OUTPUT" | $AWK '/^GP_DELIMITER_FOR_IGNORING_BASH_BANNER/ {seen = 1} seen {print}' | $TAIL -n +2 )
+  if [ $RETVAL -eq 255 ]; then
+     # 255 is ssh's own status for a connection or authentication failure:
+     # the remote command never ran.  Display on stderr: this function's
+     # stdout is the remote command's output, captured by every caller.
+     LOG_MSG "[FATAL]:- Command $CMD on $HOST failed with error status $RETVAL" 2 >&2
+  elif [ $RETVAL -ne 0 ]; then
+     # The remote command ran and exited non-zero.  Callers probe with
+     # grep/test commands for which a non-zero status is a legitimate
+     # negative answer, so record it without raising an error; the status
+     # is passed through to the caller below.
+     LOG_MSG "[INFO]:-Command $CMD on $HOST exited with status $RETVAL"
   else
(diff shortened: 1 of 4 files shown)


gpMgmt/bin/lib/gp_bash_functions.sh
```diff
@@ -261,11 +261,23 @@ REMOTE_EXECUTE_AND_GET_OUTPUT () {
...
-  OUTPUT=$( $TRUSTED_SHELL "$HOST" "$CMD" | $AWK '/^GP_DELIMITER_FOR_IGNORING_BASH_BANNER/ {seen = 1} seen {print}' | $TAIL -n +2 )
+  # Run the remote command on its own so RETVAL is the ssh/remote status;
+  # a pipeline into the banner filter would leave $? holding the filter's
+  # status instead, hiding connection failures.
+  OUTPUT=$( $TRUSTED_SHELL "$HOST" "$CMD" )
   RETVAL=$?
-  if [ $RETVAL -ne 0 ]; then
- 
-     LOG_MSG "[FATAL]:- Command $CMD on $HOST failed with error status $RETVAL" 2
+  OUTPUT=$( $ECHO "$OUTPUT" | $AWK '/^GP_DELIMITER_FOR_IGNORING_BASH_BANNER/ {seen = 1} seen {print}' | $TAIL -n +2 )
```
tkepe, line 269: may we rename RETVAL var to something more meaningful, .e.g., [SSH/REMOTE]_RETVAL?
Or move `OUTPUT=$( $ECHO "$OUTPUT" ....` assign after RETVAL checks? Maybe both?

tkepe (approved):
It looks good to me. I just dropped a nitpick, but it is not a blocker.

gpMgmt/bin/lib/gp_bash_functions.sh
adam8157: Thanks. It's a good point, but RETVAL is the convention throughout gp_bash_functions.sh (34 places), better to keep them.

PR #254, opened 2026-08-27
Hide indexes with non-default key collations from ORCA

avamingli:
## Problem

ORCA returns wrong results when a query uses an index whose keys have a non-default collation:

```sql
CREATE TABLE t (id int, val text) DISTRIBUTED BY (id);
CREATE INDEX ON t (val COLLATE "en_US.utf8");  -- database default is "C"
SET optimizer = on;
SELECT count(*) FROM t WHERE val >= 'a';       -- silently missing rows
```

The Postgres planner returns correct results (it refuses such an index for a default-collation predicate).

## Root cause

ORCA's metadata and DXL carry no collation information: during DXL-to-plan translation every collation is reconstructed as the type's default one (`gpdb::TypeCollation()`), and the only guard, `check_collation()`, inspects the query parse tree alone — a collation that lives only in `pg_index.indcollation` is never noticed. ORCA then emits index quals carrying the default collation against a btree ordered by a different comparator; the descent lands on the wrong leaf and silently misses rows.

## Fix

`IsIndexSupported()` now treats an index as unsupported when any of its key collations differs from the collation ORCA will assume for the key's type. The index becomes invisible to ORCA — the same outcome as the planner refusing it — while ORCA keeps planning the rest of the query, per the existing per-index-skip philosophy in that function (rather than a whole-query fallback). Comparing against `gpdb::TypeCollation()` rather than `DEFAULT_COLLATION_OID` keeps `name`-typed keys (whose type default is the C collation) usable. A multi-column index is hidden as a whole even when only a trailing key is affected, since ORCA's index metadata has no per-column granularity.

## Testing

- New `bfv_index` section: a `COLLATE "C"` index (portable — the check compares collation OIDs, no locale dependency) must not be used by ORCA even with table scans discouraged; a multi-column `(id, val COLLATE "C")` index is hidden as a whole; an index with the default collation is still chosen. Both planner and ORCA answer files updated.
- Negative control: on an unfixed build the new test fails deterministically — ORCA picks the mismatched index with an Index Cond in all three probes.
- The wrong-results scenario was verified manually in an `LC_COLLATE=C` database with an `en_US.utf8` index: planner and ORCA now agree (row-level EXCEPT diff is 0 in both directions), and ORCA plans a seq scan instead of the mismatched index.
- Local verification is arm64; CI amd64 is authoritative.


adam8157 (dismissed):
The fix itself looks good. The `NAMEOID || NAMEARRAYOID` exemption is a hole though: `namecmp()` does use the collation unless it is C, and ORCA reconstructs `name[]` with `DEFAULT_COLLATION_OID`, not C (see the `COLLATE "default"` in your own expected output). So both exempted cases still let ORCA scan a btree with a different comparator than the one it was built with.

Reproducer, on this branch, in a database with default collation `en_US.UTF-8`:

```sql
create table t1 (id int, nm name) distributed by (id);
insert into t1 select g, (chr(45 + g % 79) || g)::name from generate_series(1, 6500) g;
create index t1_idx on t1 (nm collate "en_US.utf8");
analyze t1;

set optimizer = off;
select count(*) from t1 where nm >= '5';   -- 5837, planner refuses the index
set optimizer = on;
set optimizer_enable_tablescan = off;
select count(*) from t1 where nm >= '5';   -- 5608, Index Scan using t1_idx

create table t2 (id int, tags name[]) distributed by (id);
insert into t2 select g, array[(chr(45 + g % 79) || g)::name] from generate_series(1, 6500) g;
create index t2_idx on t2 (tags);
analyze t2;

set optimizer = off;
select count(*) from t2 where tags >= array['5'::name];   -- 5837
set optimizer = on;
select count(*) from t2 where tags >= array['5'::name];   -- 5048, Index Scan using t2_idx
```

Dropping the exemption fixes both (verified here: both indexes get hidden, a plain `name` index is still used since 950 == 950). If you want to keep `name[]` indexes usable, fix `gpdb::TypeCollation()` to return C for `NAMEARRAYOID` as well; that also removes the existing planner/ORCA mismatch on `name[]` comparisons. For a portable regress case on `name`, `collate "POSIX"` (oid 951) works like `collate "C"` does on text.

Nit: stray line break in the commit message ("A\nmulti-column index").

avamingli:
Good catch, thanks. You're right that the name/name[] exemption was a real hole. I reproduced both cases on the branch (5608 and 5048 versus the planner's 5837).

I dropped the exemption entirely. The bare check (indcollation equal to TypeCollation of the table column's type) is already correct on its own for every case: a plain name index still matches, since its collation is C and TypeCollation(name) is also C, while a name index built with a non-default collation and a name[] index are now hidden, which is exactly what should happen. Added regress probes for both hidden cases (name with COLLATE "POSIX" per your portability tip, and name[]), and fixed the stray line break in the commit message.

I left TypeCollation() itself alone rather than teaching it to return C for name[]. That is a broader change to shared code and would be a separate follow-up if we want name[] indexes usable by ORCA.

adam8157 (commented):
Thanks for the quick turnaround, the fix looks right to me now. I re-ran my reproducer on this branch: the name + en_US index is hidden, and name[] returns the same count as the planner both with and without the index (it did not before, so the TypeCollation() change also fixes ORCA's name[] comparisons in general; might be worth a line in the commit message).

A few small things inline. Also, the first commit message still says name[] indexes are "correctly hidden" while the second commit makes them visible again; squashing the two would probably be the cleanest way to sort that out.

src/test/regress/regression.out
```diff
@@ -0,0 +1,5 @@
+Optimizer enabled. Using optimizer answer files whenever possible
```
adam8157, line 1: Looks like regression.out and regression.diffs got committed by accident.

src/test/regress/sql/bfv_index.sql
```diff
@@ -658,3 +658,78 @@ reset enable_seqscan;
...
+create index bfv_index_collate_name_idx on bfv_index_collate_name_tbl (nm);
+vacuum analyze bfv_index_collate_name_tbl;
+explain (costs off) select count(*) from bfv_index_collate_name_tbl where nm >= 'n99';
+select count(*) from bfv_index_collate_name_tbl where nm >= 'n99';
+-- a "name" column indexed with a NON-default collation is hidden: name
+-- comparisons do consult the collation, and ORCA reconstructs the qual with
+-- the type's default (C), so it would scan a differently-ordered btree.
+-- "POSIX" (like "C") orders identically to the default here, so the row count
+-- is unchanged and the assertion is on the plan shape
+create index bfv_index_collate_name_posix_idx on bfv_index_collate_name_tbl (nm collate "POSIX");
```
adam8157, line 715: bfv_index_collate_name_idx is still around here, so an unfixed ORCA also picks that one and the plan looks the same (I checked on main). Dropping the plain index first would make this a real negative probe: unfixed ORCA then uses the POSIX index, fixed ORCA falls back.

src/test/regress/sql/bfv_index.sql
```diff
@@ -658,3 +658,78 @@ reset enable_seqscan;
...
+vacuum analyze bfv_index_collate_name_tbl;
+explain (costs off) select count(*) from bfv_index_collate_name_tbl where nm >= 'n99';
+-- a "name[]" column has the same C type-default collation as "name" but its
+-- indcollation (950) differs from TypeCollation(name[]) (100, since only
+-- NAMEOID is special-cased there), so the visibility check must exempt
+-- NAMEARRAYOID too -- mirroring check_collation()'s NAMEOID || NAMEARRAYOID
+-- whitelist -- or the index would be wrongly hidden from ORCA
```
adam8157, line 722: This comment still describes the old exemption / check_collation() whitelist; the index is now visible simply because TypeCollation(name[]) returns C.

src/test/regress/regression.out
avamingli: Removed, those were stray pg_regress outputs. Also squashed the two commits into one and added a paragraph to the message about TypeCollation() fixing ORCA name[] comparisons in general.

src/test/regress/sql/bfv_index.sql
avamingli: Done, the plain name index is dropped before the POSIX one is created, so it is the only candidate. Fixed ORCA falls back and the plan is a Seq Scan now; expected files updated for both optimizer modes.

src/test/regress/sql/bfv_index.sql
avamingli: Rewritten to describe the TypeCollation() behavior instead of the old whitelist.

adam8157 (approved):

PR #255, opened 2026-08-27
Fix ORCA CTE consumer cardinality collapse from re-applied join predicates

JunfengYang:
## Problem

On a TPC-DS SF1000 cluster, Q67's finest ROLLUP HashAggregate is estimated at 8,495,943 rows vs 47,879,532 actual (5.6x under), under-provisioning the hash-agg spill layout ("Memory wanted 547,718,632 kB", 1,244s runtime).

The real defect is not in group-by estimation: every CTE consumer's row estimate collapses to producer/8.789 (for a 3-join producer), and the finest-level group estimate is simply clamped to that collapsed input. ROLLUP/CUBE/GROUPING SETS and `WITH ... MATERIALIZED` all read shared CTEs, so any such plan with >=2 joins in the producer is affected (shrink = 2.5x for 2 joins, 8.79x for 3, 32.6x for 4).

## Root cause

`CJoinStatsProcessor::CalcAllJoinStats` joins N child statistics incrementally but passes the full inner-join predicate conjunction at every step. A conjunct already applied at an earlier step has both sides inside the accumulated input, so `CStatsPredUtils::AssignExprsToOuterAndInner` no longer recognizes it as a join predicate and `ExtractJoinStatsFromJoinPred` re-applies it as an "unsupported" filter with the hardcoded default selectivity 0.4 — C(n,2) stale re-applications for n joins. CTE producer statistics are always derived on the standalone NAry-join expression (`CCTEInfo::InitDefaultStats`), and every consumer copies that cached result, which is why producer row counts look correct in EXPLAIN while all consumers are collapsed.

## Changes

- **CalcAllJoinStats**: split the inner-join predicates into conjuncts once and track consumed conjuncts in a bitset; each inner step only receives not-yet-consumed conjuncts, and conjuncts covered by a step's inputs are marked consumed afterwards (covered <=> applied, by the same `FCovered` test the extractor uses). Per-child LOJ ON predicates are used exactly once already and are untouched; 2-way joins run a single step and keep byte-identical behavior.
- **CStatsPredUtils**: new `ExtractJoinStatsFromJoinPredConjuncts` overload taking a conjunct array (non-owning); the existing entry point becomes a thin wrapper, so other callers are unchanged.
- **Guard for mismatched stats arrays**: `DeriveStatsWithOuterRefs` calls CalcAllJoinStats with arrays that are not the NAry join's children; the per-child LOJ predicate indexes then pair unrelated positions and can be read out of bounds. When sizes disagree, ignore the indexes and use inner-join semantics.
- **fix_mdps.py**: the plan-replacement regex matched from `<dxl:PlanHint`, silently truncating any minidump containing a plan hint (1172 of 1188 files); anchored the pattern and added refuse-to-write guards. `--dryRun` never worked (missing `global`); fixed.
- **New minidump** `CTE-Producer-3-Way-Join-Stats.mdp` pins producer == consumer row estimates (1,000,000) for a 3-join materialized CTE; fails against the old code (consumers at 113,778).

## Verification

- gporca ctest: 179/179 green in both Debug and RelWithDebInfo. Only two minidumps shifted, both as predicted: UnnestSQJoins (rows 199.68 -> 702, exactly the 3.515625 stale-reapplication signature) and Subq2OuterRef2InJoin (same plan shape, two inner row estimates 1.5 -> 2.0 from the LOJ-index guard).
- Cluster verification (demo cluster, debug build): CTE producer/consumer estimate ratio is now 1.0 for 2/3/4-join producers (was 2.5 / 8.789 / 32.587); Q67-shaped ROLLUP shows all 9 consumers equal to the producer estimate.
- ICW `installcheck-good` with optimizer=on: no row-estimate drift in any expected file; remaining failures are pre-existing build-config issues (missing libxml/JIT) verified unrelated.

Estimates on affected paths can only increase, so plans may change for workloads that relied on the erroneous shrink; the minidump corpus and ICW suggest a small blast radius.


JunfengYang:
Note for after merge: this needs a backport to WHPG_6X_STABLE. Code-walk on that branch confirms the same defect chain — CalcAllJoinStats re-passes the full inner-join conjunction at every incremental step, ExtractJoinStatsFromJoinPred re-applies covered conjuncts as unsupported filters (scale factor 1/0.4), and CTE producer stats are derived once on the NAry expression (CCTEInfo::InitDefaultStats) then copied by every consumer. Same expected collapse: 0.4^C(n,2). fix_mdps.py on 6X has the same two script bugs, though no 6X minidump currently contains <dxl:PlanHint>, so the truncation bug is latent there.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -160,13 +162,35 @@ CJoinStatsProcessor::CalcAllJoinStats(CMemoryPool *mp,
 							expr->Pop()->Eopid());
 				inner_or_simple_2_way_loj_preds =
 					(*expr)[GPOPT_ZERO_INNER_JOIN_PRED_INDEX];
+
+				if (predIndexes->Size() != num_stats)
```
adam8157: This size check is a heuristic and misses the most common shape. Under the default `optimizer_join_order=exhaustive2`, `EopttraceEnableLOJInNAryJoin` is on and `PexprCollapseJoins` turns a plain `a LEFT JOIN b` into a 2-child NAry join with predIndexes `[0,1]`. The second `CalcAllJoinStats` call in `DeriveStatsWithOuterRefs` always passes a 2-element `[outer_stats, local]` array, so `Size() == num_stats == 2` and the guard never fires; the mis-paired LOJ pairing is kept there. The first call has the same hole whenever the outer stats context count happens to equal the child count, so the outer-ref estimate of the same query depends on how many outer contexts there are.

I'd rather have `DeriveStatsWithOuterRefs` state explicitly that the LOJ indexes don't apply (a BOOL parameter or similar) than have the callee infer it from array sizes.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -160,13 +162,35 @@ CJoinStatsProcessor::CalcAllJoinStats(CMemoryPool *mp,
...
+				{
+					// the statistics objects do not correspond to the NAry
+					// join's logical children (e.g. they are outer-reference
+					// stats contexts), so the per-child LOJ predicate indexes
+					// do not apply; treat all steps as inner joins using the
+					// inner-join predicates only
+					predIndexes = nullptr;
```
adam8157, line 183: When the guard does fire, only child 0 of the `CScalarNAryJoinPredList` is used below. `SeparateOuterRefs` splits per child, so the outer-ref fragments of the LOJ ON clauses (children 1..n) are dropped from the outer-ref derivation entirely. That is what moves Subq2OuterRef2InJoin from 1.5 to 2.0: the correlation in that test lives only in the ON clause and the inner pred.

Either flatten all children into the inner conjunct set on this path (they are all applied with inner-join semantics anyway), or state the trade-off in the comment and the commit message.

src/backend/gporca/scripts/fix_mdps.py
```diff
@@ -46,9 +46,23 @@ def replacePlanSize(filename, actual, expected):
 def replacePlanInFile(filename, innerContent):
     with open(filename, "r") as fp:
         filedata = fp.read()
-    filedata = re.sub('    <dxl:Plan.*</dxl:Plan>\n',innerContent,filedata,flags=re.DOTALL)
+    # anchor on "<dxl:Plan " (trailing space) so the match cannot start at
+    # a <dxl:PlanHint> element earlier in the file
+    plan_pattern = re.compile(r'    <dxl:Plan .*</dxl:Plan>\n', flags=re.DOTALL)
```
adam8157: `PartTbl-NLJ.mdp` has an attribute-less `<dxl:Plan>` (line 542), which this pattern won't match. `replacePlanInFile` then prints "Could not find" and returns `None`, but `processLogFile` still prints "Changed query plan in ..." unconditionally, so the operator is told the file was updated when it wasn't. Suggest matching `<dxl:Plan[ >]` and returning a bool that the caller checks before printing.

src/backend/gporca/scripts/fix_mdps.py
```diff
@@ -46,9 +46,23 @@ def replacePlanSize(filename, actual, expected):
...
+    if not plan_pattern.search(filedata):
+        print("Could not find the plan section in %s, please update this file by hand" % (filename))
+        return
+    newdata = plan_pattern.sub(innerContent, filedata)
```
adam8157: `plan_pattern.sub(innerContent, filedata)` treats the plan XML as a regex replacement template, so a backslash inside a Const value (e.g. `\d` in a regex literal) raises `re.error: bad escape` on Python 3.7+, and `\n`/`\\` get rewritten silently. `plan_pattern.sub(lambda _m: innerContent, filedata)` avoids that. Pre-existing, but this line is being rewritten anyway.

adam8157 (dismissed):
Core fix looks right to me. I traced the covered <=> applied equivalence against ExtractJoinStatsFromJoinPred / AssignExprsToOuterAndInner / FCovered and the bitset bookkeeping is consistent with it; refcounting and CBitSet usage are fine; 2-way joins stay byte-identical; the UnnestSQJoins shift is purely the removed re-application (the new code never applies fewer distinct conjuncts than the old one, in any child order). Two comments on the guard commit and a couple of fix_mdps.py nits inline.

One more fix_mdps.py note that I can't anchor inline (not in the diff): in `failed_tests_file` mode `run_command` yields bytes while `processLogFile` matches with str patterns, so it TypeErrors on Python 3. `text=True` on the Popen fixes it. Pre-existing, optional.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
JunfengYang: Agreed, the size comparison was the wrong signal. 46cb0b1e23 replaces it with an explicit stats_are_join_children parameter: DeriveJoinStats and CLogicalIndexApply pass true, both CalcAllJoinStats calls in DeriveStatsWithOuterRefs pass false, and with true the old predIndexes->Size() == num_stats invariant is asserted again. I instrumented the new path on Subq2OuterRef2InJoin: both DeriveStatsWithOuterRefs calls now take it (2 stats, 2 predicate children — exactly the shape the heuristic let through).

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
JunfengYang: Took the first option in 46cb0b1e23: when the stats are not the join's children, the whole CScalarNAryJoinPredList is flattened into the inner conjunct set (inner-join predicates plus every LOJ ON predicate, constant-true children skipped), so the outer-reference fragments SeparateOuterRefs left in children 1..n are applied again, with the inner-join semantics that derivation documents. Subq2OuterRef2InJoin keeps its current numbers with the ON fragments back in, so the 1.5 -> 2.0 move in the guard commit comes from inner-join semantics replacing the mis-paired LOJ semantics rather than from the dropped fragments; noted in the commit message. No other minidump changed; ctest 179/179 on Debug and RelWithDebInfo.

src/backend/gporca/scripts/fix_mdps.py
JunfengYang: Fixed in 09cce3ce18: the pattern is now <dxl:Plan[ >], replacePlanInFile returns whether it rewrote the file, and processLogFile prints "Changed query plan" only in that case. Checked against PartTbl-NLJ.mdp (attribute-less <dxl:Plan>): its plan is matched and replaced, with the Metadata/Query sections intact.

src/backend/gporca/scripts/fix_mdps.py
JunfengYang: Done in 09cce3ce18 — substituting through a callable; verified that a backslash sequence inside the replacement plan survives literally.

JunfengYang:
Follow-ups pushed for the review: 46cb0b1e23 (explicit stats_are_join_children parameter instead of the size heuristic, predicate list flattened on the outer-reference path), 09cce3ce18 (fix_mdps.py: attribute-less <dxl:Plan>, literal replacement, and the bytes-vs-str TypeError in failed_tests_file mode via text=True — thanks for the extra note), and 8ef2ac53a3 (comment header for the new minidump, mirroring the request on the counterpart PR). gporca ctest 179/179 on Debug and RelWithDebInfo; no minidump changes.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -158,15 +164,64 @@ CJoinStatsProcessor::CalcAllJoinStats(CMemoryPool *mp,
 			{
 				GPOS_ASSERT(COperator::EopScalarNAryJoinPredList ==
 							expr->Pop()->Eopid());
-				inner_or_simple_2_way_loj_preds =
-					(*expr)[GPOPT_ZERO_INNER_JOIN_PRED_INDEX];
+				if (stats_are_join_children)
+				{
+					GPOS_ASSERT(predIndexes->Size() == num_stats);
```
haolinw, line 169: FYI: a pre-existing hints bug, surfaced earlier now. 

A partial Leading hint on two inner-join children next to a LEFT JOIN breaks the LOJ index array. The hints preprocessor emits one index per old child. The hinted tables become one child. Counts differ.
- On main: malformed plan, translator error, fallback to Postgres planner.
- With this PR: the new assertion at CJoinStatsProcessor.cpp:169 fires first. Same fallback.
  Repro:
```SQL
/*+ Leading((t1 t2)) */
SELECT * FROM t1 JOIN t2 ON t1.a=t2.a LEFT JOIN t3 ON t2.a=t3.a JOIN t4 ON t1.a=t4.a;
```
Not a blocker. Same user-visible result as before. joinhints.sql has partial hints, but each covers one LOJ pair, so this shape is untested. Suggest a separate fix in CJoinOrderHintsPreprocessor plus a joinhints case.

haolinw (approved):
Looks good to me.

src/backend/gporca/libnaucrates/src/statistics/CJoinStatsProcessor.cpp
JunfengYang: Will open a ticket to track it.

PR #257, opened 2026-08-28
[WHPG6] Restrict psql meta-commands in plain-text dumps 

linwen:
A plain-text dump is replayed by psql at restore time.  A hostile source server can cause the dump output to contain psql backslash meta-commands (e.g. \!), which psql then executes on the client running the restore, allowing arbitrary code execution.  (CVE-2025-8714)

Add \restrict and \unrestrict to psql: while "restricted", psql refuses every backslash command except a \unrestrict with the matching key. pg_dump, pg_dumpall and pg_restore wrap their plain-text output in \restrict KEY ... \unrestrict KEY, where KEY is a random alphanumeric string (pg_strong_random) or a user-supplied --restrict-key, so meta-commands injected between the markers are not run by a patched psql. Protection needs both a patched dumper and a patched psql; an old psql treats \restrict as an unknown command and is not protected.

Based on upstream commit c204bd39f7b57284363390f50e81abfc350a0fb9 (CVE-2025-8714) and its whpg7 backport
42005f421549d283e77f0c83c03e4208f2a95142, reimplemented for the PostgreSQL 9.4 base:

- command.c is monolithic on 9.4 (no exec_command_* split, no \if/active_branch), so the handlers go inline in exec_command() and the "restricted" gate directly in HandleSlashCmds().
- The key is read with OT_NO_EVAL, not OT_NORMAL, so a hostile dump cannot smuggle a backquote substitution through \restrict/\unrestrict; this avoids CVE-2026-18408, which upstream fixed only as a later follow-up.
- 9.4 has no DumpOptions; the key rides on RestoreOptions (AH->ropt), which the archiver already consults, so emission lands in RestoreArchive() and _reconnectToDB().  Uses pg_strong_random and pg_malloc (palloc is backend-only).
- generate_restrict_key()/valid_restrict_key() live in a new pg_backup_restrict.c, not the shared dumputils.c: on 9.4 dumputils.c is also linked into src/bin/scripts and psql, which do not link libcrypto, so the pg_strong_random() call there would break their link (undefined RAND_status).

- pg_dumpall also toggles restricted mode around the \connect that dumpCreateDB() emits to relocate a database onto a non-default tablespace -- a 9.4-only path (absent upstream) that psql would otherwise reject at restore time.  The pg_upgrade test scripts pass a fixed --restrict-key so their before/after pg_dumpall outputs stay byte-comparable.

The commands are also listed in psql's \? help and tab completion; the pg_dump, pg_dumpall, pg_restore and psql reference pages document --restrict-key and \restrict/\unrestrict; and src/test/regress/sql/psql.sql covers the gate (a normal backslash command is refused, a wrong key does not release, the matching key does).

Reported-by: Martin Rakhmanov
Reported-by: Matthieu Denais <litezeraw@gmail.com>
Reported-by: RyotaK <ryotak.mail@gmail.com>
Security: CVE-2025-8714


SKeyerror (approved):
LGTM.

PR #258, opened 2026-08-28
Backport CVE-2026-14680: reject SQL calls to internal-typed functions; honest nulls in combine functions

SKeyerror:
## Summary

Backport of the upstream PostgreSQL fix for **CVE-2026-14680** (CVSS 8.8) to WHPG7, per [PTT-1578](https://enterprisedb.atlassian.net/browse/PTT-1578).

Two upstream commits cherry-picked (`git cherry-pick -x`) from `REL_14_STABLE`, with the original commit messages unchanged. Since this repo merges with rebase-and-merge, the WHPG regression test coverage is folded into the corresponding cherry-picked commit so each commit is self-contained (upstream ships no regression tests for this CVE on any branch — verified on `REL_14_STABLE` and `master`):

1. **Reject calls from SQL to functions that take or return type internal** (upstream `155dacbc5479a9bfbcb47e88ff68a758c73b6a8c`)
   - Locks down function syntax, operator syntax, casts (`can_coerce_type` / `find_coercion_pathway`), and plpgsql's I/O-coercion fallback.
   - Conflict resolution: in `can_coerce_type()` the GPDB-specific ANYTABLE special case occupies the same spot as the new INTERNALOID check; both are kept (INTERNALOID check first, matching upstream placement right after the same-type short-circuit). In `pl_exec.c` the conflict was comment-text only; upstream's new comment is taken.
   - Adds the `internal_type_calls` regression test (greenplum schedule): rejection of internal-returning and internal-accepting function calls (the latter via `PREPARE p(internal)`), and rejection of casts to/from internal.
2. **Return nulls honestly in aggregate "combine" functions** (upstream `913fe0c316a6d024434e66f084acbb605293265a`)
   - WarehousePG's `numeric_combine`, `numeric_avg_combine`, `numeric_poly_combine` and `int8_avg_combine` already contain the `state1 == NULL -> PG_RETURN_NULL()` logic, so the backend part of this commit lands as the upstream comments only. Kept for traceability of the CVE fix.
   - Extends `internal_type_calls` with multi-stage aggregation coverage (the internal-typed combine-function path GPDB depends on): numeric/int8/int2 avg/sum/stddev/var over a distributed table, honest NULL states for all-NULL inputs, and the array_agg combine path.

### DIFF
--- a/src/backend/parser/parse_coerce.c
+++ b/src/backend/parser/parse_coerce.c
@@ -594,6 +594,10 @@ can_coerce_type(int nargs, const Oid *input_typeids, const Oid *target_typeids,
 		if (inputTypeId == targetTypeId)
 			continue;
 
+		/* reject all cases of casting something else to/from "internal" */
+		if (inputTypeId == INTERNALOID || targetTypeId == INTERNALOID)
+			return false;
+
 		/* 
 		 * ANYTABLE is a special case that can occur when a function is 
 		 * called with a TableValue expression.  A table value expression
@@ -2210,6 +2214,10 @@ find_coercion_pathway(Oid targetTypeId, Oid sourceTypeId,
 	/* Domains are always coercible to and from their base type */
 	if (sourceTypeId == targetTypeId)
 		return COERCION_PATH_RELABELTYPE;
+
+	/* Reject all cases of casting something else to/from "internal" */
+	if (sourceTypeId == INTERNALOID || targetTypeId == INTERNALOID)
+		return COERCION_PATH_NONE;
 
 	/* Look in pg_cast */
 	tuple = SearchSysCache2(CASTSOURCETARGET,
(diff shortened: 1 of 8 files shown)


SKeyerror:
CI: https://github.com/EnterpriseDB/warehouse-pg-packaging/actions/runs/33347189735

linwen:
The code change looks good — I reviewed it against upstream and it's a clean, correct backport of CVE-2026-14680.

Non-blocking suggestion: both commits resolve conflicts and add WarehousePG-specific regression tests, but neither carries a `WarehousePG adaptation:` block or a `Signed-off-by`. The write-up in the PR body is clear; it'd just be good to fold that same note into each commit message as an adaptation block plus a Signed-off-by. Not required for correctness — leaving it to your call.

Approving on the code.

linwen (approved):

PR #259, opened 2026-08-28
[WHPG6] Restrict accesses to non-system views and foreign tables during pg_dump

linwen:
pg_dump is vulnerable to a privilege-escalation race: a user could replace a table with a view or foreign table between the time pg_dump acquires its lock and the time it reads the data, so that a query pg_dump issues as the dumping (often superuser) role instead runs attacker-controlled SQL or a foreign-data wrapper of the attacker's choosing.  (CVE-2024-7348)

Add a restrict_nonsystem_relation_kind GUC (values "view" and "foreign-table") that makes the backend refuse to expand a non-system view or access a non-system foreign table.  pg_dump sets it to "view, foreign-table" for the duration of the dump, temporarily relaxing it to "view" only while reading a foreign table's own data.  The GUC is USERSET and defaults to empty, so it changes nothing unless pg_dump (or a user) asks for it; pg_dump sets it via set_config() gated on pg_settings, so it is silently ignored when dumping from a server that predates the GUC.

Based on upstream commit 79c7a7e29695a32fef2e65682be224b8d61ec972 (CVE-2024-7348) and its WHPG7 backport
b3e6063eb7d3f7ba1c47b0e9e8d99cbba2c95a44, reimplemented for the PostgreSQL 9.4 base:

- The view and foreign-table checks land in the same call sites as upstream (ApplyRetrieveRule/rewriteTargetView, get_relation_info, make_modifytable, GetFdwRoutine), all of which exist in 9.4.
- Dropped upstream's incidental non-CVE churn that came bundled in the cherry-pick context: the PG12 row-security SET in setup_connection, the generated-column INSERT handling and getPolicies()/other PG12-only functions in pg_dump.c, and the PG12-only rel->serverid assignment in get_relation_info(); none exist on the 9.4 base.
- restrict_nonsystem_relation_kind is added to unsync_guc_name.h so the dispatcher does not try to sync it to the segments (matching WHPG7); restore_command, which upstream also lists there, is a PG12 GUC absent from 9.4 and was not added.
- WHPG6 native external tables (relstorage 'x') are not foreign tables and do not pass through the FDW path, so they are out of scope here; views, the primary escalation vector, are fully covered.

The new GUC is documented in config.sgml (this doc hunk was present in the upstream commit but missing from the WHPG7 backport).

The view restriction is covered by src/test/regress/sql/create_view.sql. The foreign-table restriction is tested in gp2pg_postgres_fdw.sql.

Reported-by: Noah Misch
Security: CVE-2024-7348

### DIFF
--- a/contrib/postgres_fdw/expected/gp2pg_postgres_fdw.out
+++ b/contrib/postgres_fdw/expected/gp2pg_postgres_fdw.out
@@ -2390,6 +2390,15 @@ DELETE FROM ft2 WHERE c1 = 9999 RETURNING tableoid::regclass;
  ft2
 (1 row)
 
+-- test restriction on non-system foreign tables (CVE-2024-7348).
+SET restrict_nonsystem_relation_kind TO 'foreign-table';
+SELECT * from ft1 where c1 < 1; -- ERROR
+ERROR:  access to non-system foreign table is restricted
+INSERT INTO ft1 (c1) VALUES (1); -- ERROR
+ERROR:  access to non-system foreign table is restricted
+DELETE FROM ft1 WHERE c1 = 1; -- ERROR
+ERROR:  access to non-system foreign table is restricted
+RESET restrict_nonsystem_relation_kind;
 -- Test that trigger on remote table works as expected
  \! env PGOPTIONS='' psql -p ${PG_PORT} contrib_regression -f sql/postgres_sql/gp2pg_postgres_create_trigger.sql
 CREATE FUNCTION
--- a/contrib/postgres_fdw/expected/gp2pg_postgres_fdw_optimizer.out
+++ b/contrib/postgres_fdw/expected/gp2pg_postgres_fdw_optimizer.out
@@ -2390,6 +2390,15 @@ DELETE FROM ft2 WHERE c1 = 9999 RETURNING tableoid::regclass;
  ft2
 (1 row)
 
+-- test restriction on non-system foreign tables (CVE-2024-7348).
+SET restrict_nonsystem_relation_kind TO 'foreign-table';
+SELECT * from ft1 where c1 < 1; -- ERROR
+ERROR:  access to non-system foreign table is restricted
+INSERT INTO ft1 (c1) VALUES (1); -- ERROR
+ERROR:  access to non-system foreign table is restricted
+DELETE FROM ft1 WHERE c1 = 1; -- ERROR
+ERROR:  access to non-system foreign table is restricted
+RESET restrict_nonsystem_relation_kind;
 -- Test that trigger on remote table works as expected
  \! env PGOPTIONS='' psql -p ${PG_PORT} contrib_regression -f sql/postgres_sql/gp2pg_postgres_create_trigger.sql
 CREATE FUNCTION
--- a/contrib/postgres_fdw/expected/gp_postgres_fdw.out
+++ b/contrib/postgres_fdw/expected/gp_postgres_fdw.out
@@ -611,6 +611,15 @@ TRUNCATE TABLE "S 1"."GP 1";
 -- validate writes on master (mpp_execute set to master)
 -- ===================================================================
 ALTER FOREIGN TABLE gp_ft1 OPTIONS ( SET mpp_execute 'master' );
+-- test restriction on non-system foreign tables.
+SET restrict_nonsystem_relation_kind TO 'foreign-table';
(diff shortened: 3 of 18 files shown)


adam8157 (commented):
LGTM overall. I diffed the backend hunks against upstream 79c7a7e line by line
and they match, apart from the correct varlena.h -> builtins.h swap and the
unsync_guc_name.h entry. A few small things on the commit message and the
pg_dump side:

1. Commit message: the WHPG7 backport is cited as
   b3e6063eb7d3f7ba1c47b0e9e8d99cbba2c95a44, but on main the commit is
   b3e6063eb7d189568778873a8d7829b33f5d3620 (only the first 11 hex digits
   match). Perhaps this was pasted from a private branch or a rebased copy?
   Worth fixing so the reference resolves for anyone chasing the history later.

   In the same paragraph, "restore_command, which upstream also lists there"
   does not quite hold: the WHPG7 commit only added
   restrict_nonsystem_relation_kind to unsync_guc_name.h, and restore_command
   is a pre-existing neighbouring entry that showed up as diff context. I would
   just drop that sentence.

2. The re-added `/* Note: this syntax is only supported in 8.2 and up */` in
   dumpTableData_copy is not part of the upstream fix (it is unchanged context
   there); 6X deliberately removed it in 7dd89499335 along with the other
   pre-8.2 remnants. It looks like it came along with the WHPG7 cherry-pick.
   Please drop it so the diff against 79c7a7e stays minimal.

3. The `set_restrict_relation_kind(fout, "view")` relax/restore pairs in
   dumpTableData_copy and dumpTableData_insert are dead code on WHPG6:
   makeTableDataInfo returns early for RELKIND_FOREIGN_TABLE, so no
   TableDataInfo is ever created for a foreign table and these branches never
   run. Upstream REL_12 has the same dead code (PG12 has no
   --include-foreign-data either), so keeping them for parity is fine by me,
   but then the setup_connection comment ("This restriction is adjusted when
   dumping foreign table data") and the PR text about "temporarily relaxing it
   to 'view' only while reading a foreign table's own data" describe a window
   that never opens here. Either remove the blocks, or keep them with a
   one-line note that they are inert on the 9.4 base.

linwen:
> LGTM overall. I diffed the backend hunks against upstream 79c7a7e line by line and they match, apart from the correct varlena.h -> builtins.h swap and the unsync_guc_name.h entry. A few small things on the commit message and the pg_dump side:
> 
> 1. Commit message: the WHPG7 backport is cited as
>    b3e6063eb7d3f7ba1c47b0e9e8d99cbba2c95a44, but on main the commit is
>    [b3e6063](https://github.com/warehouse-pg/warehouse-pg/commit/b3e6063eb7d189568778873a8d7829b33f5d3620) (only the first 11 hex digits
>    match). Perhaps this was pasted from a private branch or a rebased copy?
>    Worth fixing so the reference resolves for anyone chasing the history later.
>    In the same paragraph, "restore_command, which upstream also lists there"
>    does not quite hold: the WHPG7 commit only added
>    restrict_nonsystem_relation_kind to unsync_guc_name.h, and restore_command
>    is a pre-existing neighbouring entry that showed up as diff context. I would
>    just drop that sentence.
> 2. The re-added `/* Note: this syntax is only supported in 8.2 and up */` in
>    dumpTableData_copy is not part of the upstream fix (it is unchanged context
>    there); 6X deliberately removed it in [7dd8949](https://github.com/warehouse-pg/warehouse-pg/commit/7dd894993351d8e6cf443d0ed274cbdb7a473710) along with the other
>    pre-8.2 remnants. It looks like it came along with the WHPG7 cherry-pick.
>    Please drop it so the diff against 79c7a7e stays minimal.
> 3. The `set_restrict_relation_kind(fout, "view")` relax/restore pairs in
>    dumpTableData_copy and dumpTableData_insert are dead code on WHPG6:
>    makeTableDataInfo returns early for RELKIND_FOREIGN_TABLE, so no
>    TableDataInfo is ever created for a foreign table and these branches never
>    run. Upstream REL_12 has the same dead code (PG12 has no
>    --include-foreign-data either), so keeping them for parity is fine by me,
>    but then the setup_connection comment ("This restriction is adjusted when
>    dumping foreign table data") and the PR text about "temporarily relaxing it
>    to 'view' only while reading a foreign table's own data" describe a window
>    that never opens here. Either remove the blocks, or keep them with a
>    one-line note that they are inert on the 9.4 base.

Thanks for reviewing, all issues have been addressed.

adam8157 (approved):

PR #261, opened 2026-08-31
Avoid overflow in Levenshtein distance calculations (CVE-2026-15742)

SKeyerror:
Backport of upstream PostgreSQL commit 9505175f2 (REL_14_STABLE) for CVE-2026-15742 (CVSS 8.8).

levenshtein() and levenshtein_less_equal() let the caller specify insertion, deletion, and substitution costs as arbitrary 32-bit integers, but distances were computed with 32-bit arithmetic. Large costs overflow and produce nonsensical results, and certain inputs to levenshtein_less_equal() cause out-of-bounds writes. Only reachable where the fuzzystrmatch extension is installed.

The fix converts the distance computation to 64-bit arithmetic and errors with "levenshtein distance out of range" whenever the final result won't fit in the returned 32-bit integer.

Cherry-picked with `-x`, applied with zero conflicts (the `pg_mblen_range()` context dependency already landed with the earlier bounds-checked pg_mblen backports). Upstream regression tests for fuzzystrmatch are included in the commit.

### DIFF
--- a/contrib/fuzzystrmatch/expected/fuzzystrmatch.out
+++ b/contrib/fuzzystrmatch/expected/fuzzystrmatch.out
@@ -35,6 +35,14 @@ SELECT levenshtein('GUMBO', 'GAMBOL', 2, 1, 1);
            3
 (1 row)
 
+SELECT levenshtein('GUMBO', 'GAMBOL', 1, 1, 2000000000);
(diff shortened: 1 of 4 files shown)


linwen (approved):
Looks good to me.

PR #263, opened 2026-08-31
Backport two Aug-2026 CVE fixes: to_char() TZ overflow and PL/Perl tied-value hardening

adam8157:
## Summary

Backports two independent upstream security fixes from the August 2026 set:

- **CVE-2026-14669** (CVSS 8.8) — `to_char()` copied the time-zone abbreviation
  into its fixed per-item output buffer with an unguarded `strcpy()`. A
  user-supplied overlength zone abbreviation (e.g. via a custom abbreviation
  file / `timezone` setting) could overflow the allocation. Both the `TZ` and
  `tz` format codes now bound the copy by `n->key->len * DCH_MAX_ITEM_SIZ` and
  raise `time zone format value too long` otherwise, matching the style of the
  CVE-2015-0241 fix.

- **CVE-2026-14670** (CVSS 8.8) — PL/Perl did not defend against Perl arrays
  and hashes that have been *tied*, which can run arbitrary Perl at unexpected
  moments and report inconsistent sizes on each inspection.
  `plperl_array_to_datum()` now reads each array's length exactly once (the
  caller passes it in via `dims[]`), and `plperl_to_hstore()` no longer trusts
  `hv_iterinit()`'s return as the hash size, growing the `Pairs` array
  dynamically instead.

## Background

Both are plain `cherry-pick -x` from `REL_14_STABLE` (the upstream branch whose
code structure matches this PG12 base); each commit keeps its original message
and provenance line.

- The `to_char()` fix applied verbatim.
- The PL/Perl fix required trivial conflict resolution: the `hstore_plperl.c`
  hunk adopts upstream's dynamic `Pairs` growth (initial guess 64 +
  `repalloc_array`), which also lets us drop the now-unused
  `#include "storage/shmem.h"` that had been carried only for `mul_size()`.
  `palloc_array()` / `repalloc_array()` are already present on `main`, so no
  incidental revert was needed.

Neither upstream fix ships regression tests. No catalog, ABI or WAL changes.

## Test plan

Verified on a PG12 demo cluster built from this branch:

- **to_char TZ**: normal abbreviations (`PDT`/`pdt`) unchanged; a 40-char zone
  abbreviation raises `time zone format value too long` for `TZ`, `tz` and
  `TZTZ`; a 20-char abbreviation (within the per-item bound) still prints.
- **PL/Perl**: 1-D and 2-D array round-trips correct; an hstore built from a
  200-entry Perl hash (exercising the `repalloc` growth past the initial 64)
  returns all 200 keys with correct values.
- `formatting.o`, `plperl` and `hstore_plperl` all compile warning-clean.

### DIFF
--- a/contrib/hstore_plperl/hstore_plperl.c
+++ b/contrib/hstore_plperl/hstore_plperl.c
@@ -2,7 +2,6 @@
 
 #include "fmgr.h"
 #include "hstore/hstore.h"
-#include "storage/shmem.h"		/* for mul_size() */
 #include "plperl.h"
 #include "plperl_helpers.h"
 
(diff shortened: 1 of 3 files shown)


linwen (approved):
Looks good to me. 
The first commit is a clean cherry-pick. The second contains conflicts resolve. A non-blocking suggestion: add WarehousePG adaptation/notes and signed-off-by in the second commit message.

PR #265, opened 2026-09-02
Fix wrong results from the one-item visibility cache for combo CIDs

### LINKED ISSUE #264: Cursor over a heap table returns an extra row after an aborted subtransaction DELETE
A cursor on a heap table can silently return a row its snapshot must not see. Reproduced on current main (0c7be1fcadc) and on 7.2.1.

`heapgetpage()` keeps a one-item (xmin, cmin) tuple-visibility cache: after one tuple is verified visible, consecutive tuples on the same page with the same raw xmin and raw command id skip `HeapTupleSatisfiesVisibility()` entirely. The cache keys on `HeapTupleHeaderGetRawCommandId()`, but real visibility for own-transaction tuples goes through `HeapTupleHeaderGetCmin()`, which resolves combo CIDs: for a tuple carrying `HEAP_COMBOCID`, the raw command-id field holds a backend-local combo-CID array index, not a real cmin.

The cache gate (`HEAP_XMAX_INVALID || HEAP_XMAX_IS_LOCKED_ONLY`, from fec5f0b56c3 "Fix the one-element tuple visibility cache in heap scans, for multi-xids.") happens to exclude most combo-CID tuples, because creating a combo CID requires a same-transaction delete/update, which sets a real xmax. The exception: the deleting subtransaction aborted. The tuple then keeps `HEAP_COMBOCID` with raw cid = combo index, and once any later scan sets the `HEAP_XMAX_INVALID` hint bit, the tuple passes the gate. Combo indexes and cmins are both small integers counted from 0, so a collision with another same-page, same-xmin tuple's plain cmin is easy. A cursor whose snapshot curcid lies between the two real cmins then returns a wrong extra row: the plain tuple primes the cache, the combo tuple falsely hits it.

Reproduce (single backend, e.g. a utility-mode session; it also reproduces through a dispatched cursor when the QE is stalled on the motion buffer — see the regression test in the fix):

```sql
CREATE TABLE combo_repro(a int);
CREATE TABLE filler(x int);

BEGIN;
INSERT INTO combo_repro VALUES (1);            -- t2: cmin = 0  (raw cid 0)
INSERT INTO filler VALUES (1);                 -- burn cid 1
INSERT INTO filler VALUES (2);                 -- burn cid 2
DECLARE c CURSOR FOR SELECT a FROM combo_repro;-- cursor snapshot curcid = 3
INSERT INTO filler VALUES (3);                 -- burn cid 4
INSERT INTO combo_repro VALUES (2);            -- t1: cmin = 5 (BEFORE the savepoint: keeps xmin = main xact)
SAVEPOINT s;
DELETE FROM combo_repro WHERE a = 2;           -- combo(5,6) -> combo index 0: t1 raw cid becomes 0
ROLLBACK TO s;
-- sets HEAP_XMAX_INVALID hint on t1 (deleter subxact aborted)
SELECT count(*) FROM combo_repro;              -- 2
-- cursor (curcid=3) must NOT see t1 (real cmin=5)
FETCH ALL FROM c;                              -- CORRECT: 1 row.  BUG: 2 rows.
COMMIT;
```

Observed FETCH output:

```
 a
---
 1
 2
(2 rows)
```

Correct is 1 row: the cursor's snapshot (curcid 3) must not see (2), whose real cmin is 5.

Control experiment: burn combo index 0 on another table first, so t1's combo index becomes 1 and no longer collides with t2's cmin 0 — the same sequence then returns the correct 1 row everywhere, so the wrong result depends exactly on the raw-value collision. pageinspect fingerprint of the bad tuple: infomask 0x0820 (`HEAP_COMBOCID|HEAP_XMAX_INVALID`), raw cid 0 colliding with the neighbor's plain cmin 0.

Found by inspection of the cache keying.

avamingli:
Fixes #264.

`heapgetpage()`'s one-item (xmin, cmin) tuple-visibility cache keys on `HeapTupleHeaderGetRawCommandId()`, but for a tuple carrying `HEAP_COMBOCID` that raw field holds a combo-CID array index, not a real cmin, while the actual visibility check resolves it through `HeapTupleHeaderGetCmin()`.

The cache gate from fec5f0b56c3 ("Fix the one-element tuple visibility cache in heap scans, for multi-xids.") excludes most combo-CID tuples, since creating a combo CID requires a same-transaction delete which sets a real xmax. But if the deleting subtransaction aborted, the tuple keeps `HEAP_COMBOCID` with raw cid = combo index, and once a later scan sets the `HEAP_XMAX_INVALID` hint bit it passes the gate. Combo indexes and cmins both count from 0, so the index easily collides with another same-page, same-xmin tuple's plain cmin; a cursor whose snapshot curcid lies between the two real cmins then sees a row it must not: the plain tuple primes the cache and the combo tuple falsely hits it (full repro in the linked issue).

Fix: exclude `HEAP_COMBOCID` tuples from the cache. Their raw command id is not comparable with a plain cmin; they merely fall back to the full visibility check, so eligible-tuple performance is unchanged. The gate covers both admission legs, which also closes the variant where a tuple carrying a stale combo cid is later locked-only by the same transaction.

Test: a dispatch-mode case appended to `combocid_gp` (both expected variants). The naive shape cannot trigger the bug through dispatch — QEs execute a cursor plan eagerly at DECLARE, so the tiny page is scanned before the combo state exists. The test therefore stalls the QE by scanning a large same-distribution-key table first via UNION ALL, so the page of interest is reached only at FETCH, after the hint-bit-setting scan. Verified the test fails on the unfixed binary and passes with the fix, under both ORCA and the Postgres planner.

Local verification is arm64 (full parallel_schedule green on the fixed build); CI amd64 is authoritative.

### DIFF
--- a/src/backend/access/heap/heapam.c
+++ b/src/backend/access/heap/heapam.c
@@ -473,11 +473,19 @@ heapgetpage(TableScanDesc sscan, BlockNumber page)
 				 * the xmax too, but the visibility rules get more complicated
 				 * with locked-only tuples and multi-XIDs, so it seems better
 				 * to just give up early.
+				 *
+				 * Don't use the cache for a HEAP_COMBOCID tuple either: its
+				 * raw command id is an index into the backend-local combo-CID
+				 * array, not a real cmin, so it must not be compared with
+				 * another tuple's plain command id. Such a tuple can still
+				 * have HEAP_XMAX_INVALID set: if the deleting subtransaction
+				 * aborted, the combo cid stays behind.
 				 */
 				bool		use_cache;
 
-				if ((theader->t_infomask & HEAP_XMAX_INVALID) != 0 ||
-					HEAP_XMAX_IS_LOCKED_ONLY(theader->t_infomask))
+				if (((theader->t_infomask & HEAP_XMAX_INVALID) != 0 ||
+					 HEAP_XMAX_IS_LOCKED_ONLY(theader->t_infomask)) &&
+					(theader->t_infomask & HEAP_COMBOCID) == 0)
 					use_cache = true;
 				else
 					use_cache = false;
--- a/src/test/regress/expected/combocid_gp.out
+++ b/src/test/regress/expected/combocid_gp.out
@@ -253,3 +253,58 @@ FETCH ALL FROM c;
 (99 rows)
 
 rollback;
+--
+-- Test that the one-item tuple-visibility cache in heapgetpage() ignores
+-- tuples carrying HEAP_COMBOCID.  Such a tuple's raw command id is an index
+-- into the backend-local combo-CID array, not a real cmin, so it must not be
+-- compared against another tuple's plain command id.
+--
+-- The scenario: a tuple whose deleting subtransaction aborted keeps its
+-- combo cid, and a later scan sets the HEAP_XMAX_INVALID hint bit, which
+-- used to make it eligible for the cache.  Combo-CID indexes and command
+-- ids both count from zero, so the combo index easily collides with the
+-- plain cmin of another tuple on the same page.  A cursor whose snapshot
+-- lies between the two real cmins then returned a wrong extra row: the
+-- plain tuple primed the cache and the combo tuple falsely hit it.
+--
+-- All rows use the same distribution key, so they land on the same segment
+-- and the same heap page.  The cursor query scans a big table before the
+-- table of interest (UNION ALL), so that the QE fills the motion send
+-- buffer and blocks at DECLARE time, and only reaches the page of interest
+-- during FETCH, after the hint bit has been set.
+--
+CREATE TEMP TABLE combocid_vischeck (a int, distkey int) distributed by (distkey);
+CREATE TEMP TABLE combocid_vischeck_filler (x int, distkey int) distributed by (distkey);
+CREATE TEMP TABLE combocid_vischeck_big (a int, distkey int) distributed by (distkey);
+INSERT INTO combocid_vischeck_big SELECT 0, 0 FROM generate_series(1, 200000);
+BEGIN;
+INSERT INTO combocid_vischeck VALUES (1, 0);          -- cmin 0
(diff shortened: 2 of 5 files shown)


adam8157:
The fix looks right to me. One concern about the test.

It depends on the cursor's QE stalling on the motion send buffer at DECLARE, which is really an interconnect implementation detail. That stall only happens reliably under UDPIFC (~48 KB of send/receive queue). `ic_tcp.c` sets no SO_SNDBUF and just parks in select(), so on loopback the kernel absorbs the 200000 rows without blocking; ic_proxy behaves similarly on hosts with large socket buffers. In those configurations the QE scans the small page before the combo state exists, and the FETCH result no longer says anything about the fix. The PR CI is UDPIFC-only, so it's green here, but the full release CI runs other interconnect types and this test could behave differently there.

`cursor.sql` (around lines 54-63) already has a deterministic recipe for exactly this shape: `gp_inject_fault('qe_got_snapshot_and_interconnect', 'suspend', <dbid>)` parks the cursor's reader QE right after snapshot + interconnect setup and before the plan runs, then `gp_wait_until_triggered_fault` / `resume` / FETCH. Arm it for a single occurrence, since the later `SELECT count(*)` also has a Gather motion. That also drops the 200000-row table, the UNION ALL and the MOVE, and fails loudly if the ordering ever breaks instead of passing silently.

avamingli:
Good point, thanks. Reworked the test with the cursor.sql recipe in 49ead17d868: suspend qe_got_snapshot_and_interconnect on the rows' segment before DECLARE (single occurrence, since the later count(*) passes the same fault point), set up the combo tuple and the hint bit, then resume and FETCH. The distribution key now hashes to content 0 to match the injected dbid, and combocid_gp moved to its own schedule group as fault-injecting tests require.

Re-verified on an unfixed build that the reworked test still catches the bug (FETCH returns the extra row) and passes with the fix, under both optimizers.

adam8157 (approved):

PR #268, opened 2026-09-03
Remove the unused gpMgmt/bin/go-tools Go utilities

adam8157:
## Summary

Delete `gpMgmt/bin/go-tools`, the Go-based `gp` hub/agent utility. WHPG has
never built, installed, tested or referenced it, and has no plan to maintain
it; the shipped management tooling remains the Python `gpMgmt/bin` scripts.

## Background

Dependabot keeps monitoring the directory's `go.mod`/`go.sum` and opening
bump PRs against it for code we never compile. Removing the directory stops
that at the source instead of suppressing it with a `dependabot.yml` ignore
rule or turning off security updates for the whole repository.

Pure deletion of 124 files; the code remains in history should the Go
utilities ever be revived.

## Test plan

- [x] `git grep` confirms nothing outside the directory references
      `go-tools`, `gp init cluster` or `gpservice`; no Makefile or CI
      workflow builds it
- [x] Change is deletion-only, no C/Python code touched


linwen (approved):
looks good to me.

PR #270, opened 2026-09-05
Check for USAGE privilege on types used by stored expressions, ALTER TABLE OF and CREATE TYPE AS RANGE (CVE-2026-6470)

linwen:
Backport of the upstream fix for CVE-2026-6470 (CVSS 4.3).

Roles without USAGE on a type could still create objects that depend on
it — stored expressions (defaults, CHECK constraints, index and partition
expressions, policies, triggers, rules, domain constraints, function
parameter defaults), `ALTER TABLE ... OF`, and `CREATE TYPE ... AS RANGE` —
which could then prevent the type's owner from altering it.  The fix adds
the missing USAGE checks in the command paths.

## Upstream commits
Cherry-picked from `REL_14_STABLE` (REL_12/REL_13 are EOL and did not
receive the fix), in upstream order:

- 323af0a4ee5 — Check for USAGE privilege on the subtype in CREATE TYPE AS RANGE.
- 1a358b8f2a2 — Check for USAGE privilege on types used by stored expressions.
- b6e45b4f824 — Check for USAGE privilege on the composite type in ALTER TABLE OF.

Upstream authorship is preserved; each commit message carries a
`WarehousePG adaptation:` block describing the deviations.

## Adaptations (details in the commit messages)
- `heap.c`: WarehousePG's `StoreAttrDefault()` takes extra QD/QE
  missing-value arguments; the GPDB call is kept and the USAGE check is
  inserted before it, where upstream places it.
- The `prosqlbody` (SQL-standard function bodies) and `stxexprs`
  (extended statistics on expressions) hunks are omitted — PostgreSQL 14
  features absent from this codebase.  WarehousePG has no
  dependency-recording call sites beyond upstream's, so no path is left
  unchecked.
- `create_type.sgml`: only the new USAGE paragraph is added (upstream's
  hunk carried the PG14 multirange paragraph in its context).
- `privileges.sql`: `test12` is created `DISTRIBUTED RANDOMLY` so that
  `ALTER TABLE test12 ALTER COLUMN c TYPE bigint` exercises the
  rebuild-without-USAGE path instead of failing on the distribution key.
- `privileges_optimizer.out` / `rangetypes_optimizer.out`: the GPORCA
  answer files receive the same new expected output.

## ABI
Upstream records one ABI change for this CVE in `.abi-compliance-history`
(18ec528b8c3): `AlterDomainAddConstraint()` gains a `bool is_readd`
parameter.  In WarehousePG the function has exactly two in-tree callers
(`utility.c`, `tablecmds.c`), both updated by the patch, and no callers in
`contrib/` or `gpcontrib/`.  The two new `dependency.h` functions are
additive; no struct or enum layout changes.

## Testing
- Backend builds cleanly; no new warnings.
- Live check on a gpdemo cluster: `CREATE TYPE ... AS RANGE` on a subtype
  without USAGE fails with `permission denied for type`, succeeds after
  `GRANT USAGE`.
- `rangetypes` and `privileges` pass in both planner (`optimizer=off`) and
  ORCA modes.  (`privileges` depends on `int8_tbl` from the `int8` test,
  so it was run as `int8 privileges`.)

### DIFF
--- a/.abi-check/7.1.0/postgres.symbols.ignore
+++ b/.abi-check/7.1.0/postgres.symbols.ignore
@@ -15,3 +15,4 @@ gp_resgroup_print_operator_memory_limits
 pqFunctionCall3
 SetTuplestoreDestReceiverParams
 build_gpqeid_param
+AlterDomainAddConstraint
--- a/doc/src/sgml/ref/create_type.sgml
+++ b/doc/src/sgml/ref/create_type.sgml
@@ -176,6 +176,11 @@ CREATE TYPE <replaceable class="parameter">name</replaceable>
     the range type.  See <xref linkend="rangetypes-defining"/> for more
     information.
    </para>
+
+   <para>
+    To be able to create a range type, you must have <literal>USAGE</literal>
+    privilege on the subtype.
+   </para>
   </refsect2>
 
   <refsect2>
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -1772,6 +1772,83 @@ recordDependencyOnSingleRelExpr(const ObjectAddress *depender,
 }
 
 /*
+ * We require USAGE on a type to store a dependency on it.  This helper
+ * function does the appropriate privilege checks.
+ *
+ * NB: Other objects have privileges of their own, but recording those
+ * dependencies doesn't require holding them.  For example, an expression may
+ * reference a function for which the user lacks EXECUTE.  Instead, EXECUTE is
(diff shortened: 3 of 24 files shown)


JunfengYang (approved):
Verified hunk-by-hunk against the three upstream commits; GP-specific surfaces (dependency call sites, LIKE paths, ABI, QD/QE) all clean. LGTM.

Worth a release note: restoring a pre-fix dump as a non-superuser can now fail on objects this fix forbids — may need a GRANT USAGE first.

PR #273, opened 2026-09-07
Test ORCA multivariate ndistinct with a system grouping column

JunfengYang:
## Why

Follow-up to #251. Its review found that a system column in the grouping list (e.g. `GROUP BY gp_segment_id, x, y`) made `ApplyCorrelatedStatsToNDistinctCalculation()` abandon multivariate ndistinct for the whole list; the merged version skips the system column and keeps the user columns on the `(ndistinct)` item. That fix landed without a test, and the review of its port asked for one so a future refactor cannot silently reintroduce the fallback.

## What

A new `gporca` block: a table whose `x` and `y` are functionally related (50 real combinations, 2500 under independence) with an `(ndistinct)` statistics object, grouped once on `x, y` and once on `gp_segment_id, x, y`. Estimates are read from the first `EXPLAIN ANALYZE` line because the test's matchsubs hide `rows=` in plain `EXPLAIN` output. The block then drops the statistics object and repeats both queries, so the file also records what the independence fallback looks like.

| query | ORCA, with stats | ORCA, no stats | planner, with stats | planner, no stats |
|---|---|---|---|---|
| `group by x, y` | 50 | 1407 | 50 | 1000 |
| `group by gp_segment_id, x, y` | 85 | 1780 | 150 | 1000 |

Test-only; no code change.

## Verification

- The test discriminates: rebuilding with the pre-fix bail-out restored (system column → `return false`) makes the second query estimate 1780, the same as with no statistics object; with the fix it estimates 85.
- `gporca` passes with `optimizer=on` and `optimizer=off` on a 3-segment demo cluster.

### DIFF
--- a/src/test/regress/expected/gporca.out
+++ b/src/test/regress/expected/gporca.out
@@ -14986,6 +14986,66 @@ explain (analyze, costs off, summary off, timing off) with cte as (select * from
  Optimizer: Postgres-based planner
 (5 rows)
 
+-- Multivariate ndistinct statistics must keep applying when the grouping list
+-- also contains a system column. No statistics object can cover a system
+-- column (it has no positive attno), so it is estimated on its own while the
+-- user columns still use the (ndistinct) item. x and y are functionally
+-- related (50 real combinations vs 2500 under independence), so the group
+-- estimate exposes which path was taken. Row estimates are read from the first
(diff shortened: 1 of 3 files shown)


tkepe (approved):
LGTM.

PR #274, opened 2026-09-07
Fix issues on ppc64le for WarehousePG 7

adam8157:
## Summary
Server-side fixes that make WHPG 7 build and pass the full smoke suite on
ppc64le (el8/el9). Eight self-contained commits, each fixing one issue found
on Power: a real portability bug in ORCA, a shared-memory sizing assumption,
an interconnect index type, and four test/CI-timing issues.

## Background
Earlier ppc64le testing recorded a number of failures; this lands the actual
fixes. The load-bearing one is the ORCA stack walker reading the return
address from the wrong slot on the ppc64 ELF ABI, which crashed 58 of the 179
ORCA unit tests in Debug builds. The rest: mirror_replay / gp_fastsequence /
crash_recovery_dtm timing assumptions, 64KB-page shared-memory drift, the
cursor IC table uint8 index, the behave ~/.bashrc backup order, and dropping
the now-unused vendored xerces build.

## Test plan
- [x] Full smoke matrix on el8 + el9 ppc64le, all jobs green
- [x] amd64 full matrix to confirm no regression from the all-platform changes

### DIFF
--- a/src/backend/gporca/libgpos/src/common/CStackDescriptor.cpp
+++ b/src/backend/gporca/libgpos/src/common/CStackDescriptor.cpp
@@ -70,8 +70,17 @@ CStackDescriptor::BackTrace(ULONG top_frames_to_skip)
 		}
 		else
 		{
-			// get return address (one above the base pointer)
+			// get return address of the function owning this frame
+#if defined(__powerpc64__)
+			// ELFv1/ELFv2: a function saves its LR into the caller's frame,
+			// two words past that frame's back chain; *next_frame passed the
+			// range check above, so the dereference stays on the stack
+			ULONG_PTR *frame_address = (ULONG_PTR *) *next_frame + 2;
+#else
+			// x86-64 / AArch64: the return address sits right after the
+			// saved frame pointer in the function's own frame
 			ULONG_PTR *frame_address = (ULONG_PTR *) (next_frame + 1);
+#endif
 			m_array_of_addresses[m_depth++] = (void *) *frame_address;
 		}
 
--- a/concourse/scripts/unit_tests_gporca.bash
+++ b/concourse/scripts/unit_tests_gporca.bash
@@ -3,15 +3,6 @@ set -exo pipefail
 
 GPDB_SRC_PATH=${GPDB_SRC_PATH:=gpdb_src}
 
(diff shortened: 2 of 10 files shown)


src/backend/gporca/libgpos/src/common/CStackDescriptor.cpp
```diff
@@ -70,8 +84,9 @@ CStackDescriptor::BackTrace(ULONG top_frames_to_skip)
 		}
 		else
 		{
-			// get return address (one above the base pointer)
-			ULONG_PTR *frame_address = (ULONG_PTR *) (next_frame + 1);
+			// get return address
+			ULONG_PTR *frame_address =
+				(ULONG_PTR *) (next_frame + GPOS_RETURN_ADDRESS_INDEX);
```
linwen: On ppc64 this reads the LR slot from the wrong frame. A function's saved LR is at +16 in its caller's frame (*next_frame), not its own frame. GCC's rs6000_return_addr() and glibc's ppc64 backtrace() both dereference the back chain first. The current code produces a valid but one-frame-shifted trace and breaks top_frames_to_skip semantics; entry 0 is only a valid address because IWorker::Self() happens to save its LR into BackTrace()'s own frame.

Suggested fix:

#if defined(__powerpc64__)
                      // ELFv1/ELFv2: the LR save slot is two words past the back chain
                      // in the caller's frame.
                      ULONG_PTR *frame_address = (ULONG_PTR *) *next_frame + 2;
#else
                      ULONG_PTR *frame_address = (ULONG_PTR *) (next_frame + 1);
#endif

*next_frame already passed the range check at the top of the loop, so the dereference is safe. GPOS_RETURN_ADDRESS_INDEX can then be removed.

linwen (commented):
This PR overall looks good to me.
One point needs check in CStackDescriptor.cpp‎ before merge.

src/backend/gporca/libgpos/src/common/CStackDescriptor.cpp
adam8157: Good catch, thanks. Fixed up.

linwen (approved):
looks good to me.

PR #244, opened 2026-08-21
[CVE-2026-14679] parser/executor/pl: guard FUNC_MAX_ARGS-sized arrays against overrun

Bonartze:
## Summary

Backport for WHPG7 / CVE-2026-14679: several fixed-size arrays dimensioned at `FUNC_MAX_ARGS` could be overrun by a function or aggregate declared with an extreme argument count.

The real upstream fix already present in the  history and cherry-picked both of its commits with `-x`, preserving Tom Lane's original authorship and adding `(cherry picked from commit ...)` trailers:

1. `c7f4628382071ef87f5e7fee8939f09db1e5e022` — "Protect some fixed-size arrays that have FUNC_MAX_ARGS elements." — `parse_func.c`, `parse_agg.c`, `nodeWindowAgg.c`, `funcapi.c`, `pl_comp.c`, `pltcl.c`. One conflict (in `funcapi.c`, an incidental `palloc_array()` cleanup unrelated to the security fix itself, caused by this fork's file predating later upstream refactors) — resolved by taking upstream's side, documented as a `WarehousePG adaptation:` note in that commit.
2. `cf44ff5e66353264451ac239b39e9053e8d138e6` — "Replace fixed-size, too-short array with a palloc'd one." — `namespace.c`'s `MatchNamedCall()`: `arggiven[]` needs `pronallargs` entries (which can exceed `FUNC_MAX_ARGS` when a function has OUT parameters), not a fixed `FUNC_MAX_ARGS` array. Applied cleanly, no adaptation needed.

PR #275, opened 2026-09-09
[6X][CVE-2026-14677 / CVE-2026-14679] Backport palloc_array() machinery and FUNC_MAX_ARGS overrun guards

SKeyerror:
## Summary

Backport for WHPG6 (PostgreSQL 9.4 base) of two CVEs from the 2026-08-10 upstream security batch, as four commits:

1. **Expand palloc/pg_malloc API for more type safety** — introduces the `palloc_array()`/`palloc0_array()`/`repalloc_array()` (and `pg_malloc_*` frontend) macros. Clean cherry-pick of upstream `b7f37af7c19`.

2. **Make palloc_array() and friends safe against integer overflow** — moves `add_size()`/`mul_size()` from shmem.c into mcxt.c/palloc.h, adds `palloc_mul()`/`palloc0_mul()`/`repalloc_mul()` (plus frontend copies in fe_memutils.[hc]) so the array macros always overflow-check. Adapted from upstream `61a9b4b6e4f` / master commit `01ec5cea3f1`. **6X adaptations:** the `*_extended()` variants are dropped (they depend on `palloc_extended()`/`MCXT_ALLOC_*`, which arrived in PG 9.5 and don't exist here, and nothing in these backports needs them); `repalloc_huge()` keeps its existing 9.4/GPDB implementation; `pg_add_size_overflow()`/`pg_mul_size_overflow()` come from the `common/int.h` size_t helpers already present on this branch; `pg_nodiscard` is added to c.h; mocker.py learns to strip `pg_noinline`. Note: this commit carries `Security: CVE-2026-6473`, inherited from its upstream source commit `61a9b4b6e4f` — a later CVE-2026-6473 audit of this branch will find that infrastructure here.

3. **CVE-2026-14677 — Use palloc_array() in pltcl and plperl to avoid overflow.** pltcl and plperl computed allocation sizes from argument counts with 32-bit arithmetic that could overflow. Adapted from upstream `aff9dac1c19` / master commit `19e803db4cf`. **6X adaptations:** the 9.4-era `compile_pltcl_function()`/`compile_plperl_function()` keep their malloc'd prodesc with fixed `FUNC_MAX_ARGS` arrays, so the upstream hunks converting per-nargs prodesc allocations don't apply (no dynamic allocation to overflow there); `pltcl_quote()` still uses Tcl's string-based API, so its bound is wrapped as `palloc(add_size(mul_size(strlen(...), 2), 1))`; `pltcl_build_tuple_result()` doesn't exist in 9.4 (hunk dropped); trigger paths were mapped onto the 9.4 `modattrs`/`modvalues`/`modnulls` triple.

4. **CVE-2026-14679 — Protect fixed-size arrays against overrun by extreme argument counts.** Adds argument-count checks in parse_func.c, parse_agg.c, nodeWindowAgg.c, fmgr.c, pl_comp.c, pltcl.c, plperl.c and plpython; converts `MatchNamedCall()`'s `arggiven[]` to a palloc'd array sized by `pronallargs`; uses `palloc_array()` in `get_func_arg_info()`. Adapted from upstream `c7f462838` + `cf44ff5e663` / master commit `49f92300403` (#244). **6X adaptations:** `get_aggregate_argtypes()` predates `Aggref.aggargtypes`, so the argument count is computed the way its own loops visit the arguments (direct args plus non-resjunk args); `get_func_trftypes()` doesn't exist in 9.4 (hunk dropped); instead of upstream's Tcl_DString conversion (which depends on the 9.6 pltcl rewrite), a `pronargs > FUNC_MAX_ARGS` guard is added before the fixed prodesc arrays are filled in **both** pltcl and plperl — the plperl guard has no upstream counterpart because PG12+ plperl had no fixed arrays left to fix. The same treatment covers two more 9.4-only fixed arrays with no upstream hunk to map: PL/Python's `PLyTypeInfo args[FUNC_MAX_ARGS]` in `PLy_procedure_create()` (the check tests `proc->nargs`, the count of IN args actually stored into `args[]` as derived from `proallargtypes`, rather than `pronargs`, since a corrupted catalog row could make the two differ), and the V0 old-style C-function path in `fmgr_info_C_lang()` (removed upstream in PG 10), whose `Oldstyle_fnextra.arg_toastable[FUNC_MAX_ARGS]` is filled from `pronargs`. `compile_pltcl_function()`'s local `buf[]` is also widened to 64 bytes as upstream did in the same fix.


linwen (changes requested):
### Blocking: `fmgr.c` V0 path still has an unguarded `FUNC_MAX_ARGS` array

The PR's sweep mirrors upstream/whpg7, but PG 9.4 still carries the V0 (api_version 0)
old-style C-function path that upstream removed in PG 10 — so there is no upstream or whpg7
hunk for it, and it was missed. In `fmgr_info_C_lang()`:

```c
case 0:
    fnextra = (Oldstyle_fnextra *)
        MemoryContextAllocZero(finfo->fn_mcxt, sizeof(Oldstyle_fnextra));
    ...
    for (i = 0; i < procedureStruct->pronargs; i++)
        fnextra->arg_toastable[i] =            /* bool arg_toastable[FUNC_MAX_ARGS], fmgr.c:69 */
            TypeIsToastable(procedureStruct->proargtypes.values[i]);

There is no pronargs > FUNC_MAX_ARGS check, so a pg_proc row with pronargs > 100 writes
arg_toastable[100..] past the palloc'd chunk and reads proargtypes.values[] past dim1.
This happens at the first fmgr_info() on the function — before GPDB's fmgr_oldstyle()
rejection, which only runs at call time. It is the same threat model this PR already guards in
pl_comp.c, plpy_procedure.c, pltcl.c and plperl.c (a catalog row edited with
allow_system_table_mods, or written by a binary built with a larger FUNC_MAX_ARGS).

Suggested fix, matching the wording of the other guards in this PR:

case 0:
    /*
     * Protect against overrun of the fixed-size arg_toastable array.
     * Ordinarily the parser would have checked this long since, but it's
     * possible that we are looking at a pg_proc entry that was made by a
     * server executable with a different value of FUNC_MAX_ARGS.
     */
    if (procedureStruct->pronargs > FUNC_MAX_ARGS)
        ereport(ERROR,
                (errcode(ERRCODE_TOO_MANY_ARGUMENTS),
                 errmsg_plural("cannot pass more than %d argument to a function",
                               "cannot pass more than %d arguments to a function",
                               FUNC_MAX_ARGS,
                               FUNC_MAX_ARGS)));

(Dropping the V0 branch entirely would also close it, since GPDB never executes V0, but the
guard is the smaller change.)

Before merge: commit metadata (author / reporters / cited SHA)

Since these commits keep upstream-author attribution, the metadata needs to match upstream.
Verified with git show -s --format=%an and git cat-file -t:

- f53da96404b ("Use palloc_array() in pltcl and plperl") is authored as Tom Lane, but
  upstream aff9dac1c19 and the whpg7 master commit are both by Heikki Linnakangas.
  Its Reported-by: Zhenpeng Lin does not match upstream, which credits
  Tulya Project, Team Dhiutsa, Bitecope Technologies Private Ltd.
  It also cites a master SHA that does not exist (19e803db4cf29b5d…); the real commit is
  19e803db4cf171abfdfc962a10ea930c34e5aa32.
- 6b0d269b543 ("Protect fixed-size arrays…") carries Reported-by: Yuhang Wu, which
  matches neither upstream commit: c7f4628382 credits Masahiko Sawada (also Reviewed-by),
  and cf44ff5e66 credits Zheng Yu and ylwangtju (Reviewed-by Michael Paquier /
  Masahiko Sawada). Both Reviewed-by trailers were dropped.

Please re-author f53da96404b as Heikki Linnakangas, fix the cited SHA, and restore the
upstream Reported-by / Reviewed-by trailers on both commits (if the CVE advisory reporter
should also be credited, add them rather than replace the upstream ones).

Minor (non-blocking)

- plpython guard quantity. The new check tests procStruct->pronargs, but the args[]
  loop it protects runs to proc->nargs, which is derived from proallargtypes minus
  OUT/TABLE entries. ProcedureCreate keeps the two equal, so this only diverges for a
  hand-edited catalog row — but for a self-consistent guard, check proc->nargs once it is
  computed (pltcl/plperl don't have this gap since prodesc->nargs = pronargs there).
- plpython guard is undocumented. It has no upstream or whpg7 counterpart (same situation
  as the plperl guard, which the commit message does justify). Please add plpython to the file
  list and the "6X adaptations" note in commit 6b0d269b543 and in the PR description.
- pltcl char buf[32]. snprintf(buf, sizeof(buf), "array set %d $__PLTcl_Tup_%d\n", i+1, i+1)
  fills exactly 31 chars + NUL at i+1 == 100 — zero headroom. Safe today (bounded snprintf),
  but upstream widened it to 64 in the same CVE fix; a one-line char buf[64]; avoids silent
  truncation if FUNC_MAX_ARGS is ever raised.
- commit 15d64183d9b carries Security: CVE-2026-6473 (inherited from upstream
  61a9b4b6e4f) while the PR title is 14677/14679. Worth one sentence in the PR description
  so a later 6473 audit isn't surprised to find this commit under that tag.

SKeyerror:
Thanks for the thorough review — all points confirmed and addressed in the force-pushed update (`66633046bd9`):

**Blocking — fmgr.c V0 path:** confirmed missed; the 9.4-only V0 branch has no upstream/whpg7 hunk to inherit. Added the suggested `pronargs > FUNC_MAX_ARGS` guard in `fmgr_info_C_lang()` before `arg_toastable[]` is filled, with the same wording as the other guards. Kept the guard rather than dropping the V0 branch, as the smaller change.

**Commit metadata:** all corrected.
- "Use palloc_array() in pltcl and plperl" is now authored as Heikki Linnakangas, its `Reported-by:` restored to upstream's (Tulya Project, Team Dhiutsa, Bitecope Technologies Private Ltd), and the cited master SHA fixed to `19e803db4cf171abfdfc962a10ea930c34e5aa32` — the previous value was an incorrectly expanded short SHA; apologies for that.
- "Protect fixed-size arrays…" now carries the upstream trailers from both source commits: `Reported-by:` Masahiko Sawada / Zheng Yu / ylwangtju, `Reviewed-by:` Masahiko Sawada / Michael Paquier. The incorrect `Reported-by: Yuhang Wu` (a reporter of a different CVE in the same batch) is removed.

**Minor points:** all taken.
- The plpython guard now tests `proc->nargs` after it is computed from `proallargtypes` (minus OUT/TABLE), i.e. the exact count the `args[]` loop stores — with a comment explaining why `pronargs` alone would not be self-consistent.
- plpython is now listed in the commit's file list and has its own "6X adaptations" bullet (as does fmgr.c); the PR description is updated to match.
- `compile_pltcl_function()`'s `buf[32]` is widened to 64 as upstream did.
- The PR description now notes that the palloc-overflow-safety commit carries `Security: CVE-2026-6473` inherited from its upstream source, for later audits.

Verified: net diff vs the previous branch tip is exactly the three code changes above (fmgr.c +16, plpy_procedure.c check move, pltcl.c one line); fmgr/plpython/plperl/parser all rebuild cleanly. pltcl still needs the CI `--with-tcl` build as noted in the description.

linwen (approved):
looks good to me.

PR #276, opened 2026-09-09
Backport ascii()/to_char()/PL/Perl Aug-2026 CVE fixes (CVE-2026-18024, CVE-2026-14669, CVE-2026-14670)

adam8157:
Backport three Aug-2026 upstream security fixes to `WHPG_6X_STABLE`.

## Summary

- **CVE-2026-18024** (CVSS 4.3) — `ascii()` trusted the lead byte of a UTF-8
  string and read the continuation bytes it implied without checking that the
  input was long enough, validating them only with `Assert()`. Invalid multibyte
  input could disclose a few bytes of QE memory; on a cassert build it fails the
  assertion. The length is now checked up front and every byte is validated
  with a real error. Includes the upstream regression test.
- **CVE-2026-14669** (CVSS 8.8) — `to_char()`'s `TZ`/`tz` patterns copied the
  time-zone abbreviation into the output buffer with an unguarded `strcpy()`.
  A POSIX-style `SET TIME ZONE` string with a long abbreviation overruns the
  12-bytes-per-format-character allocation. Both sites now error out with
  "time zone format value too long" when the abbreviation exceeds the budget.
- **CVE-2026-14670** (CVSS 8.8) — PL/Perl's `plperl_array_to_datum()` read
  `av_len()` of each input array twice, so a tied array that reports a
  different size on each call produced a corrupt result array. The length is
  now read once and carried in `dims[]`.

## Background

All three were fixed on `main` earlier (#236 and #263); these are the 6X
counterparts. The `ascii()` and `to_char()` commits are the upstream REL_14
fixes applied verbatim. The PL/Perl fix takes only the `plperl.c` half of the
upstream commit: `contrib/hstore_plperl` does not exist on this branch (it
arrived in PostgreSQL 9.5), so the `plperl_to_hstore()` hunk has nothing to
apply to.

## Test plan

- [x] `encoding` / `euc_kr` / `strings` regression tests green on a gpdemo cluster
- [x] `src/pl/plperl` `installcheck` 12/12 (rebuilt with `--with-perl`)
- [x] `to_char(..., 'TZ')` with a 20-char abbreviation prints; 40-char errors with "time zone format value too long" (TZ, tz and TZTZ)
- [x] `ascii()` rejects invalid multibyte input instead of over-reading; valid é / € / 😀 still return their code points
- [x] plperlu functions returning tied arrays that grow / shrink under
      iteration return well-formed arrays
- [x] `plperl_modify_tuple()` bounded-index rewrite: a trusted-plperl BEFORE
      trigger returning a tied hash crashes a 6X segment (SIGSEGV) before this
      change and inserts cleanly after; `src/pl/plperl` installcheck 12/12

Rebased onto `WHPG_6X_STABLE` (2026-09-10) after the palloc_array() overflow
series landed there; the `plperl_modify_tuple()` rewrite now uses
`palloc0_array()`.  The to_char() commit gained a note recording why its
cherry-pick reference is the REL_17 SHA (REL_14/15/16 have no standalone
to_char fix).

### DIFF
--- a/src/backend/utils/adt/formatting.c
+++ b/src/backend/utils/adt/formatting.c
@@ -2474,10 +2474,18 @@ DCH_to_char(FormatNode *node, bool is_interval, TmToChar *in, char *out, Oid col
 				INVALID_FOR_INTERVAL;
 				if (tmtcTzn(in))
 				{
-					/* We assume here that timezone names aren't localized */
+					/*
+					 * We assume here that timezone abbreviations aren't
+					 * localized, so ASCII-only downcasing is sufficient.
+					 */
 					char	   *p = asc_tolower_z(tmtcTzn(in));
 
-					strcpy(s, p);
+					if (strlen(p) <= n->key->len * DCH_MAX_ITEM_SIZ)
+						strcpy(s, p);
+					else
+						ereport(ERROR,
+								(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+								 errmsg("time zone format value too long")));
 					pfree(p);
 					s += strlen(s);
 				}
@@ -2486,7 +2494,14 @@ DCH_to_char(FormatNode *node, bool is_interval, TmToChar *in, char *out, Oid col
 				INVALID_FOR_INTERVAL;
 				if (tmtcTzn(in))
 				{
-					strcpy(s, tmtcTzn(in));
+					const char *p = tmtcTzn(in);
+
+					if (strlen(p) <= n->key->len * DCH_MAX_ITEM_SIZ)
+						strcpy(s, p);
+					else
+						ereport(ERROR,
+								(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+								 errmsg("time zone format value too long")));
 					s += strlen(s);
 				}
 				break;
--- a/src/backend/utils/adt/oracle_compat.c
+++ b/src/backend/utils/adt/oracle_compat.c
@@ -857,8 +857,10 @@ ascii(PG_FUNCTION_ARGS)
 	text	   *string = PG_GETARG_TEXT_PP(0);
 	int			encoding = GetDatabaseEncoding();
 	unsigned char *data;
+	int			len;
 
-	if (VARSIZE_ANY_EXHDR(string) <= 0)
+	len = VARSIZE_ANY_EXHDR(string);
+	if (len <= 0)
 		PG_RETURN_INT32(0);
 
 	data = (unsigned char *) VARDATA_ANY(string);
@@ -881,18 +883,31 @@ ascii(PG_FUNCTION_ARGS)
 			result = *data & 0x0F;
 			tbytes = 2;
 		}
-		else
+		else if (*data > 0xC0)
 		{
-			Assert(*data > 0xC0);
 			result = *data & 0x1f;
 			tbytes = 1;
 		}
+		else
+			ereport(ERROR,
+					(errcode(ERRCODE_CHARACTER_NOT_IN_REPERTOIRE),
+					 errmsg("invalid byte sequence for encoding \"%s\"",
+							GetDatabaseEncodingName())));
 
-		Assert(tbytes > 0);
+		/* All continuation bytes are present in the input */
+		if (tbytes >= len)
+			ereport(ERROR,
+					(errcode(ERRCODE_CHARACTER_NOT_IN_REPERTOIRE),
+					 errmsg("invalid byte sequence for encoding \"%s\"",
+							GetDatabaseEncodingName())));
 
 		for (i = 1; i <= tbytes; i++)
 		{
-			Assert((data[i] & 0xC0) == 0x80);
+			if (unlikely((data[i] & 0xC0) != 0x80))
+				ereport(ERROR,
+						(errcode(ERRCODE_CHARACTER_NOT_IN_REPERTOIRE),
+						 errmsg("invalid byte sequence for encoding \"%s\"",
+								GetDatabaseEncodingName())));
 			result = (result << 6) + (data[i] & 0x3f);
 		}
 
--- a/src/pl/plperl/plperl.c
+++ b/src/pl/plperl/plperl.c
@@ -1165,6 +1165,10 @@ get_perl_array_ref(SV *sv)
 
 /*
  * helper function for plperl_array_to_datum, recurses for multi-D arrays
+ *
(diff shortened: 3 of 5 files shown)


linwen (commented):
Thanks — this is a clean set of picks. All three commits are genuine `cherry-pick -x` with the upstream author preserved and the trailers verbatim; the `ascii()` and `to_char()` commits apply to `WHPG_6X_STABLE` with zero conflicts (so no adaptation note is needed), and the PL/Perl commit's adaptation note is accurate on every point I checked. The code edits in all three are byte-identical to upstream and to the corresponding merged `main` commits. CI is green.

I have one item I'd like addressed before merge, a set of follow-ups worth a ticket, and a couple of notes.

### Before merge: the 9.4 "hash half" of CVE-2026-14670 in `plperl_modify_tuple()`

The adaptation note drops the upstream `plperl_to_hstore()` hunk because `contrib/hstore_plperl` doesn't exist on this branch — true, but 9.4's core `plperl.c` has its own tied-hash-vulnerable iteration that upstream never needed to fix. In `plperl_modify_tuple()` (plperl.c ~1728–1757):

```c
modattrs  = palloc(tupdesc->natts * sizeof(int));
modvalues = palloc(tupdesc->natts * sizeof(Datum));
modnulls  = palloc(tupdesc->natts * sizeof(char));
slotsused = 0;
hv_iterinit(hvNew);
while ((he = hv_iternext(hvNew)))
{
    int attn = SPI_fnumber(tupdesc, key);
    if (attn <= 0 || tupdesc->attrs[attn - 1]->attisdropped)   /* validates the name only */
        ereport(ERROR, ...);
    modvalues[slotsused] = ...;
    modnulls[slotsused]  = ...;
    modattrs[slotsused]  = attn;
    slotsused++;                                                /* never bounded by natts */
}
```

A tied hash placed in `$_TD->{new}` whose `FIRSTKEY`/`NEXTKEY` keep returning a valid column name drives `slotsused` past `natts` and overflows all three arrays — a heap write reachable from a trusted-`plperl` BEFORE trigger (`tie` is available in trusted plperl; `plc_trusted.pl` itself ties `%ENV`). This is the same root cause the CVE addresses (tied containers reporting inconsistent contents), just on the hash side.

Upstream PG 14 didn't need this because its `plperl_modify_tuple()` already indexes by `[attn - 1]` into `palloc0_array`'d buffers, which is inherently bounded — and `main` has that same shape, so `main` is immune. 9.4's older `slotsused++` form is not.

Suggested fix (one line, and I'd fold it into the PL/Perl commit since it's the same CVE):

```c
if (slotsused >= tupdesc->natts)
    ereport(ERROR,
            (errcode(ERRCODE_PROGRAM_LIMIT_EXCEEDED),
             errmsg("$_TD->{new} contains more entries than the tuple has attributes")));
```

at the top of the loop body — or port `main`'s `[attn - 1]` indexing. If you'd rather land this PR as-is, that's workable too, but then CVE-2026-14670 shouldn't be considered fully fixed on 6X until a follow-up lands.

### Follow-ups worth a ticket (missing on both 6X and main)

All three sit in or directly on the call chain of the code this PR edits; none is a regression introduced here.

- **`ede3188e118`** (REL_14, 2026-08-17, "Make plperl's handling of Perl arrays safer and more consistent", Backpatch-through 14) — upstream's own follow-up to this CVE a week later. It bounds the SETOF-arrayref loop in `plperl_func_handler()` (`while ((svp = av_fetch(rav, i, FALSE)) != NULL)` never terminates for a tied array, since `av_fetch` returns a fresh magical mortal for every index → uninterruptible OOM/disk DoS) and caps `av_len()`-derived lengths at `MaxArraySize`. Note `av_count()` is Perl ≥ 5.36 and 6X's ppport.h (3.19) lacks it, so this needs a hand adaptation.
- **`e520ad34b48`** (REL_14, 2026-06-24) — `get_perl_array_ref()` dereferences `*sav` from `hv_fetch_string(hv, "array")` without a NULL check; a forged `PostgreSQL::InServer::ARRAY` object with no `array` slot segfaults the backend. One-token fix (`if (sav && *sav && ...)`).
- **`900a8d526ff` / `ea96fbe772a`** (2023, "Tighten array dimensionality checks in Perl -> SQL array conversion") — `array_to_datum_internal()` still accepts non-rectangular arrays, so `product(dims)` can exceed the accumulated element count and `construct_md_array()` reads uninitialised slots. `RETURNS text[] AS 'return [1, []];'` segfaults (reproduced on `main` binaries; identical logic here).

### Provenance note (no change needed, maybe one sentence)

The `to_char()` commit cites the REL_17 SHA (`12a6206864a`) rather than a REL_14 one. That turns out to be the right call: on REL_14/15/16 upstream committed the identical `formatting.c` hunks *inside an unrelated commit* — "Fix cascading standby reconnect failure after archive fallback" (`5fb3c638910` on REL_14, 2026-07-29) — with no `to_char`/CVE mention, so there is no standalone REL_14 commit to cite. `main`'s backport cites the same REL_17 SHA. A one-line note in the commit or PR body would save the next auditor the same detour when `git log --grep` on REL_14 comes back empty.

### Minor

- Neither the `to_char()` guard nor the PL/Perl fix has a regression test (upstream added none either). The `to_char()` one is cheap and deterministic if you want it:

```sql
SET timezone = 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA-5';  -- 40-char abbreviation
SELECT to_char('2000-01-01 00:00:00+00'::timestamptz, 'TZ');
-- ERROR:  time zone format value too long
```

  (The budget is 24 bytes per `TZ` node; `TZTZ` is two 24-byte budgets, not 48, so use a ≤24-char abbreviation as the positive case.) A tied-array test is also feasible with a hand-rolled `TIEARRAY`/`FETCHSIZE`/`FETCH` package in `src/pl/plperl/sql/`.

adam8157:
@linwen  Thanks Wen. All addressed:

- modify_tuple: rewrote to the bounded attn - 1 / heap_modify_tuple form (folded into the PL/Perl commit). Note the palloc_array() series that just landed on 6X only guards the allocation size — the unbounded slotsused writes, and so the tied-hash overrun, survive it. A/B confirms a SIGSEGV before and a clean insert after.
- Follow-ups: PRs #281 (main) and #282 (6X). On 6X, 900a and ede3 are hand-adapted (9.4 threads the ArrayBuildState by return value; no av_count()).
- to_char provenance: added a note in that commit explaining the REL_17 reference.

linwen (approved):
looks good to me.

PR #277, opened 2026-09-09
Backport CVE-2026-14680: reject SQL calls to internal-typed functions; honest nulls in combine functions (WHPG6)

avamingli:
## Summary

Backport of the upstream PostgreSQL fix for **CVE-2026-14680** (CVSS 8.8) to WHPG6, the PG9.4-based counterpart of PR #258 (already merged to WHPG7/main).

Two upstream commits cherry-picked (`git cherry-pick -x`) from `REL_14_STABLE`, with the original commit messages unchanged, plus one WHPG-authored commit adding regression coverage (upstream ships none on any branch for this CVE):

1. **Reject calls from SQL to functions that take or return type internal** (upstream `155dacbc5479a9bfbcb47e88ff68a758c73b6a8c`)
   - Locks down function syntax, operator syntax, and casts (`can_coerce_type` / `find_coercion_pathway`).
   - Conflict resolution: in `can_coerce_type()` the GPDB-specific ANYTABLE special case occupies the same spot as the new INTERNALOID check; both are kept (INTERNALOID check first, matching upstream placement).
   - **`pl_exec.c` is deliberately NOT touched** on this branch: PG9.4's plpgsql already rejects pseudo-typed (including `internal`) function arguments, return types, and variables at compile time (`src/pl/plpgsql/src/pl_comp.c`), and this branch's plpgsql has no `get_cast_hashentry`/cached-expression machinery at all — the upstream hunk patches an I/O-coercion fallback path that simply does not exist here, so there is nothing to fix.
   - Adds the `internal_type_calls` regression test (greenplum_schedule): rejection of internal-returning and internal-accepting function calls (the latter via `PREPARE p(internal)`), and rejection of casts to/from internal.
2. **Return nulls honestly in aggregate "combine" functions** (upstream `913fe0c316a6d024434e66f084acbb605293265a`)
   - WarehousePG's `numeric_combine`, `numeric_avg_combine`, `numeric_poly_combine` and `int8_avg_combine` already contain the `state1 == NULL -> PG_RETURN_NULL()` logic, so this lands as the upstream comments only (same finding as the WHPG7 backport). The `gp_array_agg` extension's `array_agg_combine` was checked too and is already correct.
   - Extends `internal_type_calls` with multi-stage aggregation coverage (the internal-typed combine-function path GPDB depends on): numeric/int8/int2 avg/sum/stddev/var/var_pop over a distributed table, and honest NULL states for all-NULL inputs.

## Test plan

- Built from a clean configure (`--disable-orca`) and ran `internal_type_calls` against a live `gpdemo` cluster: rejection paths (function calls, `PREPARE`, casts) all error as expected; multi-stage MPP aggregation (avg/sum/count/stddev/var_samp/var_pop, grouped and ungrouped, including an all-NULL group) still produces correct results.
- Ran `internal_type_calls` plus ~30 unrelated core regression tests (numeric, aggregates, opr_sanity, type_sanity, create_function_1/2/3, create_cast, plpgsql, select_having, etc.) from a proper schedule prefix: zero new failures attributable to this change (checked the diffs directly for any coercion/internal-type error — none found; the pre-existing failures in that vanilla schedule are unrelated GPDB-vs-upstream baseline noise, e.g. `DISTRIBUTED BY` notices).


linwen (commented):
Thanks — the three C hunks are a faithful backport of upstream 155dacbc / 913fe0c3, they compile safely on this branch (gnu99; `declared_arg_types` length matches `nargsplusdefs`; both new checks precede `make_fn_arguments`), and CI is green. I verified the 9.4-specific claims and they hold: the `pl_exec.c` hunk has no target here (9.4 has no `get_cast_hashentry`), the `numeric.c` change lands comment-only because the base already returns SQL NULL, and there is no missed 6X-only combine function (the four internal-typed `_combine`s are exactly the ones in numeric.c).

The PR is approvable as-is. What follows is one correctness bug worth folding in, one separate ticket, and a set of test/provenance tightenings — none of them block merge.

## Worth folding into this PR: all-NaN partial states are dropped in the two combine functions you touch

`numeric_combine()` and `numeric_avg_combine()` merge `state2`'s NaN count only inside `if (state2->N > 0)`:

```c
if (state2->N > 0)
{
    state1->N        += state2->N;
    state1->NaNcount += state2->NaNcount;   /* skipped when N == 0 */
    ...
}
```

But `do_numeric_accum()` counts a NaN input with `NaNcount++; return;` **without** incrementing `N`. So a segment whose group contains only NaN rows serializes a partial state of `N == 0, NaNcount > 0`. When that arrives as `state2` and `state1` is already non-NULL, the block is skipped and the NaN count is silently lost — `stddev` / `var_*` / `avg` over distributed NaN data return a finite number instead of NaN, and the answer depends on motion arrival order.

Upstream fixed this in `36df06e25c4` (2020) by hoisting the two additions above the `if`, and WHPG7/main already carries the hoisted form — only this branch still has the nested version. Since this PR is already editing these exact functions (comment-only), the ~2-line hoist is a natural thing to fold into commit 2. (Strictly a separate defect from CVE-2026-14680, but same functions, same PR.)

## Separate WHPG security ticket (6X + main): DESCRIBE-callback type confusion — no upstream fix exists

This is not a missed backport. The `describe` callback / `anytable` machinery is a WHPG-only extension — `validate_describe_callback()`, the `WITH (describe = …)` clause, and the `anytable` pseudo-type do not exist in upstream PostgreSQL (verified: 0 occurrences on REL_14_STABLE and master; present only on the WHPG branches). CVE-2026-14680's upstream fix locks down the parser's function/operator/cast chokepoints and neither did nor could cover this path, so there is no commit to cherry-pick — it needs a WHPG-side fix.

The hole is the same vulnerability class this CVE targets (a low-privilege user reaching an internal-typed function → pointer type confusion), through a path none of this PR's three checks guard:

- `validate_describe_callback()` (functioncmds.c) accepts any `(internal) -> internal` function with EXECUTE privilege — no superuser / C-language gate.
- `gist_box_compress` (OID 2579, `internal` argument, PUBLIC-executable) returns its argument pointer unchanged.
- `addRangeTableEntryForFunction()` (parse_relation.c) then does:

```c
fcinfo.arg[0] = PointerGetDatum(funcexpr);
d = FunctionCallInvoke(&fcinfo);
tupdesc = (TupleDesc) DatumGetPointer(d);
```

So any user who can CREATE FUNCTION can set `describe = gist_box_compress` on their own SQL function and make the parser walk a `FuncExpr` node as a `TupleDesc` → backend crash / wild read of column names. It is pre-existing and present identically on main, so it does not block this PR — but it warrants its own WHPG security ticket (6X + main), fixed with a superuser / C-language requirement in `validate_describe_callback()`, mirroring the internal-transtype rule in aggregatecmds.c.

## Test coverage — the new test claims more than it exercises (this is the only test for the CVE, so it's worth getting right)

- **`numeric_poly_combine` is never reached.** `var_pop(a::int8)` is aggregate OID 2718, whose combine function is `numeric_combine`, not `numeric_poly_combine` (the poly one is used by the int2/int4 variants, OIDs 2719/2642/…). A regression in `numeric_poly_combine` — the only combine with separate int128 / non-int128 paths — would pass. Use `var_pop(a)` (int4) or `stddev(a::int2)` instead.
- **No genuinely all-NULL group.** In `GROUP BY a % 3`, the single NULL row shares group 2 with ~333 non-NULL rows, so no group is all-NULL. Add e.g. `GROUP BY (a > 1000)`.
- **Under `optimizer=on` the combine functions aren't invoked at all.** `internal_calls_t` is never ANALYZEd (gp_autostats_mode = none), so ORCA sees reltuples = 0 and picks a single-stage Aggregate → the four `_combine`s run only in the `optimizer=off` leg, and the test asserts no plan shape in either. Add `ANALYZE internal_calls_t` (or `SET optimizer_force_multistage_agg = on`) plus an `EXPLAIN (COSTS OFF)` to pin the two-stage plan — otherwise a combine-NULL-state regression passes under ORCA.
- **The "casts rejected" block tests only pre-existing failures.** `CAST(123 AS internal)` and `'foo'::internal` already errored before this patch (only the message changed). The two coercions whose parse-time acceptance actually flipped are the CoerceViaIO cases upstream's message calls out — `SELECT 'foo'::text::internal;` and `PREPARE p(internal) AS SELECT $1::text;` — and neither is tested.
- **The `make_op` chokepoint and the `internal('foo')` function-cast path have no coverage.** e.g. `CREATE OPERATOR ==== (leftarg=internal, rightarg=internal, procedure=numeric_avg_combine); PREPARE p(internal,internal) AS SELECT $1 ==== $2;` should error from `make_op`; and `SELECT internal('foo'); SELECT internal(NULL);` guard the invariant that non-strict `internal_in` is what stops the function-cast path.

## Low / parity: the dropped `pl_exec.c` hunk

Upstream's commit also patched `pl_exec.c` to reject INTERNAL in plpgsql's I/O-coercion cast fallback; this PR correctly drops it (9.4's `exec_cast_value()` is the old form with no `get_cast_hashentry`). The security invariant still holds on 9.4 — an INTERNAL on either side of `exec_cast_value` hits `internal_out`/`internal_in`, both of which are non-strict and unconditionally error (so even NULL is rejected), and post-patch SQL can no longer materialize an internal-typed value to begin with. So this is a code-parity / documentation gap, not a reachable hole. Two ways to close it, either is fine:

- add the equivalent early guard to `exec_cast_value()` so the backport is literally equivalent to upstream and gives the same error:

```c
if (valtype == INTERNALOID || reqtype == INTERNALOID)
    ereport(ERROR,
            (errcode(ERRCODE_CANNOT_COERCE),
             errmsg("cannot cast type %s to %s",
                    format_type_be(valtype), format_type_be(reqtype))));
```

  plus a regression such as `DO $$ DECLARE v internal; BEGIN v := 'x'; END $$;` (expect rejection — note the message will be `internal_in`'s "cannot accept a value of type internal", not upstream's "cannot cast type …"); or
- keep it dropped but record the real mechanism in commit 1's message (currently the PR-body rationale attributes safety to pl_comp's compile-time checks, which don't see record fields typed from a query tupdesc; the actual backstop is the unconditional I/O coercion through non-strict `internal_in`/`internal_out`).

## Provenance / cosmetic

- **Branch/title mismatch:** the branch is `cve-2026-6464-whpg6` but the content is CVE-2026-14680 (6464 is a different PR). Worth renaming or at least noting, so finding this backport later by CVE doesn't lead to the wrong branch.
- **Commit 1** drops the upstream `pl_exec.c` hunk with no adaptation note, unlike the other CVE cherry-picks on this branch — add a `WarehousePG adaptation` block (see above).
- **Commit 2** is recorded as a behavior-fixing cherry-pick but is comment-only on 6X; a one-line note ("6X already returned SQL NULL here; this only syncs upstream comments") keeps `git blame` honest.
- **`parse_coerce.c`** also strips trailing whitespace from three lines of the WHPG-only ANYTABLE comment — not in upstream or the WHPG7 counterpart; keeping the hunk to the four new lines avoids gratuitous cross-branch divergence.

## Release note (behavior removal, matches upstream)

`CREATE AGGREGATE` with `STYPE = internal` whose sfunc/finalfunc declares the state slot as `anyelement`/`"any"` is now rejected at DDL time (the new INTERNALOID early return precedes the polymorphic-type acceptance in `can_coerce_type`). This matches upstream exactly, but a `pg_dump` / `gpupgrade` restore of such a user aggregate would now fail — worth a release-note line. No in-repo aggregate relies on it.

avamingli:
Thanks for the thorough review. Folded in the actionable items as three new commits (9fce086206d, 0d7ff2c5c49, 38a494718af):

- **NaN-count bug**: confirmed it live before fixing -- built the pre-fix binary, put an all-NaN partial state on one segment and a non-NaN one on another, and `sum()` silently returned a finite number instead of `NaN`. Hoisted the two additions above the `if (state2->N > 0)` in both `numeric_combine()` and `numeric_avg_combine()` (same fix as upstream 4284e118469, already on WHPG7/main), rebuilt, and confirmed it now returns `NaN` correctly.
- **Test tightening**: switched `var_pop(a::int8)` to `var_pop(a)` so it actually hits `numeric_poly_combine` (checked `pg_aggregate` -- `var_pop(int8)` uses plain `numeric_combine`), made the all-NULL group genuine (`GROUP BY (a > 1000)` instead of `a % 3`), added `ANALYZE` + `optimizer_force_multistage_agg = on` so the combine functions are exercised regardless of which optimizer the schedule runs under, and added the CoerceViaIO (`'foo'::text::internal`, `$1::text`), make_op (`CREATE OPERATOR` over an internal-typed function), and function-cast-path (`internal('foo')`, `internal(NULL)`) cases. Ran the test live against a built cluster -- all new cases produce exactly the expected errors/results.
- **parse_coerce.c whitespace**: restored the trailing whitespace the security commit incidentally stripped from the pre-existing ANYTABLE comment, matching WHPG7/main.

Not done here, left for you to call:
- **DESCRIBE-callback ticket**: agreed this needs its own WHPG security ticket (6X + main), not this PR. Haven't filed anything yet.
- **Release note**: couldn't find a per-PR release-note convention in this repo (just one monolithic release-9.4.sgml, updated separately at release time), so didn't guess at one.
- **pl_exec.c adaptation note / commit 2 comment-only note / branch name**: these are about the already-pushed, already-reviewed commits 1/2. Didn't amend them to avoid forcing a re-push/re-review; happy to do it as a follow-up if you'd rather have it in the commit messages than here.

linwen (commented):
Thanks for the follow-up. The code changes look good. Before approval, please clean up the commit history and adaptation notes as follows. We should keep each commit self-contained and internally consistent. Please fold the two fixup commits into the commits they correct, while keeping the separate upstream NaN fix standalone with its own provenance.

## Before merge
- Squash `0d7ff2c5c49` into `18df90110d7`. It only restores whitespace changed accidentally by the earlier commit and even produces trailing-whitespace warnings on its own. While rewriting that commit, please add a `WarehousePG adaptation:` note explaining that the upstream `pl_exec.c` hunk was omitted because PG 9.4 lacks `get_cast_hashentry()`; the remaining path is guarded by unconditional, non-strict `internal_in`/`internal_out`.

- Amend `95238c47b0b` to note that WHPG6 already returned SQL NULL from these combine functions, so this cherry-pick only synchronizes the upstream comments on this branch.

- Squash `38a494718af` into `3667655ed10` and update the commit message to describe the final test coverage rather than the review follow-up.

- Keep `9fce086206d` standalone because it is a separate upstream cherry-pick with its own provenance. Moving it directly after `95238c47b0b` would also make the history easier to follow.

The intended result is four commits: three upstream-derived commits, each retaining its own `(cherry picked from ...)` trailer, followed by the regression test commit. The resulting tree must remain unchanged. If the base is unchanged:

    git diff --exit-code 38a494718af <new-head>

If the branch is also rebased onto the latest `WHPG_6X_STABLE`, compare the old and new base-to-head series with `git range-diff` instead.

## Non-blocking: optimizer coverage
`optimizer_force_multistage_agg` is consumed only by ORCA. The legacy planner uses `gp_enable_multiphase_agg` and still chooses between one- and two-stage aggregation by cost, so the test does not strictly force the combine path under `optimizer=off`.

Either add plan assertions for both optimizer modes, or keep the test as-is and amend its comments and squashed commit message so they no longer claim that both optimizers are forced onto the combine path.

Separately, please ensure the DESCRIBE-callback issue is tracked through the security process and reference the ticket here once one exists.

avamingli:
Rewrote the history per the ask -- now four commits (9d30c734ac, 96c634f19a, 2acac65183, aa68c7cc0c):

- `9d30c734ac` "Reject calls..." now has the whitespace fix folded in (net effect: the ANYTABLE comment is untouched, only the INTERNALOID checks are added -- cleaner than before) plus the `WarehousePG adaptation:` note on why pl_exec.c is omitted.
- `96c634f19a` "Return nulls honestly..." now notes WHPG6 already returned SQL NULL here, so it's a comment-only sync with upstream.
- `2acac65183` "Fix mishandling of NaN counts..." kept standalone right after it, unchanged, still carrying its own `(cherry picked from ...)` trailer.
- `aa68c7cc0c` "Add internal_type_calls..." has the test-tightening folded in with a message describing the final coverage.

Verified `git diff --exit-code 38a494718af <new-head>` is empty except for the one intentional change below, confirming the rest of the tree is byte-for-byte identical to the old history.

On the optimizer-coverage note: went with the second option -- reworded the comment/GUC block instead of adding plan assertions. `optimizer_force_multistage_agg` only forces ORCA; under `optimizer=off` it's a no-op, but `gp_enable_multiphase_agg` (on by default) already makes the legacy planner choose two-stage aggregation for this query by cost, which I confirmed live against the built cluster. So both optimizers exercise the combine path, just via different mechanisms -- the comment now says that instead of implying the GUC forces both.

DESCRIBE-callback ticket: still not filed, no update from my side.

@linwen

linwen (approved):
looks good to me.

PR #278, opened 2026-09-09
Teach psql to skip in-line COPY ... FROM STDIN data after a failure (WHPG6)

avamingli:
## Problem

If a `COPY ... FROM STDIN` command failed before the server sent `PGRES_COPY_IN`, psql
did not realize it needed to consume the in-line data that follows the command in the
script/dump. It then tried to execute that leftover data as SQL. In the best case this
is wrong; in the worst case it's a SQL-injection hazard — a hostile server (or a
crafted plain-text dump replayed by psql) can arrange for a COPY it controls to fail,
then have its own attacker-chosen SQL run on the client. (CVE-2026-6464)

## Root cause / fix

Based on the upstream fix (commit `5c51ae4556f1e058779f48f95d3ddcf305fb6e8f`) and its
whpg7 backport (PR #241, commit `45f8fe2fa8bc78dd3b99b95bc3fc88347afab749`), but
reimplemented by hand rather than cherry-picked, because this branch is PostgreSQL
9.4-based and several pieces of infrastructure the upstream fix relies on don't exist
here:

- 9.4's `psqlscan.l` has no identifier-tracking machinery at all — the
  BEGIN/END-block and COPY-FROM-STDIN detection upstream shares via
  `init_idents[]`/`psqlscan_track_identifier()` was added later (starting with the
  commit that introduced PG15's SQL-standard function bodies). Since 9.4 has no such
  function-body syntax to worry about, this backport adds only the
  COPY-FROM-STDIN-detection half of that machinery: a new `copy_stdin_count`/
  `init_idents` state in `PsqlScanStateData`, populated by
  `psqlscan_track_identifier()` off the existing `{identifier}` lexer rule — not the
  unrelated BEGIN/END tracking.
- 9.4 has no `\if`/`\endif` (added in 9.6) and no `fe_utils`/`psqlscan_int.h` split,
  so there's no `ConditionalStack` lexer-state save/restore to extend; the
  prerequisite commits whpg7 needed for that reason don't apply here.
- `psql_scan_create()`/`psql_scan_setup()` take no reentrant-lexer arguments on this
  branch (single global lexer state), so the local re-scan in `SendQuery()` and the
  `mainloop.c` call sites use the 9.4 signatures directly rather than upstream's
  `psqlscan_callbacks`-based ones.
- Conflicts otherwise resolved against this branch's own `copy.c`/`common.c`
  structure (a differently-shaped `copystream` local in `ProcessResult()`, no cursor
  `\gdesc` sibling in `ExecQueryUsingCursor()`, etc).

Mechanically: a `copy_stdin_count`/`num_copy_from_stdin` count is threaded through
`SendQuery()`/`ProcessResult()`, and `handleCopyIn()` now accepts `conn == NULL` as a
"read and discard" mode, used when a COPY fails before ever entering `COPY_IN`.

## Test coverage

- Repo-wide sweep of `pg_regress` inputs (`sql/*.sql`, `input/*.source`) for
  `COPY ... FROM STDIN` / `\copy ... from stdin` commands expected to fail before
  entering `COPY_IN` state found 14 sites (across `alter_table.sql`, `copy2.sql`,
  `copyselect.sql`, `privileges.sql`, `gpcopy.source`) that needed a `\.` terminator
  added right after them — psql now consumes in-line data up to `\.` even when the
  COPY command fails, so without the terminator it would swallow whatever followed in
  the script. Each site was individually re-verified against a live cluster
  (before/after output diffed by hand).
- New coverage for the fix itself added to `psql.sql`/`psql.out`, mirroring
  upstream's own regression test; output verified byte-for-byte against a live
  `psql -a -q -X` run before committing.
- Live-verified end-to-end against unpatched vs. patched binaries in the same
  cluster: unpatched psql executes the injected "SQL" between a failed COPY and
  `\.` (a real exploit repro); patched psql discards it silently. Also verified
  multiple `COPY FROM STDIN` in one query string joined with `\;` (both
  failing-and-succeeding combinations), the `\copy` frontend path, and that normal,
  successful COPY is unaffected.
- Have not yet run the full regress schedule locally (no local fixture chain was
  built for a full run) — CI is the first full end-to-end run for this branch.

## Notes for reviewers

Local build/verification was done under an amd64 container (same arch family as CI),
so no arm64-vs-amd64 plan/behavior caveat applies here.


linwen (commented):
## Review: Request changes

PR #278 — CVE-2026-6464, WHPG6 hand-reimplementation of upstream `5c51ae4556f`.

One HIGH-severity issue should be fixed first.

### HIGH — WHPG6's legacy COPY syntax bypasses the count protection

`psqlscan.l` keeps only the first 8 top-level identifiers of a statement:

```c
char init_idents[8];
```

This inherits upstream's assumption that at most `BINARY database.schema.table` (≤ 4 identifiers) can precede `FROM` in a COPY. But WHPG6's grammar still supports the pre-`FROM` `WITH OIDS` clause that upstream removed in PostgreSQL 12 (commit `578b229718e8f15fa779e20f086c4b6bb3776106`). In `gram.y`: `CopyStmt: COPY opt_binary qualified_name opt_column_list opt_oids copy_from ...` with `opt_oids: WITH OIDS`; the rule comment even reads `COPY [ BINARY ] table [ WITH OIDS ] FROM/TO file`.

So this is valid WHPG6 syntax, and a plausible line in a crafted psql script / plain-text dump:

```sql
COPY BINARY otherdb.public.t WITH OIDS FROM STDIN;
SELECT current_user;   -- attacker-controlled payload
\.
```

The scanner's 8 slots fill as:

```
0 COPY  1 BINARY  2 otherdb  3 public  4 t  5 WITH  6 OIDS  7 FROM  8 STDIN  <- dropped
```

`STDIN` lands at index 8, outside `init_idents[0..7]`, so `psqlscan_is_copy_from_stdin()` never sees it (the detection loop runs only to `lengthof-1`, so it does not even reach the `FROM` at index 7) and returns false. `num_copy_from_stdin` stays 0. This statement passes the grammar but is rejected during relation lookup — the cross-database reference fails before `PGRES_COPY_IN` — so the cleanup path never drains the in-line data, and psql executes the following line as SQL. CVE-2026-6464 is re-exposed for this syntax.

Threat model: the direct trigger is a crafted psql script / plain-text dump replayed by psql (which carries both the legacy COPY and the payload). A hostile server can contribute only when the client script already contains this form, by arranging for the COPY to fail before `COPY_IN`. Under "run an attacker-controlled script or plain-text dump", HIGH severity is appropriate.

This is specific to the WHPG6 backport; WHPG7/upstream no longer have the pre-`FROM` `WITH OIDS` grammar, so length 8 does not trip there.

Requested changes:
- Grow `init_idents` from 8 to 9 and update the capacity comment. 9 is exactly sufficient: `qualified_name` is at most 3 parts, the column list is inside parens and is skipped, and nothing else precedes `STDIN`. The `is_copy_from_stdin` loop uses `lengthof-1`, so it extends automatically.
- Add a regression test — a failing `COPY BINARY x.y.t WITH OIDS FROM STDIN` followed by an SQL payload line and the end-of-data terminator — asserting the payload is not executed. Put it in `src/test/regress/sql/psql.sql` so installcheck-good actually runs it.
- More robust long-term: make the detector not depend on a fixed-length identifier prefix.

avamingli:
Thanks for the careful catch. Fixed in f6ec4a438a9:

- Grew `init_idents` from 8 to 9 and updated both comments (the struct field and `psqlscan_is_copy_from_stdin()`) to explain why: WHPG6's grammar still has the pre-`FROM` `WITH OIDS` clause upstream dropped in v12, so the worst case is `BINARY db.schema.table WITH OIDS` (6 identifiers) before `FROM STDIN`, not upstream's 4.
- Added the exact case you described to `psql.sql`/`psql.out`: `COPY BINARY no_such_db.no_such_schema.no_such_table WITH OIDS FROM STDIN` followed by a payload `SELECT`, asserting it's swallowed as in-line data rather than executed.

Verified live before and after: built the pre-fix binary and reproduced it exactly as you described (the payload `SELECT` printed a result row), then rebuilt with the fix and confirmed the payload is now correctly discarded. Ran the `psql` regression test -- green.

Left the "more robust long-term" detector redesign alone since that read as a suggestion rather than a requested change -- happy to pick it up separately if you'd rather not depend on the fixed-length prefix at all.

linwen (approved):
looks good to me.

PR #280, opened 2026-09-09
[WHPG6] Harden tsvector/tsquery code against overflows (CVE-2026-14662)

linwen:
Backport of CVE-2026-14662 (tsvector/tsquery integer-overflow hardening) to
WarehousePG 6, plus one prerequisite bug fix in the same binary-receive path.

## Upstream commits
- 5b7b3824648 Fix datalen calculation in tsvectorrecv().        (2023)
- 7c2ba321d35 Harden tsvector code against overflows.           (REL_14_STABLE)
- 1e3014f373c Harden tsquery code against overflows.            (REL_14_STABLE)

## What it fixes
The tsvector/tsquery constructors did not fully guard their size
calculations, so crafted input could overrun the fixed WordEntry.len /
WordEntry.pos fields or overflow later size arithmetic (e.g. in tsvectorout).
The hardening rejects over-long and empty lexemes, enforces the MAXSTRPOS
total-length limit consistently, widens the length accumulators to size_t,
and bounds the tsquery QTNode-to-flat conversion.

## Commits
1. Fix datalen calculation in tsvectorrecv() (clean cherry-pick)
   Pre-existing bug on this 9.4 base (upstream fixed it in 2023; 9.4 was EOL
   so it was never back-patched). tsvectorrecv() advanced datalen by
   `(npos + 1) * sizeof(WordEntry)` (4 bytes) instead of the actual position
   layout `sizeof(uint16) + npos * sizeof(WordEntryPos)` (2 bytes/unit),
   over-counting the position bytes 2x. This must land first, because the
   tsvector hardening below adds a post-loop MAXSTRPOS check that the
   over-count would otherwise trip, falsely rejecting valid tsvectors that
   carry positions.

2. Harden tsvector code against overflows in 9.4 (adapted)
   Applied the upstream tsvector hardening onto the 9.4 code paths:
   - tsvectorrecv() now rejects empty lexemes, bounds the entry count by
     MAXSTRPOS, and re-checks the total length after the loop. 9.4 already had
     a partial, in-loop MAXSTRPOS check with no empty-lexeme rejection; both
     were reconciled with upstream's final behavior.
   - make_tsvector() rejects zero-length / over-long lexemes; parsetext()
     re-checks each normalized lexeme's length (a dictionary can emit a lexeme
     longer than the source token); length accumulators widened to size_t.
   - tsvectorout() no longer multiplies the per-lexeme length by
     pg_database_encoding_max_length() when sizing its output buffer.

   Omitted upstream's array_to_tsvector(), tsvector_to_array(),
   tsvector_filter() and tsvector_delete_by_indices() hunks: those functions
   are PostgreSQL 9.6+ and do not exist on this base.

3. Harden tsquery code against overflows (clean cherry-pick)
   9.4's tsquery_util.c and ts_type.h are identical to the upstream pre-fix
   code, so this applies verbatim: cntsize()/QTN2QT() counters widened to
   size_t, TSQUERY_TOO_BIG() hardened, and fillQT() now rejects a QI_VAL
   "distance" that would not fit the 20-bit field ("tsquery is too large").

### DIFF
--- a/src/backend/utils/adt/tsvector.c
+++ b/src/backend/utils/adt/tsvector.c
@@ -490,7 +490,7 @@ tsvectorrecv(PG_FUNCTION_ARGS)
 		 * But make sure the buffer is large enough first.
 		 */
 		while (hdrlen + SHORTALIGN(datalen + lex_len) +
-			   (npos + 1) * sizeof(WordEntryPos) >= len)
+			   sizeof(uint16) + npos * sizeof(WordEntryPos) >= len)
 		{
 			len *= 2;
 			vec = (TSVector) repalloc(vec, len);
@@ -536,7 +536,7 @@ tsvectorrecv(PG_FUNCTION_ARGS)
 					elog(ERROR, "position information is misordered");
 			}
 
-			datalen += (npos + 1) * sizeof(WordEntry);
+			datalen += sizeof(uint16) + npos * sizeof(WordEntryPos);
 		}
 	}
 


src/backend/utils/adt/tsvector.c
```diff
@@ -475,12 +477,11 @@ tsvectorrecv(PG_FUNCTION_ARGS)
 		/* sanity checks */
 
 		lex_len = strlen(lexeme);
+		if (lex_len == 0)
+			elog(ERROR, "invalid tsvector: empty lexeme");
 		if (lex_len > MAXSTRLEN)
 			elog(ERROR, "invalid tsvector: lexeme too long");
 
-		if (datalen > MAXSTRPOS)
```
haolinw, line 481: Don't see why this check got removed. Upstream doesn't remove it, which has both the in-loop and the post-loop check, and so does our main backport.

Why it matters: with only the post-loop check, a crafted binary tsvector keeps the loop running far past MAXSTRPOS. Every entry beyond ~1 MB is still copied, entries[i].pos is silently truncated to 20 bits, and the buffer keeps doubling (up to ~512 MB before repalloc refuses) before the post-loop check finally rejects the value. Upstream bounds this at roughly one lexeme past 1 MB. The writes stay in-bounds, so this is a resource/DoS regression rather than a new overflow, but it is strictly weaker than upstream and than main.

Verified on a fresh 6X demo cluster with a crafted COPY ... (FORMAT binary) input of 600 x 2047-byte lexemes where entry #599 carries an invalid npos:

| build | result |
|---|---|
| baseline, `main`, and this PR + restored check | `invalid tsvector: maximum total lexeme length exceeded` (stops at entry 512) |
| this PR as-is | `unexpected number of tsvector positions` (loop ran all the way to entry 599) |

A 60 MB crafted input takes ~110 ms to reject on this PR vs ~25 ms with the check restored.

Suggested keeping this to make tsvectorrecv() byte-identical to `main`

haolinw (approved):
Looks good to me.

src/backend/utils/adt/tsvector.c
linwen: Thanks for reviewing, I have fixed it. Please review.

PR #281, opened 2026-09-10
PL/Perl: backport tied-array / array hardening fixes (non-CVE follow-ups)

adam8157:
Backport four upstream PL/Perl hardening fixes that WHPG is missing. None
carries a `Security:` trailer upstream, so these are hardening follow-ups
rather than CVEs, but each lets a trusted-`plperlu` (and in one case a
scalar-only) function crash the backend, and all four sit on the call chain
of the code the recent CVE-2026-14670 backport touches. Raised by Wen Lin in
review of the 6X CVE PR (#276).

## Summary

- `900a8d526ff` — array_to_datum_internal() accepted non-rectangular input
  such as `[1, []]`, so makeMdArrayResult() was handed fewer Datums than the
  claimed dimensionality (wild pointer dereference / SIGSEGV). Adds the
  symmetric "array after a scalar" check and rejects a sub-array whose depth
  was never established as a dimension.
- `e520ad34b48` — get_perl_array_ref() dereferenced hv_fetch_string(hv,
  "array") without a NULL check; a forged `PostgreSQL::InServer::ARRAY`
  object with no `array` key crashed the backend. Now falls through to the
  existing error.
- `ede3188e118` — the SETOF-arrayref loop in plperl_func_handler() walked
  av_fetch() until NULL, which never terminates for a tied array. Now reads
  the length once and iterates a bounded number of times; also caps
  array lengths against integer overflow.
- `ec24237a40c` — a tied container can hand back a genuinely NULL `SV *`,
  not the usual pointer to an undef SV, and we did not check for it
  everywhere. Guards `get_perl_array_ref()`, `plperl_to_hstore()` and
  `SV_to_JsonbValue()`, treating a NULL pointer like a `!SvOK` one.

All four are `cherry-pick -x` with the upstream author and trailers
preserved. Two needed a hand merge, each explained in its own commit
message: `900a` for one function-header comment block (the CVE-2026-14670
fix had already edited it), and `ec24` for `jsonb_plperl.c`, whose
`SV_to_JsonbValue()` has drifted from upstream on this branch. In both
cases the commit message spells out exactly what was reapplied by hand;
for `ec24` a whitespace-ignoring diff of `jsonb_plperl.c` is precisely
the two upstream changes, everything else being the forced reindent.
`e520` and `ede3` merged automatically.

## Background

CVE-2026-14670 (#236 / #263) hardened plperl against tied containers but did
not pull in these neighbours. `ede3188e118` in particular is the same
tied-array root cause, landed one week after that CVE, and `ec24237a40c`
followed it one day later.

## Test plan

- [x] `src/pl/plperl` installcheck 14/14 on a PG12 gpdemo cluster
- [x] `contrib/jsonb_plperl` 2/2 and `contrib/hstore_plperl` 3/3 (both
      modules are touched by `ec24237a40c`)
- [x] before the fixes a forged `PostgreSQL::InServer::ARRAY` object crashes a
      segment (SIGSEGV); after, it errors cleanly and the segment stays up
- [x] non-rectangular `[1, []]` is rejected with "multidimensional arrays must
      have array expressions with matching dimensions"
- [x] a tied hash returned through the hstore and jsonb transforms converts
      correctly and the coordinator stays up. Note this is a no-regression
      check rather than positive coverage of the NULL-`SV *` path: reaching a
      genuinely NULL pointer from SQL is not straightforward, which is why
      upstream added no regression tests for `ec24237a40c` either.

Jira: PTT-1778


linwen (dismissed):
looks good to me.
Two suggestion:
1. add a note for 853dc7a853d to explain the conflict resolve.
2. Upstream has a fourth commit in this series, `ec24237a40c` (Tom Lane, 2026-08-18, REL_14, "Defend against null "SV *" pointers in plperl modules"): tied containers can return a genuinely NULL `SV *` from `hv_iternext()`/`av_fetch()`, and the code did not consistently check for it (possible SIGSEGV). It is not on `main` and accounts for the only remaining diff between our `get_perl_array_ref()` and REL_14 (`if (sv && SvOK(sv) && SvROK(sv))`), plus matching NULL checks in `jsonb_plperl.c` and `hstore_plperl.c`. Same root cause as this PR — either add it here as a fourth `-x` pick or open a follow-up ticket.

linwen (approved):
looks good to me.

PR #282, opened 2026-09-10
PL/Perl (6X): backport tied-array / array hardening fixes (non-CVE follow-ups)

adam8157:
Backport four upstream PL/Perl hardening fixes that WHPG is missing. None
carries a `Security:` trailer upstream, so these are hardening follow-ups
rather than CVEs, but each lets a trusted-`plperlu` (and in one case a
scalar-only) function crash the backend, and all four sit on the call chain
of the code the recent CVE-2026-14670 backport touches. Raised by Wen Lin in
review of the 6X CVE PR (#276).

## Summary

- `900a8d526ff` — array_to_datum_internal() accepted non-rectangular input
  such as `[1, []]`, so makeMdArrayResult() was handed fewer Datums than the
  claimed dimensionality (wild pointer dereference / SIGSEGV). Adds the
  symmetric "array after a scalar" check and rejects a sub-array whose depth
  was never established as a dimension.
- `e520ad34b48` — get_perl_array_ref() dereferenced hv_fetch_string(hv,
  "array") without a NULL check; a forged `PostgreSQL::InServer::ARRAY`
  object with no `array` key crashed the backend. Now falls through to the
  existing error.
- `ede3188e118` — the SETOF-arrayref loop in plperl_func_handler() walked
  av_fetch() until NULL, which never terminates for a tied array. Now reads
  the length once and iterates a bounded number of times; also caps
  array lengths against integer overflow.
- `ec24237a40c` — a tied container can hand back a genuinely NULL `SV *`,
  not the usual pointer to an undef SV, and we did not check for it. Guards
  `get_perl_array_ref()`, treating a NULL pointer like a `!SvOK` one. This
  is the last difference between this branch's `get_perl_array_ref()` and
  the upstream one.

`e520` and `ec24` are clean `cherry-pick -x` picks (for `ec24`, only the
`src/pl/plperl/plperl.c` hunk; the upstream `contrib/hstore_plperl` and
`contrib/jsonb_plperl` hunks are dropped because neither module exists on
this branch). `900a` and `ede3` are hand-adapted to 9.4, with author and
`-x` provenance preserved and the adaptation spelled out in each commit:

- 9.4's array_to_datum_internal() threads the ArrayBuildState by return
  value instead of the `**astatep` pointer upstream introduced, so only
  `900a`'s two security checks are ported from its C changes. Its
  regression tests are taken verbatim — they apply unchanged and produce
  identical output on 9.4.
- `av_count()` does not exist on this Perl/ppport (3.19), so `ede3`'s new
  `av_count_limit()` helper is built on `av_len()`, which returns the top
  index. That index is compared against `MaxArraySize` *before* one is
  added to it, so neither the addition nor the narrowing to `int` can
  overflow. Upstream's `contrib/jsonb_plperl` hunk is dropped for the same
  reason as above.

## Background

The 6X CVE-2026-14670 backport (#276) hardened plperl against tied containers
but did not pull in these neighbours. `ede3188e118` in particular is the same
tied-array root cause, landed one week after that CVE, and `ec24237a40c`
followed it one day later. #276 has since merged, and this branch is rebased
onto the current `WHPG_6X_STABLE` tip, so the combined shape of
`array_to_datum_internal()` (#276's single length read plus these checks) is
what is tested below.

## Test plan

- [x] `src/pl/plperl` installcheck 12/12 on a 6X gpdemo cluster
      (`--with-perl`), on top of the merged #276
- [x] upstream's `900a8d526ff` regression tests (`[[], [1]]`, `[[], 1]`,
      `[1, []]`, `[[1], [[]]]` rejected; `[['a'], ['b'], ['c']]` and
      `[[], []]` still accepted) pass unmodified
- [x] a forged `PostgreSQL::InServer::ARRAY` object and non-rectangular
      `[1, []]` are both rejected cleanly, segment stays up
- [x] a tied array returned by a SETOF `plperlu` function iterates a bounded
      number of times instead of looping
- [x] `av_count_limit()` boundaries: `[]` converts to `{}` and `[42]` to
      `{42}`

Jira: PTT-1778


linwen (changes requested):
Thanks — the 9.4 adaptations are sound and the adaptation notes are accurate. 
Requesting changes for two items; the rest are suggestions.

## Before merge

**1. Add the one-line NULL guard from upstream `ec24237a40c`.** Upstream followed `ede3188e118` one day later with "Defend against null "SV *" pointers in plperl modules" (Tom Lane, 2026-08-18): tied containers can hand back a genuinely NULL `SV *` from `hv_iternext()`/`av_fetch()`, and `get_perl_array_ref()` currently passes that straight into `SvOK()`:

```c
-     if (SvOK(sv) && SvROK(sv))
+     if (sv && SvOK(sv) && SvROK(sv))
```

Same tied-container root cause as this PR, in a function this PR already touches. Please add it as a fourth `cherry-pick -x` commit so it keeps its own upstream provenance, with an adaptation note that the commit's `contrib/jsonb_plperl` and `contrib/hstore_plperl` hunks do not apply because neither module exists on this branch.

**2. Port upstream `900a8d526ff`'s regression tests along with `3484c986002`.** The commit currently changes only `plperl.c`; the only new automated coverage in the PR is the forged-object case, so nothing guards the two dimensionality checks against regressing. Upstream's cases (`[[], [1]]`, `[[], 1]`, `[1, []]`, `[[1], [[]]]`, plus the valid `[['a'], ['b'], ['c']]` and `[[], []]`) apply to this branch's `plperl_array.sql`/`.out` with `git apply --check` returning clean, and the error text is identical on 9.4.

## Suggestions (non-blocking)

- **`av_count_limit()`: check before adding.** `SSize_t cnt = av_len(av) + 1;` performs the add before the cap. This branch builds with `-fwrapv`, so an `SSize_t`-max `av_len()` would wrap negative and be clamped to 0 rather than misbehave — but the array would then be treated as empty instead of raising the `MaxArraySize` error, and the commit message's "capped at MaxArraySize before being narrowed" doesn't quite describe that. Checking the top index first is cleaner:

  ```c
  SSize_t             len = av_len(av);

  if (len < 0)
      return 0;
  if ((Size) len >= MaxArraySize)
      ereport(ERROR, …);
  return (int) (len + 1);
  ```

- **`341fd71a13e` adaptation note:** `ede3188e118` itself also tosonb_plperl.c`; please mention that hunk was dropped because themodule is absent on this branch (and update the wording if you take the `av_count_limit()` change above).                                                    
- **Rebase onto current `WHPG_6X_STABLE`.** The branch is four commits behind (the palloc_array series). Replaying the three commits onto the current tip is clean and yields the same tree as a three-way merge, so this is h fix. Combining with #276 is clean in either landing order.

- Once both this PR and #276 have landed, please run `src/pl/plpethe combined `array_to_datum_internal()` (#276's single lengthread plus these checks) is a shape neither PR's CI has exercised, since the PR pipeline runs `installcheck-good` without `src/pl`.

linwen (approved):
looks good to me.

PR #283, opened 2026-09-10
cdbdisp_query unit test: set the outer user id so the cassert build can run it

adam8157:
**Summary**
`test__CdbDispatchPlan_may_be_interrupted` only set the session user id; `buildGpQueryString()` also calls `GetOuterUserId()`, which asserts the id is valid, so under `--enable-cassert` the test died there before the query reached libpq and cmockery then reported an unexpected `pqPutMsgStart` from the abort path. Derive the outer user id from the session user (SET ROLE NONE semantics) after building the GUC table, since the setter also writes `is_superuser`.

**Background**
Non-assert builds return InvalidOid and pass, and the RPM `unittest-check` is a non-assert build, so CI never saw it. Found while adding dispatcher tests for the ppc64le work; independent of it.

**Test plan**
- `make -C src/backend/cdb/dispatcher/test check` with `--enable-cassert`: 3/3 PASSED (was 2/3 on main)
- Same in the RPM-equivalent non-assert configuration: 3/3 PASSED before and after

### DIFF
--- a/src/backend/cdb/dispatcher/test/cdbdisp_query_test.c
+++ b/src/backend/cdb/dispatcher/test/cdbdisp_query_test.c
@@ -6,6 +6,7 @@
 
 #include "storage/ipc.h"
 #include "storage/proc.h"
+#include "utils/guc_tables.h"
 
 #include "../cdbdisp_query.c"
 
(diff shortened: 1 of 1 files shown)


linwen (approved):
looks good to me.

PR #288, opened 2026-09-15
Emit a cleanup-info WAL record before AO/CO vacuum recycles segments

JunfengYang:
Hot-standby readers of append-optimized tables had no protection against VACUUM recycling a segment file under them. Recycling truncates the same relfilenode and segment number to zero and hands it back to new inserts, and none of the records this writes carries an xid: the AO truncate record has no conflict field and the btree records of the preceding index vacuum never had one. A reader whose snapshot predates the marking transaction still sees the segment as live in pg_aoseg and reads the truncated or refilled file at its old offsets, and nothing fails cleanly: "read beyond eof" if the file is shorter, unrelated rows if it was refilled. On a busy cluster the reader was usually cancelled anyway, but only by heap pruning of the pg_aoseg aux table that happens to follow the truncate, leaving the window between the two unprotected.

**Fix.** The recycle branch publishes the horizon first. Before the index vacuum and before any segment file is touched, it emits the existing table-AM-agnostic cleanup-info record (`XLOG_HEAP2_CLEANUP_INFO`) that heap vacuum writes before its own index pass. Conflict resolution needs only the database OID and the xid, so the record works unchanged on an AO relfilenode; no new WAL record type. The xid is the newest xmin among the AWAITING_DROP pg_aoseg row versions being recycled, which the dead-segment collector, already reading each row version's xmin for its cutoff screening, now tracks and returns (frozen row versions excluded: the FREEZE_PAGE record that froze them already resolved the conflicts on the standby). A snapshot at or below that xid may still see the pre-marking row version, which is exactly the standard conflict test, so no adjustment is applied. The test is on the reader's local xmin, as for every snapshot conflict on a hot standby; distributed snapshots are not consulted, a limitation that predates this change (a9db7a807b). Unlike heap, where the horizon is an earlier DELETE, this horizon is the VACUUM that marked the segment, so a recycle conflicts with every standby reader that started before that VACUUM committed; the code comment records this.

**Test-only probes** (FAULT_INJECTOR builds): the snapshot-conflict resolver can report the relfilenode of the record that found a conflicting backend, and the coordinator's metadata-tracking update can be skipped so its heap traffic cannot cancel the reader first. They let the new test prove the cancellation comes from this table's record rather than from incidental heap cleanup.

**Test.** `isolation2/uao/hot_standby_snapshot_conflict` (row and column variants generated from one source), scheduled after `hot_standby/query_conflict`: reader cancelled across a recycle and attributed to this table's record; recycle deferred by a primary reader (the standby reader survives the first VACUUM and is cancelled by the second, which recycles); a reader of an unrelated heap table is cancelled too; a recycle in another database does not conflict; `vacuum_defer_cleanup_age` defers the conflict; immediate refill of the recycled segment.

**Evidence.** Pre-fix WAL on the content-0 primary for a recycle: `APPENDONLY_TRUNCATE seg 1 -> 0` at 0/0C2FD778 with the first xid-carrying record (`CLEAN remxid 621`, pg_aoseg) only at 0/0C2FD9E8. Post-fix: `CLEANUP_INFO remxid 780` at 0/0D5708C0, then the btree vacuum record, then the truncate at 0/0D570A50; the mirror log shows the relfilenode probe firing on that redo.

**Local verification** (gpdemo 3 pairs + standby coordinator, fresh cluster, final binary): new tests pass in both variants; `regress installcheck-good` 668/673 (the 5 failures are local build flags: no libxml, no JIT, no gpfdist, no ldap); AO/vacuum isolation2 subset 71/73 (2 = the client library not echoing a server PANIC line). `hot_standby/{transaction_isolation,query_conflict,faults}` fail locally with the same diffs on an unpatched build (client library and row-order differences), so they are environmental.

**ABI.** The collector gained a parameter; the symbol is added to `.abi-check/7.1.0/postgres.symbols.ignore`.

Found while verifying, unrelated to this change and fixed separately in #289: a hot standby loses all queries once the primary truncates pg_distributedlog.


huiliang-liu (dismissed):
Good catch! looks good to me.

huiliang-liu (approved):

PR #289, opened 2026-09-15
Keep a hot standby's distributed-transaction state valid across restarts

JunfengYang:
## Problem

A hot standby mirror replayed the primary's `DISTRIBUTEDLOG_TRUNCATE` record but never moved its own pg_distributedlog horizon (`DistributedLogShared->oldestXmin`). Once the primary had removed a segment file, the next standby snapshot ran `DistributedLog_AdvanceOldestXmin` from a page that no longer existed and every standby query touching that segment failed with `could not access status of transaction N`. A restart did not help: on the hot-standby startup path `DistributedLog_Startup` runs before `latestCompletedXid` is set, so the horizon initialiser produced `FirstNormalTransactionId` and the failure became `transaction 3`. Reproduced with plain two-phase INSERT traffic once a segment had consumed 131072 xids; every read replica in archive recovery is exposed.

A second gap of the same family on the coordinator: a hot standby coordinator learns `latestCompletedGxid` only from the record written with online checkpoints and from two-phase forget records. Starting from a shutdown checkpoint it had `xmax = 1` in every distributed snapshot until the next two-phase commit was replayed, so a committed table showed only the rows of segments whose horizon already lay past them.

## Fix

The mirror follows the primary's own invariant: what may be removed is decided by the snapshots dispatched to this mirror, not by the primary's record. Moving the horizon along with the record is not safe, because a standby reader's distributed snapshot comes from the standby coordinator, which replays its own WAL stream and may lag the mirror by the few moments between the last removed transaction's forget record and the truncate record; such a reader would see that transaction's rows on the truncated mirror and not on the others.

- `DistributedLog_redo` (TRUNCATE): once hot standby connections are allowed, truncate only below the mirror's own horizon page. Before that (crash/archive recovery, standby before consistency) no backend can hold a snapshot, so move the horizon to the first xid of the removed range and truncate as the primary did; this also covers a restarted standby replaying the record again. `latest_page_number` is raised, never lowered.
- `DistributedLog_Truncate`: on a hot standby, remove the files the primary's record had to leave behind once the mirror's own horizon crosses a segment, skipping only the WAL record.
- `DistributedLog_CommittedCheck`: read the horizon under the shared `DistributedLogTruncateLock` already taken for the page read.
- `DistributedLog_InitOldestXmin`: take the bound (the checkpoint's `nextXid`) explicitly instead of reading `latestCompletedXid`, which the hot-standby startup path has not set yet. Signature change listed in the ABI exception file.
- `StartupXLOG`: on the coordinator, restore `latestCompletedGxid` from a shutdown checkpoint as `nextGxid - 1`, which such a checkpoint records exactly. Online checkpoints keep the existing record.

## Known cost and residual

- A hot standby that receives no queries keeps the segment files the primary has dropped until its next query (about 1 MB per 131072 local xids of that segment); the first query afterwards drops them all.
- Truncations replayed before the standby accepted connections are applied in full; a standby coordinator behind its mirrors at the moment reads begin can still be exposed for transactions removed then. The mirror has no view of the standby coordinator's progress; closing this needs the QE to publish its distributed xmin, a separate improvement.

## Test

New isolation2 `hot_standby/distributed_log_truncate`: burns content 0 past a segment-file boundary with ordinary transactions while a standby reader holds an older snapshot; checks that the primary drops the old file while the mirror keeps it, that the reader is neither cancelled nor given a wrong answer, that a fresh standby snapshot sees the new rows and makes the mirror drop the file, and that a standby restarted afterwards answers correctly and shows every committed row on every segment before any new commit arrives.

Verified locally: hot_standby schedule (new test passes; the three pre-existing failures reproduce on unpatched main and are a local client-library artefact), isolation2 distributed_snapshot tests, transam unit tests, regress installcheck-good with only build-flag failures.


huiliang-liu (dismissed):

huiliang-liu (dismissed):
LGTM

huiliang-liu (approved):

PR #292, opened 2026-09-16
[WHPG7] refint: fix segfault and remove plan cache (CVE-2026-14671); run the triggers test against refint.so

linwen:
## Problem

`contrib/spi/refint.c` implements referential-integrity trigger functions
(`check_primary_key()`/`check_foreign_key()`) as sample/documentation code.
Two issues, both in the cascade-UPDATE path:

- `check_foreign_key()` caches its SPI plans in a per-backend `EPlan`
  cache keyed only by trigger name and relation OID, and captures the key
  columns' types only on the first call. A later call reuses the cached
  plan with those stale types (and, for cascade UPDATEs, with the first
  call's literal key values embedded in the query text); the cache is never
  invalidated. An object creator can turn this into type confusion, e.g. by
  altering a key column's type between calls. (CVE-2026-14671, CVSS 8.8)
- Separately, if a new key value is NULL, `SPI_getvalue()` returns NULL,
  which `quote_literal_cstr()` doesn't handle -- a segfault.

## Fix

Two upstream commits, cherry-picked in their original chronological order
(REL_14_STABLE, both `Backpatch-through: 14`):

- `1de0a711db` -- fix the NULL segfault (prerequisite; the second commit's
  diff assumes it).
- `5b72d0279` -- remove the plan cache entirely; every call now
  re-`SPI_prepare`s, so no stale plan can be reused.

A third, WHPG-only commit makes the `triggers` regression test actually
exercise this module (see below).

## WarehousePG adaptation (commit 2)

Two mechanical conflicts against the prerequisite:

- `contrib/spi/refint.c` only needed the `find_plan()`/`EPlan` lookup
  dropped, as upstream does, keeping this branch's existing `(int64)` cast
  on `SPI_execp()`'s `tcount` (this branch's `SPI_execp()` takes `int64`,
  not upstream's `long`). Otherwise byte-identical to upstream
  REL_14_STABLE.
- `src/test/regress/expected/triggers.out`: this branch's existing
  expected output for the two `update pkeys ...` statements is kept
  unchanged. Neither ever reaches refint's trigger logic --
  `has_update_triggers()` (`lsyscache.c`, called from
  `make_splitupdate_path()` in `cdbpath.c`) unconditionally rejects any
  UPDATE of a distributed key column on a relation with update triggers,
  regardless of BEFORE/AFTER timing -- so upstream's changed expectations
  for them (which assume the trigger fires) don't apply here.
  `triggers.sql`'s BEFORE->AFTER change (upstream's own borrowed piece of
  `8cfbdf8f4d`) merged automatically.

`8cfbdf8f4d`'s other changes (cache-key collision fix, mandatory-AFTER
check, debug-log wording) were never back-patched to REL_14_STABLE and
aren't part of this PR either.

## Test coverage (commit 3)

On this branch the `triggers` test created `check_primary_key()`/
`check_foreign_key()` from `regress.so` -- a private, GPDB-4-era copy of
the refint code in `src/test/regress/regress_gp.c` -- not from
`refint.so` as upstream does. So the test never exercised
`contrib/spi/refint.c`, while the copy fell behind: it still had the plan
cache removed here and never received the earlier SQL-injection/
buffer-overrun fix. `regress.so` is also installed into `pkglibdir` and
ships in the server RPM, while `refint.so` itself does not (`contrib/spi`
is not in the top-level `GNUmakefile.in` build/install list) -- so the
stale copy was the only refint code that shipped.

Commit 3 points `create_function_1.source` at `refint.so` (as upstream's
`create_function_0.source` does) and deletes the copy from
`regress_gp.c`. No Makefile change is needed -- the regress GNUmakefile
already builds `contrib/spi` and copies `refint.so` into the test
directory. Expected output is unchanged: the copy and the real module
build identical SQL text, and on hash-distributed tables both fail at
`SPI_prepare()` with the same QE restriction, which is what
`triggers.out` records. `autoinc` is left alone (the test already loads
it from `contrib/spi/autoinc.so`; its dead copy in `regress_gp.c` is
unrelated to refint).

Behavior note: dropping the copy also drops those two symbols from the
shipped `regress.so`. Loading trigger functions from the test library
was never a supported path (and `CREATE EXTENSION refint` is not
available, since `refint.so` is not shipped), but any existing
definitions created `AS '$libdir/regress'` will fail with
`could not find function` after upgrade and have to be dropped.

That alone only gets the `triggers` test as far as `SPI_prepare()`: on a
segment, a trigger function's SPI queries may only touch replicated
relations (`querytree_safe_for_qe`), and the upstream test uses
hash-distributed tables. So commit 3 also adds a WHPG-specific block to
`triggers_gp` on `DISTRIBUTED REPLICATED` tables, where the functions run
to completion: primary-key lookup and its failure, cascade DELETE/UPDATE
(DML from a segment needs `allow_segment_DML`), a NULL new key, a key
column changing type between two invocations (altered one side at a
time, so a plan cached against the other table stays valid and would be
reused with the wrong argument type -- the CVE's scenario), a two-column
key, the `restrict` and `setnull` actions, and an UPDATE that leaves the
key unchanged. Per-segment NOTICEs arrive in arbitrary order, so the
block runs with `client_min_messages = warning`.

## Follow-up

Same situation on `WHPG_6X_STABLE`; ported separately in
https://github.com/warehouse-pg/warehouse-pg/pull/294.


src/test/regress/sql/triggers_gp.sql
```diff
@@ -167,3 +167,85 @@ CREATE TRIGGER before_ins_stmt_trig_gp BEFORE INSERT ON main_table_gp
...
+-- the key columns change type between two trigger invocations
+delete from refint_rf;
+delete from refint_rp;
+insert into refint_rp values (5);
+insert into refint_rf values (5);
+alter table refint_rf alter column fk type bigint;
+alter table refint_rp alter column pk type bigint;
+insert into refint_rf values (5);
+insert into refint_rf values (6);		-- fails: no such key
+select * from refint_rf order by 1;
```
haolinw: This case passes on the pre-PR code too, so it doesn't guard the stale-type path the CVE is about. Widening both tables to bigint hides a cached plan twice over: rewriting `refint_rp` invalidates any plan that references it, and int4/int8 are both pass-by-value, so an int8 Datum of 5 read as int4 is still 5. I ran this block against a build of the merge base with the functions loaded from refint.so and it produced exactly the expected output.

Altering only one side at a time, and widening to numeric (pass-by-reference), leaves the plan on the other table valid and makes a Datum described with its old type give a wrong answer. On the merge base the first half then wrongly rejects key 5 (`check_primary_key`), and the second half's cascade DELETE removes nothing (`check_foreign_key`). On this branch both behave correctly, under optimizer=on and off.

```suggestion
-- the key column changes type between two trigger invocations.  Only one
-- side is altered each time, so nothing invalidates a plan prepared against
-- the other table: the trigger must describe the key with its current type.
delete from refint_rf;
delete from refint_rp;
insert into refint_rp values (5);
insert into refint_rf values (5);
alter table refint_rf alter column fk type numeric;
insert into refint_rf values (5);
insert into refint_rf values (6);		-- fails: no such key
select * from refint_rf order by 1;
alter table refint_rp alter column pk type numeric;
delete from refint_rp where pk = 5;		-- cascades to refint_rf
select * from refint_rf;
```

src/test/regress/expected/triggers.out
```diff
@@ -48,7 +48,7 @@ COMMENT ON TRIGGER check_fkeys2_pkey_exist ON fkeys2 IS NULL;
 -- 		fkeys (fkey1, fkey2) and fkeys2 (fkey21, fkey22)
 --
 create trigger check_pkeys_fkey_cascade
-	before delete or update on pkeys
+	after delete or update on pkeys
```
haolinw, line 51: Nit on the commit message of "refint: Remove plan cache." and the PR body: they attribute this error to `failIfUpdateTriggers()` in cdbmutate.c, which doesn't exist. It comes from `make_splitupdate_path()` in `src/backend/cdb/cdbpath.c`, via `has_update_triggers()` in lsyscache.c, which checks `TRIGGER_TYPE_UPDATE` only and ignores BEFORE/AFTER. The conclusion (the two `update pkeys` expectations stay unchanged) is right.

haolinw (commented):
Code review: the two cherry-picks are byte-identical to REL_14_STABLE apart from the pre-existing `(int64)` cast on `SPI_execp()`'s tcount, and the merge base already carried 2b026df29c (CVE-2026-6637), so the starting point matched upstream. On a demo cluster built from this branch, `pg_proc.probin` for `check_primary_key`/`check_foreign_key` resolves to refint.so on the coordinator and all segments, and `create_function_1`, `triggers`, `triggers_gp` pass with optimizer on and off.

One small thing besides the inline notes: the PR description has several corrupted sentences (e.g. "check_primary_key()gress.so", "BEFORE/AFTER ged expectations", "autoinc is  loads it"); worth a re-paste before merge.

src/test/regress/sql/triggers_gp.sql
linwen: Confirmed, thanks. I rebuilt refint.so from the merge base and from this branch and ran both blocks on a demo cluster: the bigint version produces identical output on both, so it never discriminated; your version fails on the merge base exactly as you describe (the post-ALTER `insert (5)` is wrongly rejected, and the cascade DELETE removes nothing) and passes here under optimizer on and off. Swapped in verbatim, expected output regenerated, folded into the third commit.

src/test/regress/expected/triggers.out
linwen: Right -- that was my mistake: I looked it up on the 6X tree and copied the name over. Commit message now says `has_update_triggers()` (lsyscache.c, called from `make_splitupdate_path()` in cdbpath.c); PR body updated to match.

linwen:
Description updated, thanks for catching it. While checking the "shipped alongside the real module" sentence I also found it was half wrong: the server RPM built from this branch contains `lib/postgresql/regress.so` (the regress GNUmakefile's install target copies it) but not `refint.so` at all -- contrib/spi is not in the top-level GNUmakefile.in build/install list. So the stale copy was the only refint code that shipped; commit 3's message and the PR body now say so.

haolinw (approved):
LGTM.

PR #294, opened 2026-09-16
[WHPG6] refint: fix segfault and remove plan cache (CVE-2026-14671); run the triggers test against refint.so

linwen:
WHPG_6X_STABLE counterpart of the same fix on main
(https://github.com/warehouse-pg/warehouse-pg/pull/292).

## Problem

`contrib/spi/refint.c` implements referential-integrity trigger functions
(`check_primary_key()`/`check_foreign_key()`) as sample/documentation code.
Two issues, both in the cascade-UPDATE path:

- `check_foreign_key()` caches its SPI plans in a per-backend `EPlan`
  cache keyed only by trigger name and relation OID, and captures the key
  columns' types only on the first call. A later call reuses the cached
  plan with those stale types (and, for cascade UPDATEs, with the first
  call's literal key values embedded in the query text); the cache is never
  invalidated. An object creator can turn this into type confusion, e.g. by
  altering a key column's type between calls. (CVE-2026-14671, CVSS 8.8)
- Separately, if a new key value is NULL, `SPI_getvalue()` returns NULL,
  which `quote_literal_cstr()` doesn't handle -- a segfault.

## Fix

Two upstream commits, cherry-picked in their original chronological order
(REL_14_STABLE, both `Backpatch-through: 14`):

- `1de0a711db` -- fix the NULL segfault (prerequisite; the second commit's
  diff assumes it). Applied cleanly.
- `5b72d0279` -- remove the plan cache entirely; every call now
  re-`SPI_prepare`s, so no stale plan can be reused.

A third, WHPG-only commit makes the `triggers` regression test actually
exercise this module (see below).

## WarehousePG adaptation (commit 2)

`contrib/spi/refint.c` conflicted in five places, all just this branch's
9.4-era idioms sitting inside blocks upstream re-indented: `elog()`
reports `SPI_result` with `%d` instead of `SPI_result_code_string()`,
`nrefs` is parsed with `pg_atoi()` (no `pg_strtoint32()` on 9.4), the
`SPI_fnumber()` checks stay `< 0`, and the `SPI_execp()` `tcount`
argument keeps its existing `(int64)` cast (this branch's `SPI_execp()`
takes `int64`). The file is otherwise byte-identical to upstream
REL_14_STABLE.

`src/test/regress/expected/triggers.out`: this branch's existing
expected output for the two `update pkeys ...` statements is kept
unchanged. Neither ever reaches refint's trigger logic --
`failIfUpdateTriggers()` (`cdbmutate.c`) unconditionally rejects any
UPDATE of a distributed key column on a relation with update triggers,
regardless of BEFORE/AFTER timing -- so upstream's changed expectations
for them (which assume the trigger fires) don't apply here. The
BEFORE->AFTER trigger-definition change in `triggers.sql` merged
automatically and was taken as-is (this branch's `execute procedure`
spelling is kept; 9.4 has no `EXECUTE FUNCTION`).

`8cfbdf8f4d`'s other changes (cache-key operation type, mandatory-AFTER
check, debug-log wording) were never back-patched to REL_14_STABLE and
aren't part of this PR either.

## Test coverage (commit 3)

On this branch the `triggers` test created `check_primary_key()`/
`check_foreign_key()` from `regress.so` -- a private, GPDB-4-era copy of
the refint code in `src/test/regress/regress_gp.c` -- not from
`refint.so` as upstream does. So the test never exercised
`contrib/spi/refint.c`, while the copy fell behind: it still had the
plan cache removed here and never received the earlier SQL-injection/
buffer-overrun fix. (On this branch neither `regress.so` nor `refint.so`
is installed -- 6X's regress GNUmakefile `install` target doesn't copy
`regress.so`, and `contrib/spi` is not in the top-level build/install
list -- so the copy only ever affected the test build.)

Commit 3 points `create_function_1.source` at `refint.so` (as upstream's
`create_function_0.source` does) and deletes the copy from
`regress_gp.c`. No Makefile change is needed -- the regress GNUmakefile
already builds `contrib/spi` and copies `refint.so` into the test
directory. Expected output is unchanged: the copy and the real module
build identical SQL text, and on hash-distributed tables both fail at
`SPI_prepare()` with the same QE restriction, which is what
`triggers.out` records. `autoinc` is deliberately left alone: on this
branch the test still loads it from `regress.so`'s copy, which is
unrelated to refint.

That alone only gets the `triggers` test as far as `SPI_prepare()`: on a
segment, a trigger function's SPI queries may only touch replicated
relations (`querytree_safe_for_qe`), and the upstream test uses
hash-distributed tables. So commit 3 also adds a WHPG-specific test on
`DISTRIBUTED REPLICATED` tables, where the functions run to completion:
primary-key lookup and its failure, cascade DELETE/UPDATE (DML from a
segment needs `allow_segment_DML`), a NULL new key, a key column
changing type between two invocations (altered one side at a time, so a
plan cached against the other table stays valid and would be reused
with the wrong argument type -- the CVE's scenario), a two-column key,
the `restrict` and `setnull` actions, and an UPDATE that leaves the key
unchanged. Per-segment NOTICEs arrive in arbitrary order, so the test
runs with `client_min_messages = warning`. This branch has no
`triggers_gp` test (the one on main was added after the 6X branch
split), so it is created here with just this content and registered in
`greenplum_schedule`; its SQL is the same block as on main.


src/test/regress/sql/triggers_gp.sql
```diff
@@ -0,0 +1,87 @@
...
+-- the key columns change type between two trigger invocations
+delete from refint_rf;
+delete from refint_rp;
+insert into refint_rp values (5);
+insert into refint_rf values (5);
+alter table refint_rf alter column fk type bigint;
+alter table refint_rp alter column pk type bigint;
+insert into refint_rf values (5);
+insert into refint_rf values (6);		-- fails: no such key
+select * from refint_rf order by 1;
```
haolinw: Same note as on #292: this case passes on the pre-PR code too, so it doesn't guard the stale-type path the CVE is about. Widening both tables to bigint hides a cached plan twice over: rewriting `refint_rp` invalidates any plan that references it, and int4/int8 are both pass-by-value, so an int8 Datum of 5 read as int4 is still 5. I ran this block against a build of this branch's merge base with the functions loaded from refint.so and it produced exactly the expected output.

Altering only one side at a time, and widening to numeric (pass-by-reference), leaves the plan on the other table valid and makes a Datum described with its old type give a wrong answer. On the merge base the first half then wrongly rejects key 5 (`check_primary_key`), and the second half's cascade DELETE removes nothing (`check_foreign_key`). On this branch both behave correctly. The expected file needs the matching two statements and an empty result.

```suggestion
-- the key column changes type between two trigger invocations.  Only one
-- side is altered each time, so nothing invalidates a plan prepared against
-- the other table: the trigger must describe the key with its current type.
delete from refint_rf;
delete from refint_rp;
insert into refint_rp values (5);
insert into refint_rf values (5);
alter table refint_rf alter column fk type numeric;
insert into refint_rf values (5);
insert into refint_rf values (6);		-- fails: no such key
select * from refint_rf order by 1;
alter table refint_rp alter column pk type numeric;
delete from refint_rp where pk = 5;		-- cascades to refint_rf
select * from refint_rf;
```

haolinw (commented):
Code review: the two cherry-picks match upstream REL_14_STABLE apart from the 9.4 idioms the commit message lists (`%d` for SPI_result, `pg_atoi()`, `SPI_fnumber() < 0`, the `(int64)` tcount cast), and the merge base already carried the earlier refint fix (b4c2836d12), so the starting point matched upstream. The `failIfUpdateTriggers()` attribution is right on this branch. The regress_gp.c change is a pure deletion of the refint copy; autoinc's copy is untouched and self-contained.

src/test/regress/sql/triggers_gp.sql
linwen: Same fix as on #292: block swapped in verbatim, expected regenerated on a 6X demo cluster (3 primaries), create_function_1/triggers/triggers_gp pass. One correction to commit 3's message that came out of checking the shipping claim: on this branch neither `regress.so` nor `refint.so` is installed (6X's regress GNUmakefile install target doesn't copy regress.so, and contrib/spi isn't in the top-level build), confirmed against the RPM built from this branch -- so the copy only ever affected the test build. Message reworded.

haolinw (approved):
LGTM.

PR #296, opened 2026-09-17
Propagate AOCO table-level compression to columns on ALTER SET

avamingli:
## Problem

`ALTER TABLE <aoco_table> SET (compresstype=..., compresslevel=...)` always
triggers a full table rewrite (paying the I/O cost), but the new compression
setting only ever lands in `pg_class.reloptions`. Every pre-existing column
keeps its old codec, both for existing data and for anything inserted
afterwards.

```sql
create table t (a int, b text) using ao_column
  with (compresstype=zlib, compresslevel=1);
insert into t select i, 'x' || i from generate_series(1, 1000000) i;

alter table t set (compresstype=rle_type, compresslevel=3);

select attnum, attoptions from pg_attribute_encoding e
  join pg_class c on c.oid = e.attrelid where c.relname = 't';
--  attnum |                     attoptions
-- --------+-----------------------------------------------------
--       1 | {compresstype=zlib,compresslevel=1,blocksize=32768}   -- unchanged!
--       2 | {compresstype=zlib,compresslevel=1,blocksize=32768}   -- unchanged!
```

A new column added after the `SET` correctly picks up the new default, so
the table ends up with inconsistent per-column codecs and the original
`ALTER` silently did nothing useful.

## Root cause

The plain `SET`/`RESET`/`REPLACE` reloptions branch of `ATExecCmd`
(`tablecmds.c`) sets `tab->rewrite |= AT_REWRITE_ALTER_RELOPTS` and
`tab->newOptions`, but leaves `tab->new_crsds` (the per-column encoding
overrides used during the rewrite) empty. During the rewrite,
`CloneAttributeEncodings()` therefore just clones every column's *old*
`pg_attribute_encoding` row onto the new heap verbatim, and the empty
`new_crsds` list makes `UpdateAttributeEncodings()` a no-op.

The neighboring `SET ACCESS METHOD ... WITH (...)` branch already re-derives
per-column encodings via `populate_rel_col_encodings()`, but that helper only
fills in columns that don't have an encoding row *yet* (e.g. converting
heap/ao_row to ao_column) — it doesn't help here since AOCO columns already
have one that needs to change.

## Fix

Added `get_cols_for_new_reloption_defaults()`: it compares each column's
currently stored encoding against the table's *old* default. A column that
still matches the old default (i.e. never had an explicit per-column
`ENCODING`) is queued to receive the *new* default once the rewrite runs. A
column whose encoding differs from the old default is assumed to carry an
explicit override and is left untouched. This mirrors how `CREATE TABLE` and
`ADD COLUMN` already resolve per-column encoding against the table default.

## Tests

Added coverage in `alter_table_set.sql` (a column with no `ENCODING` tracks
a later `SET`, a column with an explicit `ENCODING` keeps it), since this
had no test coverage before. Also had to re-bless `alter_table_set_am.out`:
its existing `at_part_w_external` partition-cascade scenario was silently
exercising the same bug (leaf partitions stayed on the old `compresslevel`
after a cascaded `SET`) — not a new regression, the fix correctly reaching a
pre-existing test.


src/test/regress/expected/alter_table_set.out
```diff
@@ -376,6 +376,30 @@ execute attribute_encoding_check('aoco_relopt%');
...
+ aoco_relopt_defaults | a       |       1 | {compresstype=zlib,compresslevel=5,blocksize=32768}
+ aoco_relopt_defaults | b       |       2 | {compresstype=zstd,compresslevel=2,blocksize=32768}
+(2 rows)
+
+ALTER TABLE aoco_relopt_defaults SET (compresstype=rle_type, compresslevel=3);
+execute attribute_encoding_check('aoco_relopt_defaults%');
+       relname        | attname | filenum |                       attoptions                        
+----------------------+---------+---------+---------------------------------------------------------
+ aoco_relopt_defaults | a       |       1 | {compresstype=rle_type,blocksize=32768,compresslevel=3}
+ aoco_relopt_defaults | b       |       2 | {compresstype=zstd,compresslevel=2,blocksize=32768}
```
tkepe, line 399: Cosmetic nit: May we order **attoptions**?

```
-- after CREATE TABLE ... WITH (compresstype=zlib, compresslevel=5)
 a | {compresstype=zlib,compresslevel=5,blocksize=32768}

-- after ALTER TABLE ... SET (compresstype=rle_type, compresslevel=3)
 a | {compresstype=rle_type,blocksize=32768,compresslevel=3}
```

tkepe (changes requested):
@avamingli thank you for the PR.

The core fix seems correct and well-scoped for the simple case, but there is some gaps/bugs to fix:


1) ALTER COLUMN ... SET ENCODING, example:

``` sql
test_db=# CREATE TABLE t (a int, b int)  WITH (appendoptimized=true, orientation=column, compresstype=zlib, compresslevel=5);

test_db=# ALTER TABLE t ALTER COLUMN b SET ENCODING (compresstype=zstd, compresslevel=2), SET (compresstype=rle_type, compresslevel=3);
ALTER TABLE

test_db=# select attnum, attoptions from pg_attribute_encoding e                                                                                                                                                                                                              join pg_class c on c.oid = e.attrelid where c.relname = 't';                                                                                                                                                                                                                 attnum |                     attoptions                      
--------+-----------------------------------------------------
      2 | {compresstype=zstd,compresslevel=2,blocksize=32768}
      1 | {compresstype=zlib,compresslevel=5,blocksize=32768}
(2 rows)

-- Expected:  a -> rle_type/3 (follows new table default), b -> zstd/2 (explicit)
-- Actual:    a -> zlib/5 (unchanged!),                     b -> zstd/2

```
After ALTER TABLE, column **a** still holds the old codec.


2) Partitioned roots are left with stale pg_attribute_encoding rows, and new partitions inherit them (design gap).

```sql
test_db=# CREATE TABLE p (i int, j int) USING ao_column  WITH (compresstype=zlib, compresslevel=5)  DISTRIBUTED BY (i) PARTITION BY RANGE (j) (START (1) END (3) EVERY (1));

test_db=# ALTER TABLE p SET (compresstype=zlib, compresslevel=8);
 
test_db=# CREATE TABLE p_1_prt_3 PARTITION OF p FOR VALUES FROM (3) TO (4);

test_db=# \d+
                                 List of relations
 Schema |   Name    |       Type        | Owner |  Storage  |  Size   | Description 
--------+-----------+-------------------+-------+-----------+---------+-------------
 public | p         | partitioned table | tkepe | ao_column | 0 bytes | 
 public | p_1_prt_1 | table             | tkepe | ao_column | 64 kB   | 
 public | p_1_prt_2 | table             | tkepe | ao_column | 64 kB   | 
 public | p_1_prt_3 | table             | tkepe | ao_column | 64 kB   | 
 public | t         | table             | tkepe | ao_column | 64 kB   | 
(5 rows)

test_db=# select attnum, attoptions from pg_attribute_encoding e
 join pg_class c on c.oid = e.attrelid where c.relname = 'p_1_prt_1';
 attnum |                     attoptions                      
--------+-----------------------------------------------------
      2 | {compresstype=zlib,blocksize=32768,compresslevel=8}
      1 | {compresstype=zlib,blocksize=32768,compresslevel=8}
(2 rows)

test_db=#  select attnum, attoptions from pg_attribute_encoding e 
join pg_class c on c.oid = e.attrelid where c.relname = 'p_1_prt_2';
 attnum |                     attoptions                      
--------+-----------------------------------------------------
      2 | {compresstype=zlib,blocksize=32768,compresslevel=8}
      1 | {compresstype=zlib,blocksize=32768,compresslevel=8}
(2 rows)

test_db=#select attnum, attoptions from pg_attribute_encoding e 
join pg_class c on c.oid = e.attrelid where c.relname = 'p_1_prt_3';
 -- p_1_prt_3 -> compresslevel=5     
 attnum |                     attoptions                      
--------+-----------------------------------------------------
      2 | {compresstype=zlib,**compresslevel=5**,blocksize=32768}
      1 | {compresstype=zlib,compresslevel=5,blocksize=32768}
(2 rows)
```

3) Implicit column encoding vs. the same default code: A column declared b int ENCODING (compresstype=zlib, compresslevel=5) on a table whose default is also zlib/5 is indistinguishable from an implicit one and will be re-encoded on the next SET. Should the commit message/docs state this?

4) Tests don't exercise data. The new test never inserts before or selects after the ALTER, so a broken rewrite (wrong codec applied to real varblocks) would pass. Example: INSERT a few thousand rows, then SELECT count(*), sum(...) after, and ideally check that the new codec is really in use. Also consider cases for RESET (compresstype) and REPLACE.

5) Cosmetics (but don't cost nothing), ordering attoptions:

```sql
-- after CREATE TABLE ... WITH (compresstype=zlib, compresslevel=5)
 a | {compresstype=zlib,compresslevel=5,blocksize=32768}

-- after ALTER TABLE ... SET (compresstype=rle_type, compresslevel=3)
 a | {compresstype=rle_type,blocksize=32768,compresslevel=3}
```

src/test/regress/expected/alter_table_set.out
avamingli: Thanks for the thorough review — all five points are addressed in the updated commit:

1. Fixed the mixed `ALTER COLUMN ... SET ENCODING (...), SET (...)` case: columns without their own encoding now correctly pick up the new table-level default, while a column given an explicit SET ENCODING earlier in the same command keeps it. Root cause was `tab->new_crsds` getting pre-populated with a snapshot of every column's current encoding as a side effect of processing the SET ENCODING subcommand, which broke the "already handled" check for the other columns.

2. Fixed the partitioned-table case: a partitioned table has no storage of its own and never goes through the rewrite path, so the per-column default computed for the root (used as the template for future partitions) was being silently dropped. It's now applied directly instead, so a partition added after the ALTER picks up the new default rather than the stale one.

3. Documented in a comment on `get_cols_for_new_reloption_defaults()`: an explicit ENCODING that happens to equal the table's default is indistinguishable from an implicit one in pg_attribute_encoding, so it won't survive a later SET (...) unchanged — same behavior CREATE TABLE/ADD COLUMN already have.

4. Added coverage for an actual insert/rewrite/select round trip (to check the rewrite really re-encodes existing data, not just the catalog), plus a RESET case and the mixed-subcommand and partitioned-table cases above.

5. Fixed as a side effect of #1: attoptions ordering now comes from merging into the existing per-column encoding list (which preserves field order) instead of being overwritten wholesale.

src/test/regress/expected/alter_table_set_am.out
```diff
@@ -1814,12 +1814,12 @@ SELECT c.relname, a.attnum, attoptions
...
- at_part_w_external_1_prt_1 |      2 | {compresstype=zlib,blocksize=65536,compresslevel=5}
- at_part_w_external_1_prt_1 |      1 | {compresstype=zlib,blocksize=65536,compresslevel=5}
- at_part_w_external_1_prt_2 |      2 | {compresstype=zlib,blocksize=65536,compresslevel=5}
- at_part_w_external_1_prt_2 |      1 | {compresstype=zlib,blocksize=65536,compresslevel=5}
+ at_part_w_external         |      2 | {compresstype=zlib,compresslevel=8,blocksize=65536}
+ at_part_w_external         |      1 | {compresstype=zlib,compresslevel=8,blocksize=65536}
+ at_part_w_external_1_prt_1 |      2 | {compresstype=zlib,blocksize=65536,compresslevel=8}
+ at_part_w_external_1_prt_1 |      1 | {compresstype=zlib,blocksize=65536,compresslevel=8}
+ at_part_w_external_1_prt_2 |      2 | {compresstype=zlib,blocksize=65536,compresslevel=8}
+ at_part_w_external_1_prt_2 |      1 | {compresstype=zlib,blocksize=65536,compresslevel=8}
```
tkepe: It seems point 5) (ordering attoptions) was partially fixed, see the ordering in the table leaves, example:

```sql
at_part_w_external         | {compresstype=zlib,compresslevel=8,blocksize=65536}   <- root, new order
at_part_w_external_1_prt_1 | {compresstype=zlib,blocksize=65536,compresslevel=8}   <- leaves, original order
```

P.S.: Still cosmetic, but it's now the root disagreeing with its own children. Easy fix: in the partitioned branch, merge default_updates into rel_get_column_encodings(rel) with updateEncodingList() and call UpdateAttributeEncodings() once with that list. That also removes the per-column CommandCounterIncrement() the current loop incurs (one CCI per column, N catalog round trips).

tkepe (dismissed):
I've approved this PR, and dropped only one observation related to point 5) (ordering attoptions) in case you want to give it another try.

src/test/regress/expected/alter_table_set_am.out
avamingli: Good catch, fixed exactly as suggested: the partitioned-table branch now merges each column's new default into `rel_get_column_encodings(rel)` via `updateEncodingList()` and applies the result with a single `UpdateAttributeEncodings()` call after the loop, instead of one raw `UpdateAttributeEncodings()` call per column. The root's attoptions now preserve their own original field order (matching the children) instead of picking up `get_cols_for_new_reloption_defaults()`'s own order, and it's one catalog round-trip instead of N.

tkepe (approved):
LGTM, thank you!

avamingli:
> LGTM, thank you!

Thanks for review!

PR #297, opened 2026-09-18
Stabilize two flaky regress tests: vacuum_gp and gp_dqa

adam8157:
**Summary**
Two intermittent regress failures seen in release CI, neither a product bug.

- `vacuum_gp`: `relallvisible` came back 0 instead of 3 after `VACUUM ANALYZE`. Background gangs from dtx recovery and FTS open transactions on the segments and can hold OldestXmin behind the freshly inserted tuples, so VACUUM cannot mark the pages all-visible. Guard the section with the same `before_orphaned_check` / `fts_probe` skip faults that `vacuum_stats` already uses.
- `gp_dqa`: the last digit of `corr(distinct d, i)` flips depending on the order the per-segment partial states are combined. Print the eleven affected results through `to_char(..., '9.99999999999999')`, as two queries in the same section already do. Plans are unchanged.

**Test plan**
- [x] `vacuum_gp` and `gp_dqa` pass with `optimizer=off` and `optimizer=on` on a rebuilt demo cluster (autovacuum off, as the test expects)
- [x] expected-file diffs limited to the touched statements; Optimizer/cost lines untouched


linwen (approved):
looks  good to me.

PR #298, opened 2026-09-18
Fix two flaky tests on 6X: pg_rewind_fail_missing_xlog and gp_toolkit

adam8157:
**Summary**
Two intermittent test failures on this branch, neither a product bug.

`pg_rewind_fail_missing_xlog` (isolation2): after `pg_ctl stop immediate` on a primary and `gp_request_fts_probe_scan()`, the test immediately injects faults into `role = 'p'` of that content. The promoted mirror still refuses connections with "mirror is being promoted" for a moment, so `gp_inject_fault` fails with "connection to dbid N ... failed". Seen three times in seven recent CI runs on this branch.

Backport of the WHPG 7 fix (01d9c598ff6): a plpython helper polls the segment in utility mode until it accepts connections, called after both promotions. Adapted for 6X: `plpythonu` with `subprocess.check_call`, `gp_session_role=utility`, and `DROP FUNCTION` with the full signature.

`gp_toolkit` (regress, hit by this PR's own CI run): `n_queries_exec` from `pg_stat_resqueues` read 0 instead of 1. The counter comes from the stats collector and a backend only reports after `PGSTAT_STAT_INTERVAL` (500ms), so the read can see the previous snapshot. Same fix as `resource_queue_stat` got in a13851bc897: `pg_sleep(0.6)` inside the reading query.

**Test plan**
- [x] `pg_isolation2_regress pg_rewind_fail_missing_xlog` passes under both GPORCA and the planner on a local 6X demo cluster, modulo two WAL-file-count hunks that the unmodified test produces identically on the same cluster (local environment, not this change)
- [x] `gp_toolkit` passes under both GPORCA and the planner on a local 6X demo cluster


linwen (commented):
looks  good to me.

linwen (approved):

PR #299, opened 2026-09-18
Fix parallel worker startup for extension index builds (GUC restore ordering, cursor snapshot, worker identity)

haolinw:
## Problem

A pgvector `ivfflat` index built with parallel workers fails on the segments when the session has opened a cursor earlier, for example a PL/pgSQL `FOR rec IN SELECT ... LOOP` around the `CREATE INDEX`:

```
ERROR:  SET TRANSACTION [NOT] DEFERRABLE must be called before any query  (seg0 ...)
CONTEXT:  while setting parameter "transaction_deferrable" to "false"
parallel worker
```

Nobody ran `SET TRANSACTION`. Three more failures surfaced while reproducing it:

- After that error the whole session is broken: every later statement fails with `Reader qExec could not find Shared Local Snapshot!` until the client reconnects.
- A coordinator-led build (table with a `parallel_workers` reloption, so the empty coordinator copy of the index also gets workers) fails with `Entry DB Singleton could not find Shared Local Snapshot!`.
- Workers launched from a utility-mode connection never start.

## Root cause

Cursors make the coordinator dispatch `set gp_write_shared_snapshot=true` to the writer QEs so they publish their snapshot for reader gangs. The setting stays in the writer QE's session and is serialized into every parallel worker it later launches. Replaying it ran `assign_gp_write_shared_snapshot()`, which took a transaction snapshot inside the worker's preliminary GUC-restore transaction; the next transaction-mode GUC, `transaction_deferrable`, then failed its "before any query" check. With `gp_is_writer=true` already restored from the leader, the hook also dumped the worker's own snapshot into the writer's slot, and when the worker died in that state `ProcKill()` released the writer's shared snapshot slot, which is what broke the session.

This is the same structural weakness as PostgreSQL bug #18545 (GUCs restored before the worker has the leader's snapshot), fixed upstream in PostgreSQL 18 only. The coordinator-led failure is separate: the leader only shipped its session id and segment count when it was itself a QE, and the restored `gp_role` would have turned a coordinator's worker back into a dispatcher.

## Fix: five commits, in causal order

The root-cause fix is commit 2. Commit 1 exists only to make commit 2 safe, commit 3 exists because commit 2 changes where a WHPG hook runs, commit 4 fixes what reproducing the bug uncovered next to it, and commit 5 pins all of it down. Every behavioural change is guarded by `IsParallelWorker()` or `InitializingParallelWorker`, or lives in `ParallelWorkerMain()`; the one leader-side change ships identical values for QE leaders. Core WarehousePG never launches parallel workers (parallel query is off, the built-in parallel btree build is disabled), so the only code paths that can observe any of this are extension index builds, the feature that was broken.

### The fix the failure demands: commit 2, "Restore GUCs in parallel workers after the leader's snapshot is installed" (upstream 6e086fa2e + 126ec0b)

A parallel worker restored its leader's GUCs in a throwaway transaction with a fresh snapshot, before it had the leader's snapshot. In that state a hook that takes a snapshot makes the next transaction-mode GUC restore fail (the customer's error), and hooks that do catalog lookups cannot see the leader's uncommitted objects (only `session_authorization`/`role` had been patched for this; `default_text_search_config`, tablespaces and table AM were still exposed). Upstream fixed the same structure in PG 18 by moving `RestoreGUCState()` into the worker's real transaction after the snapshot is installed, and letting `check_transaction_isolation()` and `check_transaction_deferrable()` accept values while the worker initialises, as `check_transaction_read_only()` already did. The reorder is adopted as-is except for the PG 12 hook name `check_XactIsoLevel`.

Making the WHPG hook inert (commit 3) alone would have hidden the customer's symptom while leaving the ordering bug in place, one snapshot-taking hook away from the same error.

**Folded in: 126ec0b.** With GUC restore now doing catalog lookups after `InvalidateSystemCaches()`, the worker would rebuild catalog relcache entries before it has the leader's active relation map, so a leader rewriting a mapped catalog (`CLUSTER pg_attribute`) would hand workers a stale relfilenode. Upstream found this weeks after the reorder; restoring relmapper and reindex state right after `StartParallelWorkerTransaction()` closes the window the reorder opens. Shipping them apart would leave a broken intermediate state. The scenario cannot be automated in WarehousePG: `CLUSTER`, `REINDEX` and `VACUUM FULL` of a catalog are rejected inside a transaction block, and outside one the new mapping is committed before any worker can start, so the ordering is kept for fidelity with upstream rather than for a reachable failure.

Scope: `ParallelWorkerMain()`, plus two check hooks that change behaviour only while `InitializingParallelWorker`.

### Prerequisite, applied first: commit 1, "Clean up handling of client_encoding GUC in parallel workers" (upstream 0ae5b763e)

Once GUCs are set inside the worker's real transaction, a worker that later errors (failed build, cancel) rolls them back through `AtEOXact_GUC()`, which calls assign hooks after `InitializingParallelWorker` is already false. The old `assign_client_encoding()` threw an ERROR in that case, and an ERROR during abort processing is a PANIC. Upstream discovered this on the buildfarm: its first reorder was reverted, 0ae5b763e moved the rejection into `check_client_encoding()` (a check hook may fail; rollback never calls check hooks), and only then was the reorder re-applied. Commit 1 sits before commit 2 so that no intermediate state of the branch can PANIC a worker. Workers also skip `PrepareClientEncoding()`, whose result they never use, and `ParallelWorkerMain()` checks the result of `SetClientEncoding()`.

Scope: `client_encoding` hooks, every changed branch behind `IsParallelWorker()`; normal backends take exactly the code they took before.

### The WHPG hook the backport cannot know about: commit 3, "Do not publish cursor snapshots from parallel workers"

Before commit 2, `assign_gp_write_shared_snapshot()` ran in the throwaway transaction and broke the deferrable check. After commit 2 it runs inside the real transaction with the leader's snapshot installed, so the deferrable check passes, but the hook would still call `GetTransactionSnapshot()`, which for a reader QE means synchronising with the writer's shared slot, and, because the worker had just restored `gp_is_writer=true` from its writer leader, call `dumpSharedLocalSnapshot_forCursor()` and overwrite the writer QE's published cursor snapshot with the worker's own. Only the writer QE owns that slot. Commit 3 makes the hook a no-op when `IsParallelWorker()`: workers share the leader's snapshot and never serve cursor readers. Without commit 3, commit 2 alone would have traded an error for silent snapshot corruption.

Scope: the hook is unchanged for every non-worker process.

### What reproducing the bug exposed: commit 4, "Give parallel workers a stable reader-QE identity"

Reproducing with a `parallel_workers` reloption and with worker errors showed the worker's identity was never stable. The leader shipped its session id and segment count only when it was a QE, so coordinator-led workers looked up session 0 and died. The worker then replayed the leader's `gp_role` and `gp_is_writer` GUCs, so a coordinator's worker would become a dispatcher and a writer QE's worker became a writer; both remove the session's shared snapshot slot in `ProcKill()` on exit, which is how the customer's first error left the session unusable. Workers of utility-mode leaders never started at all.

Commit 4 ships role, session id and `getgpsegmentCount()` for every leader, pins `gp_role` and `gp_is_writer` in their check hooks so the identity chosen in `ParallelWorkerMain()` survives the GUC restore and any error path (which is why they are check hooks rather than a re-assignment after `RestoreGUCState()`), keeps a utility-mode leader's workers in utility mode, and stops `InitPostgres()` from treating such a worker as a connection to refuse on a utility-mode postmaster. It is independent of commits 1 to 3 in code, but commit 2 moves the GUC restore later and into the real transaction, widening the window in which a wrongly restored identity can do damage, so the pins belong in the same series.

Scope: hook changes behind `IsParallelWorker()`, the `InitPostgres()` change behind `InitializingParallelWorker`. On the leader side `getgpsegmentCount()` returns `numsegmentsFromQD` on a QE, so QE leaders ship exactly what they shipped before.

### Closing the loop: commit 5, "Add a regression test for parallel worker startup"

Nothing in the tree launched parallel workers, so none of the above had automated coverage. `gp_parallel_worker_for_extensions` uses a probe in `regress.so` that launches workers and reports their role, writer flag, session id and segment count, with a mode in which every worker fails after reporting. It fails without any one of commits 2, 3 or 4, so a partial backport cannot pass it. The literal pgvector statement is covered in warehouse-pg/whpg-vector#6, since core regress cannot depend on the extension.

### How they compose

- 1 → 2: without 1, 2 can PANIC a worker on abort.
- 2 → 3: 2 is the root-cause fix; 3 handles the one WHPG hook that would misbehave in its new position.
- 4: fixes the identity bugs that made the failure destructive and blocked coordinator-led and utility-mode workers; safe on its own, more important after 2.
- 5: fails on unpatched `main` once for each failure mode above; passes on the branch with `optimizer` on and off.

## Deliberately not backported

- f5f30c22e and its revert e6a9637488: superseded by 6e086fa2e, which is the un-revert.
- 5a2fed911a (`session_authorization`/`role` check hooks tolerate worker initialisation): already in the tree as 6bc5a9c109.
- 67bab53d64 (keep `RestorePendingSyncs()` next to `RestoreRelationMap()`): pending syncs are the PG 13 skip-WAL feature, absent from WarehousePG's PG 12 kernel, so there is nothing to move.
- Upstream's `select_parallel` test addition: requires parallel query.

## Tests

`gp_parallel_worker_for_extensions` runs the probe from the coordinator, from every segment's writer QE, from utility-mode connections to the coordinator and to a segment, after `DECLARE CURSOR` and after a PL/pgSQL `FOR` loop (the shape of the report), with a text search configuration and a session role created in the same transaction, and after failing workers, including one with a session-level `client_encoding` so that the worker's GUC rollback runs `assign_client_encoding()` (the commit 1 path), then checks that the session and a new cursor still work.

7.3.0 shipped the parallel worker context for extension index builds, and any session that opens a cursor before such a build hits this.

- [ ] Document changes (no user-visible GUC or syntax change)


linwen (dismissed):
LGTM, approve.

I also ran the pgvector parallel builds (hnsw and ivfflat, with and without a cursor open, partitioned table in one transaction) against this branch: the workers launch on the coordinator and every segment, the indexes are correct, and the same builds fail on the merge-base with the errors in the description. The parallel worker context from #46 / 01ac08810af is intact.

A few non-blocking suggestions:

1. **Turn the reader-only invariant into an assertion.** `check_gp_role()` / `check_gp_is_writer()` hold only because `gp_is_writer` (`PGC_S_OVERRIDE` from `gpqeid`) and `gp_role` (`PGC_S_ARGV` / `PGC_S_CLIENT`) never get a GUC stack entry, so a worker abort cannot restore the leader's value. If a QE identity GUC ever gets a stacking source, the slot removal would quietly come back. A cheap guard in `ProcKill()`:

   ```c
   #include "access/parallel.h"
   ...
   		is_slot_creator = (Gp_role == GP_ROLE_DISPATCH ||
   						   (Gp_role == GP_ROLE_EXECUTE && Gp_is_writer));
   		/* A parallel worker is a reader of its leader's session, never the slot's owner. */
   		AssertImply(IsParallelWorker(), !is_slot_creator);
   ```

2. **Drop the test's objects at the end.** `regress_pw_tab`, `regress_pw_loop()` and `regress_parallel_worker_probe(int, bool)` are left behind in the `regression` database. If that is intentional, a comment would help.

3. **Report the leader's writer flag in the probe.** The segment rows exercise the "writer QE leader" path only because a single-slice `gp_dist_random('gp_id')` query lands on the writer gang, which the expected output cannot show. Printing `Gp_is_writer` for the leader (`leader=execute writer=t content=0 ...`; the coordinator will show `writer=f`, which is fine) makes the tested path explicit.

4. **`\!` lines.** The two `psql -X -At -d regression` invocations hard-code the database name and do not pass `-h "$PGHOST" -p $PGPORT` as `cursor.sql` does; the test fails under `pg_regress --dbname=<other>`. Adding `-h/-p` at least matches the existing precedent.

haolinw:
Thanks for the review and for running the pgvector builds against the branch. All four suggestions are taken, in one follow-up commit (33ca771715):

1. **Assertion.** `ProcKill()` now has `AssertImply(IsParallelWorker(), !is_slot_creator)`, placed right after `is_slot_creator` is computed. Your reading of why the pins hold is exactly right: `gp_role` arrives with `PGC_S_ARGV` / `PGC_S_CLIENT` and `gp_is_writer` with `PGC_S_OVERRIDE`, so `set_config_option()` takes the `makeDefault` path and never stacks a rollback entry. The assertion catches the day either of them acquires a stacking source.

2. **Objects left behind.** Kept on purpose: the repo convention (CLAUDE.md) is that regression tests leave their objects in the `regression` database for the downstream tools, and the other `regress.so` functions are left the same way. Added a closing comment saying so.

3. **Leader writer flag.** The probe now prints `leader_writer=t/f` next to `leader=`. This turned out to be more than cosmetic: it showed that the PL/pgSQL loop block was probing from a *reader* gang (`leader_writer=f` on every segment), because `RETURN QUERY ... FROM gp_dist_random('gp_id')` opens a portal and cursors are dispatched to reader gangs. An `INSERT ... SELECT FROM gp_dist_random(...)` had the same problem, since the scan slice under the motion is also a reader gang. The block now launches the segment probes from the `RETURNING` list of an `UPDATE` inside the loop, which runs in the writer slice like the `CREATE INDEX` of the original report, and the expected output shows `leader_writer=t` for all segment probes in every block. The plain and explicit-cursor blocks were already on writers.

4. **`\!` lines.** Both utility-mode invocations now pass `-h "$PGHOST" -p $PGPORT` like `cursor.sql`. `-d regression` stays: every other shell-out in the suite hard-codes it too, and psql performs no variable interpolation on `\!` lines, so there is no portable way to pick up the database name.

Verified on a cassert build under `optimizer` on and off; the failing-worker probes do not trip the new assertion.

linwen (approved):
look good to me.

PR #303, opened 2026-09-20
Count per-host dispatch targets under the role a hot-standby QD dispatches to

huiliang-liu:
## Problem

`getCdbComponentInfo()` counts, per host, the segments this dispatcher sends QEs to, and hands each QE its host's count through gpqeid (`hostPrimaryCount`). The counting loop only ever entered PRIMARY rows into the hash, while the two read-back loops admitted MIRROR rows on a hot-standby QD. A host carrying only mirror rows therefore had no entry to read back: `Assert(found)` fires on assert builds, and release builds dereference a NULL hash entry and the backend segfaults at connect time, before any query runs.

The trigger is wider than the ticket describes. The standby coordinator is itself a `role='m'` row, so any multi-host hot-standby deployment where the standby has its own host (the normal layout) hits this on every standby session, not just clusters with dedicated mirror hosts and not only after failover. Single-host demo clusters hide it by co-hosting every role under one hostname, which is why the hot_standby isolation2 suite never saw it.

A secondary defect on co-hosted layouts: a mirror row received the count of primaries on its host, while the standby QD actually runs one QE per mirror there. After role drift the two differ and resource groups divide per-host memory by the wrong number.

## Fix

Compute one `dispatch_role` per build and use it in the counting loop and both read-back loops, so every row read back was counted:

- MIRROR for a catalog-built table on a hot-standby QD, so a standby's count is mirrors-per-host, which is what the QEs' per-host resource math wants.
- PRIMARY otherwise: a live QD, or any topology-built table. `applyDispatchTopology()` stamps every synthesized row PRIMARY, so the ticket's suggested fix (count MIRROR whenever `IS_HOT_STANDBY_QD()`) would leave every count at 0 on the PR #248 path and each QE would FATAL in `cdbgang_parse_gpqeid_params`.

The read-back `Assert(found)` becomes an `elog(ERROR)` so a future asymmetry surfaces as an error rather than a release-build segfault. Comments on `HostPrimaryCountEntry` and `hostPrimaryCount` now describe a dispatch-target count; the field name is kept to avoid churn in `cdbgang_async.c` and `resgroup.c`.

Behavioral note: on a mirrorless cluster, which was never a supported hot-standby dispatch layout, the standby QD now hands the primaries a host count of 0 and the QEs FATAL on gpqeid, where they previously ERRORed with "primary segments can only process MPP protocol messages from primary QD". Both refuse; only the message differs.

## Test

New isolation2 `hot_standby/dispatch_target_count`. `hostname` in `gp_segment_configuration` is only the grouping key for the count (connections go by `address`), so the test relabels the three segment mirrors onto a host of their own, checks that a fresh standby session builds its component table and dispatches to all of them, then restores the catalog. The standby coordinator row keeps its hostname because the isolation2 driver resolves `-1S` through it; the entry rows go through the same loop and predicate.

Verified locally on a single-host cluster with a hot standby: with the standby coordinator's own row relabeled, unpatched binaries segfault on every standby session and patched binaries work, with gdb showing the standby row counted 1. `hot_standby/dispatch_topology` still passes on the same build, so the topology path is unaffected.


JunfengYang (dismissed):
Walked the fix end-to-end: the single dispatch_role predicate restores the count/read-back invariant, live-QD behavior is unchanged, the topology carve-out is required (applyDispatchTopology stamps every synthesized row PRIMARY — verified), and the recorded-state rebuild closes the in-place-promotion window. LGTM.

Two non-blocking notes:
- hot_standby_schedule only runs via the standalone installcheck-hot-standby target, so CI never exercises the new test — is CI coverage for the hot-standby suite tracked anywhere?
- buildGpQueryString's is_hs_dispatch and the gang-creation Assert in cdbgang.c still evaluate IS_HOT_STANDBY_QD() live; a mid-command promotion pairs a mirror-target table with a primary flag (fail-closed error, self-heals on the next command). Worth confirming that's intentional.

src/include/cdb/cdbutil.h
```diff
@@ -67,7 +67,9 @@ struct CdbComponentDatabaseInfo
 
 	CdbComponentDatabases	*cdbs; /* point to owners */
 
-	int16		hostPrimaryCount;	/* number of primary segments on the same hosts */
+	int16		hostPrimaryCount;	/* number of dispatch targets (primaries, or
```
JunfengYang: Seems the name is not appropriate.

huiliang-liu: Renamed in 1617187720: hostDispatchTargetCount, HostDispatchTargetCountEntry, host_dispatch_target_count and ResGroupGetHostDispatchTargetCount(). No behavior change.

huiliang-liu:
On the two notes:

- CI does run it: the `hot_standby_rocky8` job in the gpdb_main pipeline runs `installcheck-hot-standby`, which covers both the regress and the isolation2 hot_standby schedules.
- `is_hs_dispatch` stays live on purpose. A session promoted mid-command then pairs a mirror-target table with a primary flag, and the mirror QE refuses it ("mirror segments can only process MPP protocol messages from standby QD"). The command fails and the next one rebuilds the table. Fail-closed is preferable to letting a command that started under recovery keep reading the mirrors after promotion. The cdbgang.c Assert on the coordinator row now uses the recorded state (1617187720), so a cassert build does not trip in that window.

JunfengYang (approved):

PR #304, opened 2026-09-21
Export anchor snapshots at restore-point replay on a hot standby

JunfengYang:
## Problem

A read replica built from a restore point serves reads from a cluster whose nodes are paused at that restore point. Advancing it to a newer restore point (a "step") is today a service window: nodes resume and pause independently, so between the first resume and the last verified pause the coordinator and the segments stand at different positions, and a cross-node query can see inconsistent data. The tool marks reads untrusted for the whole window; nothing in the kernel enforces it.

The way out is a snapshot every node exports at the same cluster instant, which a later query imports on every node. A restore point is that instant: `gp_create_restore_point` writes the record under the two-phase commit lock, so the record every node replays marks one cluster-wide cut. This change adds the export side and the registry; the import side (session mode, QD/QE installer) follows in a separate change.

## What this change does

- **Export at restore-point replay.** When the startup process of a hot standby replays an `XLOG_RESTORE_POINT` record it exports the node's xid snapshot at that instant to `pg_anchor_snapshots/<restore point name>` and registers `{name, xmin}` in a fixed-capacity shared-memory registry. The export is gated on archive recovery (restore points replay in crash recovery too) and on `standbyState == STANDBY_SNAPSHOT_READY`: before the running-xacts record has been applied `KnownAssignedXids` is partial and a snapshot would declare in-progress primary transactions visible, so such a restore point gets a LOG line and no anchor on that node. The export runs before the redo loop publishes the record's replay position, so a replay position at or past a restore-point record (`pg_last_wal_replay_lsn()`, the `remote_apply` feedback) implies the anchor is registered on that node or was refused for good.
- **Never an ERROR in the startup process.** An ERROR there is FATAL and a transient ENOSPC would crash-loop the standby. The export does not call `GetSnapshotData()` (its malloc, "global xmin is higher" and distributed-log paths raise ERROR); a small procarray helper, `GetKnownAssignedXidsSnapshot()`, reads `KnownAssignedXids` under `ProcArrayLock` instead. Every failure on the path (invalid or duplicate name, out of memory, any file error) is a WARNING and a skip. The startup reader of the snapshot file allocates without the out-of-memory ERROR and refuses any file larger than the writer can produce. While an anchor is pending its name counts as registered: redo can meet an earlier restore point of that name first (names are not unique on the primary and retirement frees them), and exporting it would overwrite the pending file with the wrong cut.
- **File format and durability.** The exported-snapshot text grammar restricted to what a recovery snapshot carries, framed by `fmt:` and `rp_name:` lines, the record's `tli:` and `lsn:` (its end LSN, what `gp_create_restore_point` returned) and a `crc:` trailer (CRC-32C over everything before it); written through `.tmp` + `durable_rename`, because the file must survive an unclean restart. The startup reader parses the whole grammar strictly and refuses the file on any deviation.
- **Restart.** At every server start only the file named by `whpg_hot_standby_anchor_name` is read and every other file is swept; the directory is created when an older data directory lacks it. The anchor is registered at once only when its record ends at or before the redo start point (that WAL is on disk). Otherwise it is held pending and registered, from the file's xmin, when replay meets the identical record again (nothing is re-exported: the file was written at that position with a complete standby snapshot and a fresh export could be refused below `STANDBY_SNAPSHOT_READY`): after an immediate shutdown or a crash redo restarts before the anchor, and hot standby can open for connections before re-reaching it (it opens at `minRecoveryPoint`, which lags the anchor whenever no page was written back after it), so an anchor registered at start could be imported against a data state that predates it. A file exported on a timeline the recovery history does not have at that LSN is swept, as is a pending anchor whose LSN replay passes without meeting the record. On a standby the sweep spares the GUC-named file when the start could not examine it (`hot_standby = off`, or the subsystem off), so a feature-off restart does not destroy the published anchor.
- **Publication and capacity.** When the GUC is reloaded to a registered anchor, the anchors registered before it are retired (entry and file), so the registry holds at most the anchors exported since the last publication; an unknown name retires nothing, and the export of the name the GUC already carries completes the publication, so a reload that arrives before its restore point is not lost. A full registry evicts its oldest anchor that is not the published one instead of skipping the export: only the published anchor is ever imported, so an unpublished entry has no reader, and the anchor a catch-up is heading for always lands even when the catch-up crosses more restore points than the capacity. Promotion clears registry and directory before `SharedRecoveryState` flips to done. The new file is written and made durable before the eviction, so a failed write never costs an existing anchor.
- **GUCs.** `whpg_max_anchor_snapshots` (PGC_POSTMASTER, default 64, 0 = off) sizes the registry; `whpg_hot_standby_anchor_name` (PGC_SIGHUP, configuration file only) names the published anchor. Both are unsynced.
- **Introspection without a catalog change.** `gp_toolkit.whpg_anchor_snapshots()` (gp_toolkit 1.9 -> 1.10) lists the anchors registered on the node it runs on. Nothing is added to the bootstrap catalog: a catalog version bump would end in-place minor upgrades, and a DR replica's `pg_control` is production's, so the recovery site would have to match production's catalog version exactly.
- **Directory exclusions.** `pg_anchor_snapshots/` contents are excluded from base backups, pg_rewind and gprecoverseg's differential recovery: anchors belong to the node that exported them.

## Tests

- New isolation2 case `hot_standby/anchor_export` in the hot_standby schedule: export at restore-point replay on mirrors and the standby coordinator (utility and dispatch sessions), file grammar, duplicate name skipped, invalid name rejected, eviction of the oldest unpublished anchor at capacity 3 (with and without a published anchor), publish-time retirement, unknown published name retires nothing, a publication that arrives before its restore point completed by the export, clean restart registering only the named anchor and sweeping the rest, crash restart (immediate stop, no checkpoint, replay paused at an earlier restore point) holding the named anchor pending until its record is replayed again, the export gate (`hot_standby = off` on the standby coordinator exports nothing while mirrors do), a fault-injected write failure leaves the standby serving, a checksum-damaged file refused at start, and an FTS failover shows an empty registry and directory on the promoted node. Two cases from the second review: a failed write on a full registry evicts nothing (the would-be victim survives), and a caller that waits only for `pg_last_wal_replay_lsn()` to reach the restore point finds the anchor. The crash-restart case now exports the published name twice before the crash, so redo meets the earlier instance while the later one is pending: it is skipped and the pending file keeps the later instance's snapshot.
- New unit test `src/backend/access/transam/test/anchorsnapshot_test.c`: name validation matrix, register/duplicate/full/invalidate, disabled registry, list order across slot reuse, retire-on-publish and publication-at-export semantics, eviction order with the published anchor protected, file round trip with grammar check, overflowed shape, wrong-name, flipped-digit, trailing-byte, truncated, count-mismatch, missing-field and unknown-format files, sweep.

## Verification (local, plus the `installcheck-hot-standby` CI job from #306 on this PR)

- isolation2 hot_standby schedule on the rebased tree: `anchor_export` passes (also standalone, four iterations of the new cases); `basic`, `dispatch_topology`, `setup`, `teardown` byte-identical; `transaction_isolation`, `query_conflict`, `faults` fail only with the pre-existing local signatures (client library drops PANIC/FATAL text; cumulative `confl_tablespace` counter), identical on unpatched main.
- regress `installcheck-good` 666/674; the 8 failures are build-flag or environment only (JIT: explain_format; no libxml: analyze, qp_orca_fallback; gpfdist: sreh; IPv6 pg_hba: hba_conf, dboptions; a cluster initdb'd before #305's ACL change: bfv_genfile_privileges; a gang-retry NOTICE: alter_db_set_tablespace). None touch this change.
- transam unit tests pass (anchorsnapshot 7: name matrix, register/duplicate/invalidate, disabled registry, list order, retire-on-publish and publication-at-export, eviction order, file grammar with flipped digit / trailing byte / truncation / count mismatch / missing field / unknown format).
- pg_basebackup TAP `t/010` (133), pg_rewind TAP (108) and gpsegrecovery unit tests (47) passed on the previous revision; the exclusion lists are unchanged since.
- gp_toolkit: a database created with `1.9` gains the function after `ALTER EXTENSION gp_toolkit UPDATE`.

## Notes for reviewers

- Trees configured without `--enable-depend` need a clean rebuild: `postgres --catalog-version` and the initdb-written control file otherwise disagree and gpstart refuses to start.
- Minor upgrades that pick this up need `ALTER EXTENSION gp_toolkit UPDATE` on the production cluster; the function reaches replicas by WAL replay.


JunfengYang:
**1. Crash restart registering an anchor ahead of replay.** Confirmed, and it was a design gap rather than an implementation slip. The snapshot file now carries the restore-point record's timeline and end LSN. At start the anchor is registered immediately only when that LSN lies at or before the redo start point (that WAL is already applied on disk); otherwise it is held pending in the startup process and registered when replay meets the identical record (name, timeline, LSN). A file whose timeline disagrees with the recovery history, or whose LSN replay passes without meeting the record, is swept. One precision on the window: hot standby opens at `minRecoveryPoint`, and every page written back during recovery pushes that to the current replay position, so the window exists when no page was written after the anchor was reached (an idle source between restore points, a crash inside the writer interval). It is narrow and the kernel cannot tell when it is open, which is why it is now closed by construction. New isolation2 case: immediate stop of the standby coordinator after two restore points with no checkpoint, replay paused at the earlier one, the later anchor absent while pending and present after resume.

**2. Anchors and recovery-conflict handling.** Correct observation; this is deliberately the next change (conflict linkage), whose design is fixed: at the redo sites that resolve snapshot conflicts, anchors with `xmin <= latestRemovedXid` are invalidated (entry and file) before the standard delay-then-cancel of readers, and new imports of an invalidated anchor fail outright. Replay is never delayed for an anchor. `AnchorSnapshotInvalidate()` is the primitive that change calls. Nothing in this change imports an anchor, so no reader can hold a stale one yet; the wording "importable" in the SRF comments was ahead of itself and now reads "registered".

**3. Reload before the record.** Fixed twice over. Publication is applied at export time as well as at reload: when the name just exported is the one the GUC already carries, the export completes the publication (retiring everything before it). Separately, a full registry no longer skips the export: it evicts the oldest anchor that is not the published one. Only the GUC-named anchor is ever imported, so an unpublished entry has no reader. This also closes a trap the review's scenario is one instance of: during a step's catch-up the GUC still names the previous anchor, so a catch-up crossing more restore points than the capacity would have skipped its own target. New cases: reload naming a not-yet-replayed restore point completed by its export; eviction at capacity 3 with the published anchor kept.

**4. File validation.** The file gains a `fmt:` version line and a `crc:` trailer (CRC-32C over everything before it). The startup reader parses the whole grammar strictly: every field in order, the rp_name line naming this anchor, counts matching, xids normal, nothing after the checksum line. Any deviation is a WARNING and no registration. Unit tests cover a flipped digit, a stray trailing byte, truncation, a count mismatch, a missing field and an unknown format; the isolation2 case damages the published anchor's file and restarts.

**5. hot_standby schedule not in PR CI.** Correct, and older than this change. #306 adds a GitHub Actions job running `installcheck-hot-standby` (EL 8, Postgres planner, serial, the Concourse parameters); this branch is rebased onto it so the new cases run here.

JunfengYang:
**F1 (GUC description).** Fixed: the long description of `whpg_max_anchor_snapshots` now says that a full registry evicts the oldest anchor that is not the published one, and that the export is skipped only when every entry is the published anchor.

**F2 (feature-off start sweeps the published anchor's file).** Taken. On a standby (archive recovery) the sweep now spares the file the GUC names when the start could not examine it (`hot_standby = off`, or `whpg_max_anchor_snapshots = 0`), with a LOG line saying so; the later start that has the feature on registers it again under the normal rules (immediate when its record lies before the redo start point, pending otherwise). A primary still keeps nothing. The gate case in `anchor_export` now asserts the file survives the `hot_standby = off` restart and that the anchor is registered again after the restart that switches hot standby back on.

**F3 (pending confirmation re-derives the snapshot).** Taken, both halves. When replay meets the pending record (name, timeline, end LSN all equal) the anchor is registered from the file's xmin; nothing is re-exported or rewritten. The file was written at that very position by an earlier replay with a complete standby snapshot and passed the whole-file check at start, so a fresh export could only add risk: refusal below `STANDBY_SNAPSHOT_READY` after a restart under a subtransaction-heavy load, or a different `suboverflowed` shape. The eviction and publication rules apply at that registration as they do for an export. The comment at the site says why.

**F4 (test nit).** Taken: the crash-restart case now polls `pg_is_wal_replay_paused()` until it is true before asserting anything, instead of polling for a connection.

**F5 (macro).** Taken: the retry branch of `ANCHOR_APPEND` fails the write with a WARNING when a single line does not fit the buffer, so the CRC never reads past it.

**F6 / F7.** Agreed on all points; no change. The conflict linkage that pins anchors against replayed cleanup is the next change, as the header says; `GRANT ... TO public` follows the gp_toolkit convention and exposes only restore point names; the startup parse allocating with palloc fails the start like every other startup allocation.

The design document carries the two behavioural changes (pending registration from the file; the sweep sparing the GUC-named file) as amendments in its review section.

huiliang-liu (commented):
Three points from reading the export path in `anchorsnapshot.c` and its hook in the redo loop. The registry, file grammar, CRC trailer, timeline check and the basebackup/pg_rewind/gpsegrecovery exclusions all look right to me.

src/backend/access/transam/xlog.c
```diff
@@ -7739,6 +7752,14 @@ StartupXLOG(void)
 						WalSndWakeup();
 				}
 
+				/*
+				 * Export this node's anchor snapshot at a replayed restore
+				 * point, before the pause (if any) is taken there; also
+				 * settles an anchor kept across a restart once replay meets
+				 * its record (or passes it).
+				 */
+				AnchorSnapshotExportOnRestorePoint(xlogreader);
```
huiliang-liu: The export runs after the redo loop has already published `XLogCtl->lastReplayedEndRecPtr` for this record (and woken the walsender to send the reply). So there is a window where `pg_last_wal_replay_lsn()` on this node is already >= the restore point's LSN, and `remote_apply` feedback for it has gone out, while the registry still lacks the anchor because the file write + fsync + `durable_rename` has not finished.

A tool that calls `gp_create_restore_point('rpN')`, waits for the replay LSN to reach the returned position, and then queries `gp_toolkit.whpg_anchor_snapshots()` (or, later, imports the anchor) can find no `rpN` row on that node. The isolation test does not hit this only because it inserts rows after each restore point as an extra barrier.

Could the export move before the `lastReplayedEndRecPtr` update, so that a replay position at or past the record implies the anchor is registered? Otherwise the contract should say callers must wait for a later record.

src/backend/access/transam/anchorsnapshot.c
```diff
@@ -0,0 +1,1386 @@
...
+	}
+	if (st.st_size <= 0 || st.st_size >= MaxAllocSize)
+	{
+		ereport(WARNING,
+				(errmsg("anchor snapshot file \"%s\" is empty or implausibly large", path)));
+		CloseTransientFile(fd);
+		return false;
+	}
+	size = (size_t) st.st_size;
+	buf = palloc(size + 1);
```
huiliang-liu: This runs in the startup process, and `palloc` raises ERROR on allocation failure, which is FATAL there. That breaks the "never an ERROR in the startup process" invariant this file otherwise keeps. The size check above only rejects `>= MaxAllocSize`, so a corrupt or oversized file named by `whpg_hot_standby_anchor_name` can ask for close to 1 GB; under memory pressure the startup process dies at every start and the standby crash-loops until someone removes the file by hand.

Suggest `palloc_extended(size + 1, MCXT_ALLOC_NO_OOM)` with a WARNING and `return false` on NULL, and a tighter upper bound derived from what the grammar can legitimately produce (fixed header plus roughly 11 bytes per xid, capped by `GetMaxSnapshotSubxidCount()` plus the running-xacts maximum).

src/backend/access/transam/anchorsnapshot.c
```diff
@@ -0,0 +1,1386 @@
...
+		{
+			ereport(WARNING,
+					(errmsg("anchor snapshot for restore point \"%s\" not exported: name already registered",
+							name),
+					 errdetail("The existing anchor is kept; a restore point name can become an anchor only once.")));
+			return;
+		}
+	}
+
+	if (registryFull() && !evictOldestUnpublished(name))
```
huiliang-liu: On a full registry `evictOldestUnpublished()` invalidates the oldest entry and unlinks its file here, before `writeAnchorFile()` runs below. If the write then fails (ENOSPC, EIO, or the `anchor_snapshot_export_write` fault), the old anchor is already gone and the new one is skipped, so the node ends up with one anchor fewer than before the restore point. With capacity 3 holding rp5/rp6/rp7 and rp8's write failing, only rp6 and rp7 remain. The fault-injection case in `anchor_export.sql` arranges a free slot first, so it does not cover this ordering.

Suggest writing the file first and evicting only once the write has succeeded, unlinking the new file if registration still fails afterwards. `registerAnchor()` could report full vs duplicate distinctly so the `registryFull()` pre-check can go away and the "only possible if the registry filled up meanwhile" branch below becomes the eviction path.

src/backend/access/transam/xlog.c
JunfengYang: Confirmed and fixed in 684a83c21c. The export call moved ahead of the `lastReplayedEndRecPtr` update in the redo loop (it was after `CheckRecoveryConsistency` and the walsender wake-up), so the contract is now: a replay position at or past a restore-point record implies the anchor is registered on that node or was refused for good. `pg_last_wal_replay_lsn()` and the `remote_apply` feedback can no longer run ahead of the registry. The R-5 pending registration uses the same hook and moves with it. The cost is one file write plus fsync before the position is published, only on restore-point records.

Test: a new case at the end of `anchor_export` waits for the content-0 mirror's `pg_last_wal_replay_lsn()` to reach what `gp_create_restore_point` returned and then queries the registry with no other barrier. Design K2-R5 and the file header state the contract.

src/backend/access/transam/anchorsnapshot.c
JunfengYang: Agreed, this broke the invariant the rest of the file keeps. Fixed in 684a83c21c: the reader allocates through a `startupAlloc()` helper (`palloc_extended` with `MCXT_ALLOC_NO_OOM`) and returns false with a WARNING on NULL, and the size check is now `> anchorFileMaxSize()`, a bound derived from the grammar: sixteen fixed lines of at most 80 bytes plus one `sxp:<xid>` line of at most 16 bytes per known-assigned xid, capped by `GetMaxSnapshotSubxidCount()`. The WARNING names the size and the limit.

src/backend/access/transam/anchorsnapshot.c
JunfengYang: Right, the pre-check ordering was the worst of the two. Fixed in 684a83c21c along the lines you suggest: `registerAnchor()` now reports OK / DUPLICATE / FULL, `registryFull()` is gone, and the export path writes and durably renames the file first, then calls `registerAnchorEvicting()`, which evicts the oldest unpublished anchor only on FULL and retries once. If that still fails (every entry is the published anchor) the new file is unlinked and the registry is untouched. The R-5 pending-registration path uses the same helper.

Test: a new case in `anchor_export` fills the capacity-3 registry (rp2/rp3/rp4), injects the write fault on the content-0 mirror and creates rp8: all three anchors and their files survive there, no eviction is logged, while content 1 (no fault) evicts rp2 as before.

src/backend/access/transam/anchorsnapshot.c
```diff
@@ -0,0 +1,1424 @@
...
+				(errmsg("anchor snapshot for restore point \"%s\" not exported: invalid name",
+						printableName(name)),
+				 errdetail("Restore point names used as anchors must be non-empty, contain no path separators or control characters, and not end in \".tmp\".")));
+		return;
+	}
+
+	{
+		TransactionId dummy;
+
+		if (AnchorSnapshotLookup(name, &dummy))
```
huiliang-liu: Low probability, but silent when it happens, so I think it is worth the few lines.

The duplicate check here consults only the registry, not `pendingName`. Restore point names are not unique on the primary, and retirement on publication frees both the entry and the file, so a name can legitimately be exported twice. Sequence: `X` is exported at LSN Q and later retired by a publication; `X` is exported again at LSN P and the GUC publishes it; the standby crashes and redo restarts before Q. At start the file `X` (lsn P) is held pending. Replay meets `X`@Q first: the registry has no `X`, so this path exports it afresh, overwriting the pending file with Q's snapshot, registers it, and, because the name equals the GUC, completes the publication. When replay reaches `X`@P the pending branch matches, but `registerAnchorEvicting()` returns DUPLICATE and only the "registry full" WARNING fires. The published anchor `X` now carries Q's snapshot while the tool believes it is P's.

whpg-dr names restore points with a second-resolution timestamp, so this needs a hand-made or reused name plus a crash, which is why I rate it low. Suggested fix: while `pendingSet` is true, treat `pendingName` as already registered here (skip the export with the existing "name already registered" WARNING, or a variant naming the pending anchor). Independently, the WARNING in the pending branch should distinguish DUPLICATE from FULL rather than report "registry full" for both, so that a real occurrence is not misread in the log.

huiliang-liu (commented):
Re-read 684a83c21c. The three earlier points are addressed as described: the hook now runs before `lastReplayedEndRecPtr` is published and nothing in `anchorsnapshot.c` reads the replay position; the startup reader allocates without the OOM ERROR and the size bound matches the grammar; the export writes and renames before it takes a slot, and `registerAnchorEvicting()` only evicts on FULL. One new, low-probability item below.

src/backend/access/transam/anchorsnapshot.c
JunfengYang: Good catch, and worth the lines: it is the silent-wrong-cut class. Fixed in ad7ba50fb8.

1. The export path now treats a pending name as registered: after the registry lookup it checks `pendingSet && strcmp(name, pendingName) == 0` and skips with a WARNING ("a snapshot file of that name from an earlier run is pending registration at <lsn>"), keeping the pending file. That check can only be reached for a record BEFORE the pending LSN, since the pending branch above has already handled the match and the passed-it cases, so the record it refuses is exactly the earlier instance of the name.
2. The pending branch reports DUPLICATE ("the name is already registered") and FULL separately.

Test: the crash-restart case now exports pend_y twice before the crash (pend_x, pend_y, pend_z, publish pend_z which retires the first pend_y and frees the name, pend_y again, publish pend_y) and records the second instance's LSN. After the restart redo meets the first pend_y while the second is pending: the log shows the skip WARNING, "exported anchor snapshot ... pend_y" appears exactly twice in the whole run (both before the crash), the pending record is registered from its file, and the file's `lsn:` line is the second instance's.

huiliang-liu (approved):
Re-read ad7ba50fb8. The pending name is now treated as registered on the export path and the pending branch reports DUPLICATE and FULL separately; together with the three earlier fixes (export before the replay position is published, no ERROR-raising allocation in the startup process, write before evict) this covers everything I raised. LGTM.

PR #305, opened 2026-09-21
Require superuser or pg_write_server_files for pg_file_write/rename/unlink

avamingli:
### Problem

`pg_file_write(text,text,bool)`, `pg_file_rename(text,text,text)` and
`pg_file_unlink(text)` have `proacl = NULL` (i.e. PUBLIC EXECUTE): any
role, with no GRANT, can create, overwrite, rename, or delete files
under the data directory (and log directory) — including
`postgresql.auto.conf`, which can be leveraged into code execution as
the postgres OS user via `shared_preload_libraries`/`archive_command`
after a restart/reload.

Repro (on an unpatched build):
```sql
SET SESSION AUTHORIZATION some_nonsuperuser;
SELECT pg_file_write('x.txt', 'hello', false);  -- succeeds today
```
By contrast `pg_read_file()` correctly returns `permission denied` for
the same role, because it *is* revoked from PUBLIC.

### Root cause

These functions live in-core in `src/backend/utils/adt/genfile.c`
(brought in from `contrib/adminpack` during the PG12 merge).
`src/include/catalog/pg_proc.dat` maps them to the `_v1_1` C symbols,
which intentionally contain **no** superuser check — their header
comment says privileges are handled by the GRANT system, matching
`contrib/adminpack`'s design, where every `CREATE FUNCTION ... _v1_1`
is immediately followed by a `REVOKE EXECUTE ... FROM PUBLIC`
(see `adminpack--1.1--2.0.sql`).

The merge commit that pointed the catalog entries at the `_v1_1`
symbols never added the matching `REVOKE` to `system_views.sql`, so
the in-core functions ended up with PUBLIC EXECUTE.
`convert_and_check_filename()`'s path confinement is a read-side check
only (per its own header comment) and was never a substitute for this.

### Fix

- `genfile.c`: add a `requireWriteServerFilesPrivilege()` check
  (superuser or `pg_write_server_files` member) to `pg_file_write_v1_1`,
  `pg_file_rename_v1_1`, `pg_file_unlink_v1_1` — mirrors the existing
  `pg_read_server_files` check used for the read-side functions. This
  takes effect immediately, no initdb needed, so it's the primary
  remediation for existing clusters after upgrade.
- `system_views.sql`: add the matching `REVOKE EXECUTE ... FROM public`
  as defense in depth for new initdbs.
- New regression test `bfv_genfile_privileges` covering: a plain role
  is denied, a `pg_write_server_files` member is allowed, and the
  superuser is allowed.

`gp_toolkit.gp_move_orphaned_files`, the one in-tree caller of
`pg_file_rename`, is run by administrators and is unaffected.

### Test plan
- [x] Verified live on a local build: before the fix, a non-superuser
      role could write/rename/unlink files; after the fix it gets
      `permission denied`, while superuser and `pg_write_server_files`
      members still succeed.
- [x] New `bfv_genfile_privileges` test passes.
- [x] Full local rebuild + cluster restart clean after rebase onto
      current `main`.

### Update

Also restricts `pg_logdir_ls()`, which has the same omission (read-side
listing of log file names, gated on `pg_read_server_files` instead of
`pg_write_server_files`) — per review feedback, added as a separate
commit.

### Manual mitigation for existing clusters (before upgrading)

The `genfile.c` privilege check is the real fix and needs a binary
upgrade, but the ACL-level REVOKE can be applied immediately by a
superuser as a SQL-only stop-gap, with no restart required.

**`pg_proc` privileges are per-database**, so this block must be run in
every connectable database on the cluster, including `template1` (so
that newly-created databases inherit it too) — running it in just one
database (e.g. the app's own db) leaves every other database, and any
database created afterward, still fully exposed:

```sql
REVOKE EXECUTE ON FUNCTION pg_file_write(text,text,boolean) FROM public;
REVOKE EXECUTE ON FUNCTION pg_file_rename(text,text,text) FROM public;
REVOKE EXECUTE ON FUNCTION pg_file_unlink(text) FROM public;
REVOKE EXECUTE ON FUNCTION pg_logdir_ls() FROM public;
-- restore access for whichever admins actually need these, e.g.:
GRANT EXECUTE ON FUNCTION pg_file_write(text,text,boolean) TO pg_write_server_files;
GRANT EXECUTE ON FUNCTION pg_file_rename(text,text,text) TO pg_write_server_files;
GRANT EXECUTE ON FUNCTION pg_file_unlink(text) TO pg_write_server_files;
GRANT EXECUTE ON FUNCTION pg_logdir_ls() TO pg_read_server_files;
```


linwen (commented):
My suggestion:

1. pg_logdir_ls() needs the same treatment (attach to src/backend/catalog/system_views.sql at the new REVOKE block, or to genfile.c):

pg_logdir_ls() has exactly the same defect and should be fixed in this PR. pg_proc.dat maps oid 6050 to pg_logdir_ls_v1_1, whose body carries no privilege check (same "handled by the GRANT system" design as the three write functions), and system_views.sql has no REVOKE for it either, so proacl is NULL and it is PUBLIC-executable. Verified on a current main build: a plain non-superuser role can run SELECT * FROM pg_logdir_ls() and list every log file name in the log directory. In contrib/adminpack/adminpack--1.1--2.0.sql the CREATE OR REPLACE FUNCTION pg_catalog.pg_logdir_ls() ... 'pg_logdir_ls_v1_1' is immediately followed by REVOKE EXECUTE ON FUNCTION pg_catalog.pg_logdir_ls() FROM PUBLIC; (line 43), so this is the same omission from the same merge.

Suggested change: since this is a read-side operation, gate pg_logdir_ls_v1_1 on pg_read_server_files membership (mirroring convert_and_check_filename() rather than the new write check), add REVOKE EXECUTE ON FUNCTION pg_logdir_ls() FROM public; next to the three new REVOKEs, and add a denied/allowed pair for it to bfv_genfile_privileges. Severity is lower than the write functions (file names only; contents still go through the already-revoked pg_read_file), but it is the last ungated member of this family. For reference, pg_logfile_rotate() and pg_file_length() are also PUBLIC at the ACL level but their C bodies call superuser(), so they are not exposed; adding REVOKEs for them is optional consistency, not a fix.

2. Document the REVOKE workaround for existing clusters in PR message:                                                                                                           
REVOKE EXECUTE ON FUNCTION pg_file_write(text,text,boolean) FROM public;                                                                                           REVOKE EXECUTE ON FUNCTION pg_file_rename(text,text,text) FROM publ
REVOKE EXECUTE ON FUNCTION pg_file_unlink(text) FROM public;                                                                                                       REVOKE EXECUTE ON FUNCTION pg_logdir_ls() FROM public;

avamingli:
@linwen 
Both addressed, thanks for catching pg_logdir_ls().

1. pg_logdir_ls() fixed in a separate commit (80fe1e995eb) — gated on pg_read_server_files since it's read-side (file names only), matching convert_and_check_filename()'s existing read-side pattern rather than the write check. Added denied/allowed coverage to bfv_genfile_privileges. Verified live: a plain role gets "permission denied for function pg_logdir_ls" on a fresh initdb, a pg_read_server_files member and superuser still succeed.

Left pg_logfile_rotate()/pg_file_length() alone per your note — confirmed both still call superuser() in genfile.c, so they're not actually exposed; adding REVOKEs there would be redundant, not a fix.

2. Added a "Manual mitigation for existing clusters" section to the PR description with the REVOKE/GRANT statements (including pg_logdir_ls now) as a SQL-only, no-restart stop-gap ahead of the binary upgrade.

linwen (dismissed):
The codes look good to me. I will approve it.
There is another non-blocking suggestion: add a block that actually reaches the genfile.c checks.

This test only exercises the ACL layer. Because CI runs a fresh initdb
with the new system_views.sql, the REVOKEs are in place, and PostgreSQL checks
function EXECUTE privilege during expression initialization (init_fcache() ->
pg_proc_aclcheck()), i.e. before the function body runs. The regular user is
therefore rejected by the ACL, and the new requireWriteServerFilesPrivilege() /
requireReadServerFilesPrivilege() code is never entered. You can see this in the
expected file: every asserted error is "permission denied for function ...", and
the strings this PR adds to genfile.c ("must be superuser or a member of the
pg_write_server_files role to use this function", and the read-side equivalent)
appear zero times.

I confirmed this empirically on an unpatched build (no genfile.c check at all):
replaying only the ACL part of this PR produces exactly
"ERROR: permission denied for function pg_file_write" and
"ERROR: permission denied for function pg_logdir_ls", byte-identical to the
assertions here. In other words, reverting the entire genfile.c change would leave
this test green. That matters because upgraded clusters keep proacl = NULL, so the
C check is their only line of defense, and it currently has no test protection.

Appending something like this covers it:

```sql
-- simulate an upgraded cluster (proacl still NULL) or a stray GRANT:
-- the genfile.c check must still deny
grant execute on function pg_file_write(text,text,boolean) to bfv_genfile_regular_user;
grant execute on function pg_file_rename(text,text,text) to bfv_genfile_regular_user;
grant execute on function pg_file_unlink(text) to bfv_genfile_regular_user;
grant execute on function pg_logdir_ls() to bfv_genfile_regular_user;
set session authorization bfv_genfile_regular_user;
select pg_file_write('bfv_genfile_test.txt', 'hello', false);  -- expect the genfile.c error
select pg_file_rename('bfv_genfile_test.txt', 'bfv_genfile_test2.txt', null);
select pg_file_unlink('bfv_genfile_test.txt');
select count(*) >= 0 as ok from pg_logdir_ls() as t(starttime timestamp, filename text);
reset session authorization;
```

On the same unpatched build, the GRANT-then-call sequence returns 5 (the write
succeeds), whereas with this PR it raises the genfile.c error, so this block is
what distinguishes a patched from an unpatched binary.

Two side benefits: it pins down the behaviour a DBA will actually hit if they
GRANT EXECUTE on these functions to an application role that is not a
pg_write_server_files member (the ACL passes, the C check still denies), which is
otherwise undocumented and a likely support question; and it guards against a
future refactor or merge silently dropping the require* calls.

adam8157:
One note on the "Manual mitigation for existing clusters" section: `pg_proc`
ACLs are per-database, so those statements only take effect in the database
you run them in.

So on an existing cluster the block has to run in every connectable database,
including `template1` so that new databases inherit it.

avamingli:
Both addressed.

@linwen — confirmed empirically first: GRANT EXECUTE + SET ROLE did still get "must be superuser or a member of the pg_write_server_files role..." from genfile.c, so the C check itself was right, but you're correct the test never exercised that path. Added a block in de5782e62c2 that explicitly GRANTs EXECUTE on all four functions to the plain role, then asserts genfile.c's own check still denies — verified this new block fails (goes green with the C check removed) if I revert the genfile.c change, so it now actually pins the behavior you described.

@adam8157 — good catch, verified pg_proc ACLs are per-database (REVOKE in one DB left template1's proacl untouched in a quick test). Updated the "Manual mitigation" section to say the block must be run in every connectable database including template1.

avamingli:
CI hangs, closed and reopen

linwen (approved):
Looks good to me.

PR #306, opened 2026-09-22
ci: run the hot-standby tests in GitHub Actions

JunfengYang:
## Why

The top-level `installcheck-hot-standby` target runs two things: the upstream hot-standby regress tests (`hs_primary_setup` on the primary, then `standby_schedule` against the standby coordinator) and the isolation2 `hot_standby_schedule`. `installcheck-good` runs neither, and `installcheck-world` only recurses into isolation2's plain `installcheck` (isolation2_schedule plus the resource-queue, retrieve-cursor and interconnect schedules). So far these tests ran nowhere in GitHub CI: a pull request touching hot-standby recovery code could be green without them having executed once. Only the Concourse `hot_standby_<os>` job, which runs the top-level target from the source root, covered them.

## What

- One extra matrix combination of the `installcheck` job, `suite: hot-standby` (`installcheck-hot-standby / EL8 / optimizer=off`), runs the top-level `installcheck-hot-standby` target on its own demo cluster with the Concourse job's options (`optimizer=off`, `jit=off`). It is a matrix `include`, not a copy of the job: `suite` overrides a base-matrix value, so the include is a new combination and the existing EL/optimizer combinations are untouched (`suite: regress`).
- It stays on its own cluster because the schedule restarts standbys, fails segments over and reconfigures the cluster; it cannot join an existing target. It runs on every pull request and push and ignores the `el_version` / `installcheck_target` dispatch inputs.
- `run-installcheck.bash` runs the target from the source root, like `installcheck-world`; the failure artifacts add the isolation2 results and the standby coordinator's log; artifact names carry a `-hot-standby` suffix so they do not collide with the EL 8 / optimizer=off regress combination.
- `.github/workflows/README.md` documents the combination (behaviour tables, jobs table, configuration section) and fixes the installcheck timeout it listed (120 -> 180, matching the workflow).

## Validation

The combination runs on this pull request itself. Locally, on a gpdemo cluster with mirrors and a standby coordinator, the regress half passes (hs_primary_setup, hs_standby_check, hs_standby_allowed, hs_standby_disallowed, hs_standby_functions) and the isolation2 schedule passes (12 of 12 in the first CI run of the isolation2-only version of this change).


haolinw (approved):
LGTM

PR #309, opened 2026-09-23
Keep the coordinator's background workers dormant while whpg_dispatch_topology_file is set

JunfengYang:
## Purpose

FTS, dtx recovery and the global deadlock detector never run while the postmaster is in recovery, but they start the moment recovery ends, and FTS probes (dtx recovery dispatches) before the first sleep. On a coordinator that dispatches from a topology file (`whpg_dispatch_topology_file`, #248) the catalog is not authoritative: `gp_segment_configuration` still describes the cluster the WAL came from. The moment such a coordinator is promoted, the first FTS cycle would dial those stale addresses and could send `SYNCREP_OFF` or `PROMOTE` to the source cluster's segments, and dtx recovery would run `gp_terminate_mpp_backends()` there and `ROLLBACK PREPARED` every prepared transaction it does not know. Nothing on the receiving side stops either: the FTS handler's dbid/content check is a WARNING outside assert builds, and a clone's dbids match by construction. The K1a background-worker exemption in `getCdbComponentInfo()` (written so a stray GUC on a live primary cannot crash-loop dtx recovery) is what routes those workers to the stale rows.

## Change

`dispatch_topology_bgworker_gated()` (cdbdispatchtopology.c) gates the coordinator's background workers that would act on the catalog in dispatch role:

- **FTS** skips its whole cycle before it touches the catalog: no component-table read, no `gpsegconfig_dump`, no status-version bump, no probe. `start_count`/`done_count` still move, so `FtsNotifyProber()` waiters return (the K1a refusal of a dispatch connection is itself such a waiter), and a pending probe request is consumed so the loop cannot busy-spin.
- **dtx recovery** sleeps before `recoverTM()` and skips the periodic orphan check, so `*shmDtmStarted` stays unset and dispatch connections keep being refused.
- **GDD** skips its check (in the promotion window it does not even start: it waits for the DTM, which the gate keeps unstarted; the gate matters for a GUC set on a running primary).
- **autovacuum launcher** on the coordinator starts no worker: a coordinator autovacuum worker runs in dispatch role and would otherwise be refused at its first component-table build, once per naptime.

While the DTM is unstarted, every background worker that waits for it (deadlock detector, sweepers, extension workers) does not start at all. dtx recovery shows the gated state in its ps title. When the gate lifts, the two workers that cache a component table discard it, so their next transaction rebuilds from the catalog the operator has corrected in the meantime. The exemption that lets them build that table at connection time stays (an ERROR there would crash-loop them); a table that is never dispatched on is harmless. The predicate is the GUC, not `whpg_dispatch_topology_state`: that string reads `error` on any node out of recovery, which is exactly when the gate must hold. Clearing the GUC and reloading lifts the gate on the next cycle; each worker logs its two transitions once (`<worker>: <activity> suspended while whpg_dispatch_topology_file is set` / `... resumed after whpg_dispatch_topology_file was cleared`).

The gate is not promote-specific: a GUC stray-set on a live primary silences the workers there too. User dispatch is already refused on such a node (#248), so the visible effect is that the node stays behind "waiting for distributed transaction recovery" until the GUC is cleared, with a LOG line naming the cause.

This makes the promotion of a read replica safe to sequence as: promote -> fix `gp_segment_configuration` over a utility-mode connection (workers dormant, dispatch connections refused) -> clear the GUC and reload -> the workers wake on the corrected catalog.

No new exported symbol besides the helper; no shared-memory or catalog change. New fault points `ftsLoop_topology_gated` and `dtx_recovery_topology_gated` (the latter added to the dtx-recovery allow-list in faultinjector.c).

## Tests

New isolation2 case `hot_standby/fts_topology_gate`, last in the hot_standby schedule: the standby coordinator gets a topology file and the GUC, is promoted, and the test asserts that FTS and dtx recovery start gated (gated cycles counted through the fault points, the `fts_probe` fault never reached, no `gpsegconfig_dump` written, exactly one suspension line per worker, state string `error`, a dispatch connection refused outright), and that clearing the GUC plus reload wakes both (`fts_probe` reached, dump written, resumption lines, a dispatch session and `gp_request_fts_probe_scan()` succeed). The standby is then destroyed and rebuilt with `gpinitstandby`, following `segwalrep/dtm_recovery_on_standby`; its log is kept beside the data directories and uploaded by CI on failure. The case must stay last and never run in parallel: once un-gated, the promoted clone's dtx recovery dispatches into the suite's live primaries.

Verified locally: the autovacuum launcher gate and the ps title by hand on a promoted standby with `autovacuum = on`, `autovacuum_naptime = 5` (one suspension line, no worker launched, no refusal; before the gate the same setup logged one refused worker per naptime); `make -C src/backend/fts/test check` (24/24), the full `installcheck-hot-standby` target (new case green; the three pre-existing local failures are the missing `FATAL`/`PANIC` line on connection termination from the local psycopg2 build and reproduce on main), isolation2 `gdd/*` subset, `segwalrep/dtm_recovery_on_standby`, `fts_manual_probe`, `fts_errors`, `fts_session_reset`, `fts_segment_reset` on a normal cluster, `installcheck-good` (669/674, known build-flag failures).


src/backend/cdb/cdbutil.c
```diff
@@ -544,18 +546,21 @@ getCdbComponentInfo(void)
...
-	 * ordinary user queries, and recovery completes so the cluster comes up.
+	 * dtx-recovery worker builds a component table while it connects, and
+	 * an ERROR there would crash-loop the worker.  So a background worker
+	 * builds from the raw catalog rows -- and never dispatches on them while
+	 * the GUC is set: dtx recovery and the global deadlock detector sleep
+	 * behind dispatch_topology_bgworker_gated() and discard this table when
+	 * the gate lifts, FTS skips its cycle.  A GUC stray-set on a live
+	 * primary therefore refuses user dispatch (applyDispatchTopology) and
+	 * holds *shmDtmStarted unset until the GUC is cleared, with a LOG line
+	 * naming the cause; nothing acts on rows the GUC declared untrusted.
```
huiliang-liu: Non-blocking. "nothing acts on rows the GUC declared untrusted" is a bit stronger than what holds for a bgworker that is *already running* when the GUC is stray-set on a live primary (an extension worker in dispatch role, say). Its table was built through this exemption with `topology_signature = ""`, so every `cdbcomponent_updateCdbComponents()` sees a signature mismatch against the file, destroys the table, and rebuilds it through the same exemption -- and it keeps dispatching on the raw catalog. That part is pre-existing from #248, but with FTS and the GDD now dormant that worker runs without failover detection or deadlock checks until the GUC is cleared.

I think documenting the gap here (and in the PR description) is enough for this PR; the alternative would be to ERROR on a non-startup rebuild for a bgworker on a non-hot-standby node, which is a bigger behaviour change than this change wants.

src/backend/cdb/cdbdtxrecovery.c
```diff
@@ -770,6 +744,85 @@ DtxRecoveryMain(Datum main_arg)
...
+		 * the operator has fixed the catalog and cleared the GUC.  Sleep on
+		 * the latch: SIGHUP sets it, and the reload is what lifts the gate.
+		 */
+		if (dispatch_topology_bgworker_gated(&topology_gated,
+											 "dtx recovery", "dispatch"))
+		{
+			set_ps_display("gated by whpg_dispatch_topology_file", false);
+			SIMPLE_FAULT_INJECTOR("dtx_recovery_topology_gated");
+
+			rc = WaitLatch(&MyProc->procLatch,
```
huiliang-liu: Non-blocking. This wait block (WaitLatch / ResetLatch / WL_POSTMASTER_DEATH / continue) and the `was_gated` discard block below are the same lines as in `GlobalDeadLockDetectorLoop()` (gddbackend.c:184 and the block after it). A small helper pair in cdbdispatchtopology.c, e.g. `dispatch_topology_bgworker_wait(timeout_ms, wait_event)` and `dispatch_topology_bgworker_discard_components()`, would keep them from drifting apart. Also `set_ps_display("gated by ...")` is re-issued every gated iteration; it only changes on the `!was_gated` transition, so it could move there. Fine to leave as is if you prefer.

huiliang-liu (commented):
Read through the whole change; the gate mechanism itself looks correct to me (SIGHUP wakes all three loops, the gated FTS cycle keeps the start/done handshake and consumes the pending request, the `was_gated` discard order is right, and the first un-gated FTS cycle still writes the dump through the `status_version == 0` path). Two non-blocking comments inline; neither needs to hold the merge.

src/backend/cdb/cdbutil.c
JunfengYang: Agreed, the sentence claimed more than the code guarantees. Reworded in 2494923491: the kernel's own workers (FTS, dtx recovery, the deadlock detector, the autovacuum launcher) never act on the rows while the GUC is set; a dispatch-role background worker already running when the GUC is stray-set on a live primary keeps dispatching on the raw catalog (its table carries an empty topology signature, so every `cdbcomponent_updateCdbComponents()` rebuilds it through the same exemption) and does so without failover detection or deadlock checks until the GUC is cleared. The comment now says so and names why the ERROR alternative is not taken (it would crash-loop such workers, and the promotion window has none of them: DTM-waiting workers never start while the gate holds). The PR description carries the same paragraph.

src/backend/cdb/cdbdtxrecovery.c
JunfengYang: Done in 2494923491: `dispatch_topology_bgworker_wait(timeout_ms, wait_event)` (WaitLatch / ResetLatch / postmaster-death exit) and `dispatch_topology_bgworker_discard_components()` (gangs + component table) live in cdbdispatchtopology.c and both loops call them; the dtx-recovery loop keeps its fault point and ps title at the call site, and the ps title is now set only on the not-gated-to-gated transition (`!was_gated`), cleared as before when the gate lifts. Rebuilt with cassert; the FTS unit tests pass; the hot-standby CI job re-runs `fts_topology_gate`.

huiliang-liu (approved):
Re-reviewed 2494923491 against 80f8395668. The four changed objects compile cleanly here and CI is green on the new head. LGTM.

PR #310, opened 2026-09-24
Require TLS 1.2 or later in the gpfdists and s3 libcurl clients

huiliang-liu:
## Why

SonarQube flags `cpp:S4423` ("Use stronger SSL and TLS versions") on `gpcontrib/gpcloud/lib/ini.cpp` and `gpcontrib/gpcloud/src/s3restful_service.cpp`. The `gpfdists://` client in `url_curl.c` has the same pattern but is not flagged. All three set `CURLOPT_SSLVERSION` to `CURL_SSLVERSION_TLSv1`, or do not set it at all. That constant means "TLS 1.0 or later", so the negotiated floor depends on the host crypto policy and on the peer rather than on our code. A host on the LEGACY policy, or a private S3-compatible store, can still fall back to TLS 1.0/1.1.

Option B (removing gpcloud) was not taken: it is a product-scope decision with no cross-team answer on the ticket, it would still need the `url_curl.c` change, and it does not belong on a stable branch. It can be tracked separately as a deprecation for WHPG 7.

## What

- `src/backend/access/external/url_curl.c`: `extssl_protocol` becomes `CURL_SSLVERSION_TLSv1_2`; the stale comment above it is updated. gpfdist already refuses TLS 1.0/1.1 on the server side (`SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1` in `gpfdist.c`), so this changes nothing for existing gpfdists deployments.
- `gpcontrib/gpcloud/lib/ini.cpp`: the same constant, used when the s3 config file is fetched from a `config_server=https://` URL.
- `gpcontrib/gpcloud/src/s3restful_service.cpp`: the option is set once in the `CURLWrapper` constructor, which covers get/put/post/head/delete.

`CURL_SSLVERSION_TLSv1_2` needs libcurl >= 7.34.0; EL8 ships 7.61.1 and EL9 7.76.1. The value stays a compile-time constant on purpose: this is a security floor, not something to expose as a GUC or `s3.conf` knob.

## Release note

External tables using the `s3://` and `gpfdists://` protocols now require the remote endpoint to support TLS 1.2 or later. An S3-compatible store that only speaks TLS 1.0/1.1 stops working; such a store is already unreachable from an EL9 host on the DEFAULT crypto policy.

## Test

- `url_curl.o`, `gpcloud/lib/ini.o` and `gpcloud/src/s3restful_service.o` build cleanly with the tree's warning flags. Note that GitHub Actions on `main` does not pass `--enable-gpcloud`, so CI does not compile the gpcloud part; it was built locally.
- Negotiation check with the same libcurl 7.76.1 against an `openssl s_server` capped at TLS 1.1, with the process crypto policy relaxed to allow TLS 1.0: the old floor negotiates TLSv1.1 and completes the request (HTTP 200); the new floor fails the handshake with `alert protocol version`.
- No functional change is expected for `gpfdists://` (server already TLS 1.2+) or for AWS S3 (TLS 1.2+ only since 2023), so the existing gpfdists regression is sufficient coverage once CI runs.


linwen (approved):
looks good to me.

PR #311, opened 2026-09-24
Require TLS 1.2 or later in the gpfdists and s3 libcurl clients (WHPG 6)

huiliang-liu:
Same change as #310 on `main` (Option A); the three touched lines are identical on both branches even though `url_curl.c` has otherwise diverged, so this is a hand-applied equivalent rather than a cherry-pick.

## Why

SonarQube flags `cpp:S4423` ("Use stronger SSL and TLS versions") on `gpcontrib/gpcloud/lib/ini.cpp` and `gpcontrib/gpcloud/src/s3restful_service.cpp`. The `gpfdists://` client in `url_curl.c` has the same pattern but is not flagged. All three set `CURLOPT_SSLVERSION` to `CURL_SSLVERSION_TLSv1`, or do not set it at all. That constant means "TLS 1.0 or later", so the negotiated floor depends on the host crypto policy and on the peer rather than on our code.

On 6X `--enable-gpcloud` defaults to yes, so gpcloud ships in every WHPG 6 build.

## What

- `src/backend/access/external/url_curl.c`: `extssl_protocol` becomes `CURL_SSLVERSION_TLSv1_2`; the stale comment above it is updated. gpfdist already refuses TLS 1.0/1.1 on the server side (`SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1` in `gpfdist.c`), so this changes nothing for existing gpfdists deployments.
- `gpcontrib/gpcloud/lib/ini.cpp`: the same constant, used when the s3 config file is fetched from a `config_server=https://` URL.
- `gpcontrib/gpcloud/src/s3restful_service.cpp`: the option is set once in the `CURLWrapper` constructor, which covers get/put/post/head/delete.

`CURL_SSLVERSION_TLSv1_2` needs libcurl >= 7.34.0; EL8 ships 7.61.1 and EL9 7.76.1. The value stays a compile-time constant on purpose.

## Release note

External tables using the `s3://` and `gpfdists://` protocols now require the remote endpoint to support TLS 1.2 or later. An S3-compatible store that only speaks TLS 1.0/1.1 stops working; such a store is already unreachable from an EL9 host on the DEFAULT crypto policy.

## Test

- Configured this branch with `--enable-gpcloud`; `url_curl.o`, `gpcloud/lib/ini.o` and `gpcloud/src/s3restful_service.o` build cleanly with the tree's warning flags.
- Negotiation check with libcurl 7.76.1 against an `openssl s_server` capped at TLS 1.1 (process crypto policy relaxed to allow it): the old floor negotiates TLSv1.1 and completes the request (HTTP 200); the new floor fails the handshake with `alert protocol version`. See #310 for details.


linwen (approved):
looks good to me.

